PR #302, opened 2021-05-11
Upgrade to cobra v1.1.3 from v0.0.5

ptnapoleon:
### What does this PR do?

Upgrades the version of spf13/cobra we're using from 0.0.5 to 1.1.13

### Motivation

In preparation for adding a new CLI that uses cobra, I felt we should upgrade the current version we use first.

### Testing Guidelines

I can confirm the injector still builds, and I've been able to use disruptions locally

### Additional Notes

Anything else we should know when reviewing?


takakonishimura (approved):
lgtm

PR #305, opened 2021-05-14
Better handle the case where count is set to -1

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug
- [ ] Improves Documentation

Three changes:
(1) Add `count < 0` validation to the admission webhook
(2) Calculate `targetsCount` from `matchingTargets` always, to prevent undercounting
(3) Calculate the label selector is wrong if we get no `matchingTargets` rather than `eligibleTargets`. We are already handling the case where `ignored >= matching` later on. You can have a perfectly valid label selector and get no `eligibleTargets`. The simplest way to know if the label selector is bad is to check `matchingTargets`

Motivation for change:
A user tried setting `count: -1`, and didn't get an event about their count being formatted wrong.

### Testing

### Checklist


Azoam (approved):

takakonishimura (approved):
yay!

controllers/disruption_controller.go
```diff
@@ -600,32 +600,27 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
+	}
+
 	// instance.Spec.Count is a string that either represents a percentage or a value, we do the translation here
-	targetsCount, err := getScaledValueFromIntOrPercent(instance.Spec.Count, len(eligibleTargets), true)
+	targetsCount, err := getScaledValueFromIntOrPercent(instance.Spec.Count, len(matchingTargets), true)
 	if err != nil {
 		targetsCount = instance.Spec.Count.IntValue()
 	}
 
-	// computed count should not be 0 unless the given count was not expected
```
ptnapoleon, line 624: `targetsCount` can no longer be 0 here, so we don;t need this check. This was returning an error incorrectly in the case where `ignoredTargets` == `matchingTargets` anyway.

controllers/disruption_controller.go
```diff
@@ -600,32 +600,27 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
+	// prune ignored targets from all targets
+	for _, target := range matchingTargets {
+		if !contains(instance.Status.IgnoredTargets, target) {
+			eligibleTargets = append(eligibleTargets, target)
+		}
+	}
+
 	// instance.Spec.Count is a string that either represents a percentage or a value, we do the translation here
-	targetsCount, err := getScaledValueFromIntOrPercent(instance.Spec.Count, len(eligibleTargets), true)
+	targetsCount, err := getScaledValueFromIntOrPercent(instance.Spec.Count, len(matchingTargets), true)
```
ptnapoleon, line 619: Imagine the case where count = 100%, matching = 5, ignored = 3, and eligible = 2. In that situation, we're going to get `2` as the targetsCount, where we will immediately subtract 3, get a final count of -1 and end the disruption early. We should expect when count = 100% that we will target len(matching), which is now the behavior

ptnapoleon: It is perhaps easier to explain the issue with the current code by saying that when we calculate a scaled percentage count, we are subtracting ignored, calculating the percentage from the reduced total, then subtracting ignored again. This is causing us to under-disrupt

api/v1beta1/validations.go
```diff
@@ -0,0 +1,60 @@
...
+			return 0, false, fmt.Errorf("invalid type: string is not a percentage")
+		}
+
+		v, err := strconv.Atoi(s)
+
+		if err != nil {
+			return 0, false, fmt.Errorf("invalid value %q: %v", intOrStr.StrVal, err)
+		}
+
+		return v, isPercent, nil
```
takakonishimura, line 37: nitpick, do we need `isPercent` here?

controllers/disruption_controller.go
```diff
@@ -600,32 +600,27 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
 	// return an error if the selector returned no targets
-	if len(eligibleTargets) == 0 {
+	if len(matchingTargets) == 0 {
 		r.log.Info("the given label selector did not return any targets, skipping")
 		r.Recorder.Event(instance, "Warning", "NoTarget", "The given label selector did not return any targets. Please ensure that both the selector and the count are correct (should be either a percentage or an integer greater than 0).")
 
 		return nil
 	}
 
+	// prune ignored targets from all targets
```
takakonishimura, line 611: I would even move this below `getScaledValueFromIntOrPercent`, and my personal preference would be to declare `eligibleTargets := []string{}` right above this for loop as well, though I believe that idea was rejected a few months ago, don't remember why

takakonishimura (approved):
awesome refactoring! left style comments

api/v1beta1/validations.go
ptnapoleon: You're suggesting we just change it to `true`? Hmm, it might be tougher to realize why we're setting it to true though?

controllers/disruption_controller.go
ptnapoleon: Makes sense!

api/v1beta1/validations.go
takakonishimura: In that case, what if we define `isPercent := false` in line `19` and then return:
```return 0, isPercent, fmt.Errorf("invalid type: neither int nor percentage")```

PR #307, opened 2021-05-19
CHAOS-51: Implement the "create" command of chaosli; Help users build a disruption spec

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

Adds the new command to `chaosli`: `create`, which walks a user through creating a disruption file

Motivation for change:

### Testing

### Checklist


cli/chaosli/cmd/create.go
```diff
@@ -0,0 +1,485 @@
...
+		path, _ := cmd.Flags().GetString("path")
+		err = ioutil.WriteFile(path, y, 0644)
+		if err != nil {
+			fmt.Printf("writeFile err: %v", err)
+		}
+	},
+}
+
+const intro = `Hello! This tool will walk you through creating a disruption. Please reply to the prompts, and use ctrl+c to end.
+The generated disruption will have "dryRun:true" set for safety, which means you can safely apply it without injecting any failure.`
```
Azoam, line 52: Anyway we can bold the dryRun part. When I run the command it isnt automatically obvious that dryRun is turned on. Another alternative is giving them the option of turning dryRun off or on.

cli/chaosli/cmd/create.go
```diff
@@ -0,0 +1,485 @@
...
+func getDiskPressure() *v1beta1.DiskPressureSpec {
+	if !confirmKind("Disk Pressure", "Applies IO pressure to the target") {
+		return nil
+	}
+
+	spec := &v1beta1.DiskPressureSpec{}
+
+	spec.Path = getInput(
+		"Specify a path to apply IO pressure to",
+		"Specify a specific mount point to target a specific disk",
```
Azoam, line 282: I think some examples here would be helpful, I know personally the Disk Pressure options can sometimes be confusing. A quick (e.g. /mnt/data) should suffice for example.

cli/chaosli/cmd/create.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+func getNodeFailure() *v1beta1.NodeFailureSpec {
+	if !confirmKind("Node Failure", "This will either shutdown or restart the targeted node (or node hosting the targeted pod)") {
+		return nil
+	}
+
+	spec := &v1beta1.NodeFailureSpec{}
+	spec.Shutdown = confirmOption("Would you like to shutdown the node permanently?",
+		"Choosing yes will terminate the VM completely, and a new node will be scheduled. If you don't enable this, we will just restart the target node.")
```
Azoam, line 314: I dont know if new nodes are scheduled by default. The team would require a nodegroup, not sure if we have any teams at Datadog that manage their own nodes.

cli/chaosli/cmd/create.go
```diff
@@ -0,0 +1,485 @@
...
+	spec.Services = getServices()
+
+	spec.Flow, _ = selectInput(
+		"Choose a flow direction",
+		[]string{v1beta1.FlowEgress, v1beta1.FlowIngress},
+		fmt.Sprintf("%s will affect traffic leaving the target. %s will not really affect traffic entering the target, but actually will affect replies to the inbound traffic.",
+			v1beta1.FlowEgress, v1beta1.FlowIngress),
+	)
+
+	if confirmOption("Would you like to drop packets?", "Packets will be dropped before leaving the target") {
```
Azoam, line 396: Depending on the flow, packets could be dropped upon entering the target (ingress) I believe.

Azoam: I guess this comment goes for all the places where you mention "Packets will be Xd **before** leaving the target.

cli/chaosli/cmd/create.go
```diff
@@ -0,0 +1,485 @@
...
+
+	wrappedResult := intstr.FromString(result)
+
+	return &wrappedResult
+}
+
+func getSelectors() labels.Set {
+	selectors := getSliceInput(
+		"Add a label selector[s] for targeting.",
+		"Please specify this in the form of `key=value`, e.g., `app=hello-node`. One label selector per new-line. If you specify multiple, we will only target the union.",
```
Azoam, line 449: I think an example of how multiple selectors should be added to the cli would be really helpful. The text is helpful in itself, but for visual learners, just seeing the example will be easier to digest. Will not lie, as simple as your explanation was, still found myself reading it twice to understand.

Azoam (commented):
Something else that could be nice is an official Ctrl+C, for those who no longer want to use the tool and just want out haha.

cli/chaosli/cmd/create.go
ptnapoleon: That's true, I am describing DD specific behavior, good point

cli/chaosli/cmd/create.go
ptnapoleon: Well, technically we are also dropping ingress packets before they leave the target as well, as `ingress` really means `replies`. Maybe we should rename it....

ptnapoleon:
> Something else that could be nice is an official Ctrl+C, for those who no longer want to use the tool and just want out haha.

I'll try to build in support for ctrl+D (Exit) to do so

cli/chaosli/cmd/create.go
Azoam: ah yeah thats fair. Yeah I think `ingress` is going to be confusing as more people end up using the chaos-controller.

cli/chaosli/cmd/create.go
ptnapoleon: I'm not sure it's straightforward to bold terminal output, so I'm going to switch to making this a flag that defaults to true

cli/chaosli/cmd/create.go
```diff
@@ -6,29 +6,618 @@
...
+
+	prompt := &survey.Confirm{
+		Message: query,
+		Help:    helpText,
+	}
+
+	err := survey.AskOne(prompt, &result)
+
+	if err == terminal.InterruptErr {
+		os.Exit(1)
```
Azoam, line 214: Just a thought; I know that the exit code can be changed to help better identify locations of errors. Wonder if it would be helpful in this situation to use different exit codes depending where the exit occurs.

cli/chaosli/cmd/create.go
```diff
@@ -6,29 +6,618 @@
...
+
+	spec := &v1beta1.DiskPressureSpec{}
+
+	spec.Path = getInput(
+		"Specify a path to apply IO pressure to, e.g., /mnt/data",
+		"Specify a specific mount point to target the disk mounted there",
+		survey.WithValidator(survey.Required),
+	)
+
+	if confirmOption("Would you like to apply read pressure?", "This applies read-based IO pressure (check the docs)") {
```
Azoam, line 359: I know we use the word "pressure" exclusively, but this makes me think that we stress the disk when we do not. I think its fine to refer to it as disk pressure, but maybe for these prints (read and write) we can make it clear that its thresholds that are being put in place.

"This applies read-based IO thresholds"
"Specify the target threshold in bytes per second"
As examples.

Azoam (approved):

cli/chaosli/cmd/create.go
ptnapoleon: This is a good time for me to re-read the disk pressure docs, as I dont know how they work.

cli/chaosli/cmd/create.go
ptnapoleon: If this line were in the controller, 100%, but absolutely no one should ever be using this feature outside of a human-CLI interaction, so there's no need to specially identify where we exited. It may be useful to separate _why_ we exited, and use a different exit code for ctrl+c vs. a "real" error, but as long as it's just a ctrl+c, any positive exit code (as long as we're consistent), should be good

Azoam: Oh great point, yes that makes a lot of sense, thanks for explaining.

PR #309, opened 2021-05-20
Validate Update: global disruption scope

nathan-tournant:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

- Factorised the interface generation switch and disruption existence to a generator function
- Added Validator interface to DisruptionSpec struct, notably to validate disruption incompatibility
- Other applied global rules: Cannot pass empty disruption, nor apply network and DNS disruptions at the same time, etc.
- Admission Controller now applies all validation rules

Motivation for change:
Make DNS and Network incompatible, more generally gaining control/access to global disruptions parameters for validation is the main motive, along with some code factorising/cleanup.

Still ongoing test file changes in order to fit new constraints (ie. full-on five processes disruptions are not possible anymore)

All might be reworked as research for a better global validation solution is underway

### Testing

- [X] I used existing unit tests (test adaptation ongoing)

### Checklist


api/v1beta1/disruption_types.go
```diff
@@ -91,13 +94,85 @@ func init() {
...
+		return errors.New("cannot target specific container of a node-level disruption")
+	}
+
+	// Rule: at least one disruption field
+	if s.DNS == nil &&
+		s.CPUPressure == nil &&
+		s.Network == nil &&
+		s.NodeFailure == nil &&
+		s.DiskPressure == nil {
+		return errors.New("cannot apply a disruption with no target process")
```
ptnapoleon, line 120: Is this the right error message?

api/v1beta1/disruption_types.go
```diff
@@ -91,13 +94,85 @@ func init() {
...
+	//Rule: DNS and Network disruptions are incompatible
+	if s.DNS != nil && s.Network != nil {
+		return errors.New("cannot apply DNS and Network disruptions concurrently")
+	}
+
+	return nil
+}
+
+// validates rules for disruption global scope and all subsequent disruption specifications
+func (s *DisruptionSpec) ValidateDisruptionSpec() error {
```
ptnapoleon, line 132: Having both `DisruptionSpec.Validate()` and `DisruptionSpec.ValidateDisruptionSpec()` is a bit confusing. I see that this method calls `Validate()`, I would suggest the highest level function takes the most general name, so this would be `Validate()` and the other would become ~`ValidateTopLevelOptions()` (or something else, it isnt like that's a good name). 

Also, remember that if a method shouldn't be called externally, it shouldn't be exported. I haven't finished reading, but if what is now `Validate()` is only called within `ValidateDisruptionSpec()`, it's best to lowercase the name, so that it isn't exported.

ptnapoleon, line 132: This is a big change, but I think it would be most helpful to users if this top-level validate, and the validate of each Kind returned a slice of errors, rather than a single error. That way a user can see everything wrong with their disruption at once, rather than editing->retrying and being surprised that there are more errors we didn't warn them of

api/v1beta1/disruption_types.go
```diff
@@ -91,13 +94,85 @@ func init() {
...
+	case chaostypes.DisruptionKindDNSDisruption:
+		validator, generator = s.DNS, s.DNS
+	case chaostypes.DisruptionKindCPUPressure:
+		validator, generator = s.CPUPressure, s.CPUPressure
+	case chaostypes.DisruptionKindDiskPressure:
+		validator, generator = s.DiskPressure, s.DiskPressure
+	}
+
+	// ensure that the underlying disruption spec is not nil
+	disruptionExists := !reflect.ValueOf(validator).IsNil() && !reflect.ValueOf(generator).IsNil()
```
ptnapoleon, line 175: You almost certainly don't need the `&&` here, because it's the same pointer in both situations

api/v1beta1/disruption_types.go
```diff
@@ -91,13 +94,85 @@ func init() {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// validates existence and generates instances of DisruptionKind Interfaces: DisruptionValidator, DisruptionArgsGenerator
+func (s *DisruptionSpec) DisruptionKindInterfaceGenerator(kind chaostypes.DisruptionKind) (bool, chaosapi.DisruptionValidator, chaosapi.DisruptionArgsGenerator) {
```
ptnapoleon, line 156: I've checked all the call-sites of this method, and you can simplify the return values here. Ideally, we could return only a single value, which is one of `s.NodeFailure`, `s.Network`, `s.DNS`, `s.CPUPressure`, or `s.DiskPressure`. The `bool` isn't needed, because the call-site can just use `!= nil` as they did before. We wouldn't have to return the two interfaces separately, because the returned value actually satisfies both of them. 

Why don't we do this currently? Well it's because that there's no explicit relationship between the two interface types, and there's no other "interface" or "type" that the five kinds satisfy. However, clearly from how you've been forced to return the same object twice in this fashion, it seems to me that there should be! I recommend collapsing `chaosapi.DisruptionValidator` and `chaosapi.DisruptionArgsGenerator` into a single interface that all five kinds satisfy, and making that be the sole return value here.

Then this method becomes purely a "map the name of a kind into the real kind" function, which is what the call-sites all really want.

ptnapoleon (changes requested):

api/v1beta1/disruption_types.go
nathan-tournant: That is correct and will be fixed

api/v1beta1/disruption_types.go
nathan-tournant: I agree, though it's a big change so it's something i thought i'd change later in the global validation update i'm experimenting with

nathan-tournant: Makes sense, both are now swapped and renamed (sort of)

api/v1beta1/disruption_types.go
nathan-tournant: Fixed

api/v1beta1/disruption_spec.go
```diff
@@ -91,13 +94,79 @@ func init() {
...
+	//Rule: DNS and Network disruptions are incompatible
+	if s.DNS != nil && s.Network != nil {
+		return errors.New("cannot apply DNS and Network disruptions concurrently")
+	}
+
+	return nil
+}
+
+// returns this DisruptionSpec's instance of a DisruptionSubSpec based on given kind
+func (s *DisruptionSpec) DisruptionKindPicker(kind chaostypes.DisruptionKind) chaosapi.DisruptionSubSpec {
```
takakonishimura, line 155: Noice....

api/disruption_subspec.go
```diff
@@ -11,9 +11,13 @@ import (
 	chaostypes "github.com/DataDog/chaos-controller/types"
 )
 
-// DisruptionArgsGenerator generates args for the given disruption
-type DisruptionArgsGenerator interface {
+// Contains all methods required for a disruption sub-specification (Network, DNS, CPUPressure, etc.)
+type DisruptionSubSpec interface {
```
takakonishimura, line 15: are subspec and "kind" the same thing?

ptnapoleon: Yes, but `DisruptionKind` as a type name is already taken. Not to bikeshed, whatever you want to name it is fine, but I'll put my 2 cents in for `DisruptionType` or `IDisruptionKind`. I'll even put in 4 cents for renaming `types.DisruptionKind` to `types.DisruptionKindName` and then call this type `DisruptionKind`

ptnapoleon (approved):
Other than choosing the name of the new disruption interface, this all lgtm! And even that name can be merged as-is, imo

api/disruption_subspec.go
nathan-tournant: Done ! It does feel good once changed, that was the good call
Revamping the tests today and we can push this through :)

api/v1beta1/disruption_spec.go
```diff
@@ -20,8 +20,11 @@ package v1beta1
 import (
 	"crypto/md5"
 	"encoding/json"
+	"errors"
```
ptnapoleon, line 23: Everything looks great, except for some small changes to this file that I commented on in #320, though I can move those comments here if that's more helpful

nathan-tournant: No need, I'll fix things asap, retest then merge

ptnapoleon (approved):
+1, it's fine to move the other points to follow-ups as discussed in Slack

PR #318, opened 2021-05-26
CHAOS-43: Yaml to English Feature

Azoam:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug
- [X] Improves Documentation

Adds the explanation feature of the chaos cli. Translating the Yaml of Disruptions, which can often times be daunting and confusing for new users, into a readable format.

Motivation for change:

Increase the usage of the chaos controller through accessibility. 

### Testing

I've tested the explain feature on all of the current examples located in chaos-controller/examples/...

### Checklist

- [X] My code is sufficiently commented (made use of self-commenting function names and variables)


go.mod
```diff
@@ -42,6 +42,7 @@ require (
 	golang.org/x/net v0.0.0-20200520004742-59133d7f0dd7
 	golang.org/x/sys v0.0.0-20210511113859-b0526f3d8744
 	golang.org/x/text v0.3.6 // indirect
+	gopkg.in/yaml.v2 v2.4.0
```
ptnapoleon, line 45: How do you feel about converging on `ghodss/yaml`? I chose it for `create` as that's the yaml library used in dd-go

vendor/modules.txt
```diff
@@ -189,7 +189,11 @@ github.com/inconshreveable/mousetrap
 github.com/json-iterator/go
 # github.com/konsorten/go-windows-terminal-sequences v1.0.3
 github.com/konsorten/go-windows-terminal-sequences
+<<<<<<< HEAD
```
ptnapoleon, line 192: Rebasing on `main` should clear these up

ptnapoleon (changes requested):
https://github.com/DataDog/chaos-controller/compare/philip/explain-review

Switching from `gopkg.in.yaml.v2` to `ghodss/yaml`, has a strong advantage of letting us re-use json tags, so you won't need to re-define all of the types. That's going to remove the need to redefine all of the types. That did require changing a few `if` checks.

This needs rebased onto `main` to get the other changes to the `chaosli` dir.

This needs rebased onto `main` to use the current form of the `NetworkDisruptionSpec`, which has the changes to `services` and `hosts`.

`Seperator` should be spelled `Separator`.

You were checking in multiple places for completely invalid options, e.g., missing required options. I argue that we shouldn't do that. Even if that survived unmarshaling, it would be better to simply call Validate() prior to explain, and print the list of errors for the user (alongside the explain, if possible). That way there's no duplication of validation logic, and all of your methods can assume they are explaining a valid spec.

vendor/modules.txt
ptnapoleon: Huh turns out it didn't. If you delete just the conflict lines, and then run `go mod vendor`, it should fix all of this

go.mod
ptnapoleon: I think this branch needs a `go mod tidy`

cli/chaosli/cmd/explain.go
```diff
@@ -7,26 +7,203 @@ package cmd
...
+	"path/filepath"
+	"strings"
 
+	"github.com/DataDog/chaos-controller/api/v1beta1"
+	chaostypes "github.com/DataDog/chaos-controller/types"
+	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
 )
 
+type Explanations struct {
```
ptnapoleon, line 21: This is beyond my `go` expertise, sadly, so I don't know which approach is more "idiomatic", but because this struct doesn't actually carry any state, I think you can just turn `separator` into a `const`, and make all of its methods into functions.

Azoam: I was thinking about treating it sort of like a static class and using encapsulation to keep everything in one struct. There could also be future data points and better testing working with the single struct versus importing a whole file. 

What are your thoughts on that, do you think its still weird to have it coded this way?

vendor/modules.txt
Azoam: I dont see this anymore on my branch?

go.mod
Azoam: Not too familiar with this, but when I ran it, no changes were registered and a force push does nothing.

ptnapoleon: Seems gone now? /shrug

vendor/modules.txt
ptnapoleon: Seems gone now? /shrug

cli/chaosli/cmd/explain.go
ptnapoleon: It is not weird to code it this way, it works just fine. I have a soft preference for keeping code as stateless as possible, and given there's no requirement to attach these functions to a struct, I would err on the side of not doing so. While, admittedly, I can foresee possible reasons to need state when explaining a yaml in possible futures, stateless code is "on average" more flexible to test and easier to read than stateful code. So on the principle of "yagni", I'd say it's better to drop the struct for now, but consider this a -0 at most.

cli/chaosli/cmd/explain.go
```diff
@@ -7,26 +7,242 @@ package cmd
...
+		fmt.Println(pathError)
+	}
+
+	fullPath, _ := filepath.Abs(filePath)
+	disruptionBytes, err := ioutil.ReadFile(fullPath)
+
+	if err != nil {
+		log.Printf("disruption.Get err   #%v ", err)
+		//panic(err)
+		return
```
ptnapoleon, line 228: Better to `os.Exit(1)` here, for the bad return code

cli/chaosli/cmd/explain.go
Azoam: yeah this makes sense, thanks for explaining. Lets change it.

cli/chaosli/cmd/explain.go
```diff
@@ -7,26 +7,245 @@ package cmd
...
+	fullPath, _ := filepath.Abs(filePath)
+	disruptionBytes, err := ioutil.ReadFile(fullPath)
+
+	if err != nil {
+		log.Printf("disruption.Get err   #%v ", err)
+		os.Exit(1)
 	}
 
-	return "Explanation TODO"
+	err = yaml.Unmarshal(disruptionBytes, &disruption)
```
ptnapoleon, line 234: It would be good to print out the validation output too whenever we call explain, wdyt?

ptnapoleon (approved):
Other than the question of if we should also validate when we explain (I think useful, imo? but definitely optional, esp. for this PR), and anything the linter is mad about, this is good to merge. While we can refine the exact text output forever, that can be in future PRs

cli/chaosli/cmd/explain.go
Azoam: Yes, debating if we should even run explain on invalid disruptions. Would save us a lot of headache in the explain methodology all together.

ptnapoleon (approved):
Discussed on slack how to make the linter happy, +1 after that

PR #325, opened 2021-06-01
CHAOS-85: Create all pods in chaos engineering

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

It needs linted, I'm not sure I hit everywhere in the docs, the feature definitely needs its own doc entry, but what do you think of the solution? If you had time to test outside of minikube tomorrow, that'd be swell

### Testing

### Checklist


controllers/disruption_controller.go
```diff
@@ -652,29 +648,27 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
 
 // getChaosPods returns chaos pods owned by the given instance and having the given labels
 func (r *DisruptionReconciler) getChaosPods(instance *chaosv1beta1.Disruption, ls labels.Set) ([]corev1.Pod, error) {
-	ownedPods := make([]corev1.Pod, 0)
 	pods := &corev1.PodList{}
+	var err error
+	if ls == nil {
+		ls = make(map[string]string)
+	}
+	// If this isn't a valid label, we will instead list ALL pods
```
ptnapoleon, line 656: We can be guaranteed that Namespace and Name are both legal substrings in a label value, but previously I was joining them with "/", which is _not_ okay in a label value.

examples/demo.yaml
```diff
@@ -8,6 +8,7 @@ apiVersion: v1
 kind: PersistentVolume
 metadata:
   name: demo
+  namespace: default
```
ptnapoleon, line 11: This is a fairly debatable change, it's good for us when testing locally, but not helpful for users who want to use this to demo the controller. Thoughts?

Devatoria: I would definitely keep that change if that's suitable for testing

controllers/disruption_controller_test.go
```diff
@@ -178,7 +168,7 @@ var _ = Describe("Disruption Controller", func() {
 	AfterEach(func() {
 		// delete disruption resource
 		_ = k8sClient.Delete(context.Background(), disruption)
-		Eventually(func() error { return expectChaosPod(disruption, 0) }, timeout).Should(Succeed())
+		Eventually(func() error { return expectChaosPod(disruption, 0) }, timeout*2).Should(Succeed())
```
Devatoria, line 171: Is this `*2` on timeout expected?

controllers/disruption_controller.go
```diff
@@ -652,12 +648,20 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
+		ls = make(map[string]string)
+	}
+
+	// If these aren't valid labels, we will instead list ALL pods
+	ls[chaostypes.DisruptionNameLabel] = instance.Name
+	ls[chaostypes.DisruptionNamespaceLabel] = instance.Namespace
+
+	r.log.Infow("Getting chaos pods with this label", "labels", ls.String())
+
 	// list pods in the instance namespace and for the given target
```
Devatoria, line 663:
```suggestion
	// list pods in the defined namespace and for the given target
```

controllers/disruption_controller.go
```diff
@@ -652,12 +648,20 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
 	listOptions := &client.ListOptions{
-		Namespace:     instance.Namespace,
```
Devatoria, line 660: Shouldn't we keep the namespace here but with `r.InjectorServiceAccountNamespace` instead? So we still scope the listing to a specific namespace (better to not be too harmful with apiserver I think)

controllers/disruption_controller.go
```diff
@@ -652,12 +648,20 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
 // getChaosPods returns chaos pods owned by the given instance and having the given labels
 func (r *DisruptionReconciler) getChaosPods(instance *chaosv1beta1.Disruption, ls labels.Set) ([]corev1.Pod, error) {
-	ownedPods := make([]corev1.Pod, 0)
 	pods := &corev1.PodList{}
 
+	if ls == nil {
+		ls = make(map[string]string)
+	}
+
+	// If these aren't valid labels, we will instead list ALL pods
```
Devatoria, line 657: This is creepy, do you have any concrete example of such a thing happening?

controllers/disruption_controller.go
```diff
@@ -652,12 +648,20 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
+	ls[chaostypes.DisruptionNameLabel] = instance.Name
+	ls[chaostypes.DisruptionNamespaceLabel] = instance.Namespace
+
+	r.log.Infow("Getting chaos pods with this label", "labels", ls.String())
```
Devatoria, line 661:
```suggestion
	r.log.Infow("getting chaos pods matching the disruption labels", "labels", ls.String())
```

controllers/disruption_controller.go
ptnapoleon: Good point

controllers/disruption_controller.go
ptnapoleon: Yes (this is the same bug we hit in December). If you do `disruption-namespaced-name=namespace/name`, the `/` character isn't valid, and it will list all pods. confirmed locally. But we can guarantee that `namespace` and `name` only have legal characters, so this is just a warning for the next person who changes this code

controllers/disruption_controller_test.go
ptnapoleon: Yes, I've seen the `afterEach` (and only that) not finish deleting within a single reconcile call, and then it pauses 9 seconds before trying again. This ends up failing the next test because of leaked pods. Not sure why this PR would be slower on delete

controllers/disruption_controller.go
```diff
@@ -652,12 +648,21 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
-	// list pods in the instance namespace and for the given target
+	if ls == nil {
+		ls = make(map[string]string)
+	}
+
+	// If these aren't valid labels, we will instead list ALL pods
+	ls[chaostypes.DisruptionNameLabel] = instance.Name
+	ls[chaostypes.DisruptionNamespaceLabel] = instance.Namespace
+
+	r.log.Infow("Searching for chaos pods with label selector...", "labels", ls.String())
```
Devatoria, line 661:
```suggestion
	r.log.Infow("searching for chaos pods with label selector...", "labels", ls.String())
```
Suggesting this because we have no capital letter on other lines of logs (definitely not the most useful suggestion)

Devatoria (approved):

PR #327, opened 2021-06-01
Fix the Validatecount experience

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

The chaosli currently always outputs spec.count as a string. It is reasonable, and technically possible, to pass in count as a string. Therefore, we should support specifying spec.count as a string, even when its an integer and not a % value.

### Testing

### Checklist

### DIFF
--- a/api/v1beta1/validations.go
+++ b/api/v1beta1/validations.go
@@ -73,7 +73,7 @@ func GetIntOrPercentValueSafely(intOrStr *intstr.IntOrString) (int, bool, error)
 func ValidateCount(count *intstr.IntOrString) error {
 	value, isPercent, err := GetIntOrPercentValueSafely(count)
 	if err != nil {
-		return err
+		return fmt.Errorf("error determining value of spec.count: %w", err)
 	}
 
 	if isPercent {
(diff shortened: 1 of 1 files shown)


api/v1beta1/validations.go
```diff
@@ -52,19 +52,19 @@ func GetIntOrPercentValueSafely(intOrStr *intstr.IntOrString) (int, bool, error)
 		return intOrStr.IntValue(), false, nil
 	case intstr.String:
 		s := intOrStr.StrVal
+		isPercent := false
 
 		if strings.HasSuffix(s, "%") {
```
Devatoria, line 57: `foo%` matches this condition but is not an expected percentage value, as well as `101%`. Do we consider that it is ok (the `foo%` will generate an error when casting the value to an integer anyway but the `101%` will still match) or should we validate that through a regex instead?

ptnapoleon: I consider that ok. `foo%` will get the right return value, imo `0, false, error`. `101%` is a valid percentage, just not for `count`. That's on the caller to care about (and our callers all do, i believe).

Devatoria (approved):

PR #335, opened 2021-06-09
Allow to apply network disruptions on init containers

Devatoria:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

This PR adds a mutating webhook in the admission webhooks which inserts an init container in pods having a specific label. This containers is inserted as the very first init container and prevents other containers from being created, waiting for a `SIGUSR1` signal to exit properly. Because the network namespace is shared across all containers within a pod, this container allows a disruption to be injected before any other containers are created in the targeted pod.

Motivation for change:

It allows us to apply a (network) disruption on containers before they are created and started so we can disrupt startup dependencies easily.

### Testing

### Checklist


webhook/chaos_handler.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type ChaosHandlerMutator struct {
+	Client  client.Client
+	Log     *zap.SugaredLogger
+	Image   string
+	decoder *admission.Decoder
+}
+
+func (m *ChaosHandlerMutator) InjectDecoder(d *admission.Decoder) error {
```
ptnapoleon, line 26: Who calls this? I wasn't able to find any usages

webhook/chaos_handler.go
```diff
@@ -0,0 +1,65 @@
...
+	m.decoder = d
+
+	return nil
+}
+
+func (m *ChaosHandlerMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	pod := &corev1.Pod{}
+
+	err := m.decoder.Decode(req, pod)
```
ptnapoleon, line 35: `m.decoder` can be nil here, should we check that?

cli/handler/main.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+var (
+	logger *zap.SugaredLogger
+
+	rootCmd = &cobra.Command{
+		Use:   "cobra",
+		Short: "A generator for Cobra based Applications",
```
ptnapoleon, line 24: Just a placeholder so we remember to change this

docs/features.md
```diff
@@ -32,6 +32,23 @@ How can you target a specific pod by name, if it doesn't have a unique label sel
...
+
+> :memo: This mode has some restrictions:
+>   * it only works for network related (network and dns) disruptions
+>   * it only works with the pod level
+>   * it does not support containers scoping (applying a disruption to only some containers)
+
+It can be handy to disrupt packets on pod initialization, meaning before containers are actually created and started, to test startup dependencies or init containers. You can do this in only two steps:
+
+* redeploy your pod with the specific label `chaos.datadoghq.com/disrupt-on-init` to hold it in the initialization state
+  * the chaos-controller will inject an init containers named `chaos-handler` in your pod
```
ptnapoleon, line 45:
```suggestion
  * the chaos-controller will inject an init containers name `chaos-handler` as the first init container in your pod
```

docs/features.md
```diff
@@ -32,6 +32,23 @@ How can you target a specific pod by name, if it doesn't have a unique label sel
...
+  * this init container is lightweight and does nothing but waiting for a `SIGUSR1` signal to complete successfully
+* apply your disruption [with the init mode on](../examples/on_init.yaml)
```
ptnapoleon, line 47: It works if the disruption is applied first, yes? As the disruption wont have any targets to pick yet and will wait for the applicable pods to appear

webhook/chaos_handler.go
```diff
@@ -0,0 +1,87 @@
...
+
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	// define pod name for logs
+	// if the pod is created from a replicaset (deployment, statefulset)
+	// the name won't be populated yet because it relies on the name generation
+	podName := pod.Name
+	if podName == "" {
+		podName = pod.ObjectMeta.GenerateName
```
ptnapoleon, line 57: This is only going to be a prefix of the full pod name. That's fine because we're only using it to log, afaict, but is that worth mentioning explicitly in the comments?

webhook/chaos_handler.go
```diff
@@ -0,0 +1,87 @@
...
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Args: []string{
+			"--timeout",
+			m.Timeout.String(),
+		},
+	}
+
+	// prepend chaos handler init container to already existing init containers
+	inits := []corev1.Container{}
+	inits = append(inits, init)
```
ptnapoleon, line 75: Why choose to do this over two lines rather than `inits = append(inits, init, pod.Spec.InitContainers...)`?

ptnapoleon (approved):

webhook/chaos_handler.go
Devatoria: `append` can't be used with this syntax in Go ([this answer](https://stackoverflow.com/questions/18946837/go-variadic-function-and-too-many-arguments) for more explanations). However I can indeed shorten this to a single line with another syntax, thanks for the heads up.

PR #336, opened 2021-06-09
Move demo resources to `chaos-demo` namespace by default

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [x] Improves Documentation

- Removes the specification of namespace in the example commands in the docs
- Moves all examples to the `chaos-demo` namespace

Motivation for change:

- We want to put our examples outside `chaos-engineering` for 2 reasons. To make it obvious that disruptions can be applied anywhere, not just in chaos-engineering, and to make running `demo.yaml` locally for testing exercise the codepaths related to operating in namespaces outside of `chaos-engineering`
- Users don't need it explained how to specify a namespace; the specification is redundant as all the examples specify a namespace; it clutters the command explanations

### Testing

### Checklist

### DIFF
--- a/examples/demo.yaml
+++ b/examples/demo.yaml
@@ -8,7 +8,7 @@ apiVersion: v1
 kind: PersistentVolume
 metadata:
   name: demo
-  namespace: default
+  namespace: chaos-demo
   labels:
     type: local
 spec:
@@ -24,7 +24,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: demo
-  namespace: default
+  namespace: chaos-demo
 spec:
   storageClassName: manual
   accessModes:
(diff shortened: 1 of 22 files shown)


examples/demo.yaml
```diff
@@ -8,7 +8,7 @@ apiVersion: v1
 kind: PersistentVolume
```
Devatoria, line 8: Can we add a namespace resource in the demo manifest so when we apply it, the `chaos-demo` namespace is created (and deleted when we delete it)? Otherwise it means we need to manually create it but as a user I'd expect the demo manifest to work without any additional steps than applying it.

ptnapoleon: I didn't even realize you could add namespaces via manifests, I've always done it manually. I'll add that

Issue #223, opened 2021-01-15
Inaccurate comment in node_failure?


noqcks:
Hi, quick question.

In node_failure.yaml the namespace metadata has a comment about a disruption resource needing to be in the same namespace as a targeted pod. I don't think this is accurate, because we're not targeting a pod, right? Or does chaos-controller require you to target pods with a label selector and then the nodes backing those pods are put into failure mode?
```
metadata:
  name: disruption-sample
  namespace: chaos-engineering # disruption resource must be in the same namespace as targeted pods
spec:
  selector: # label selector to target pods
    app: demo
```
For node level disruptions, can disruptions be run in any namespace?

Thanks, a little confused.

PR #345, opened 2021-06-17
Minor updates in docs.

nikos912000:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [x] Improves Documentation

No changes in the implementation, just minor improvements in the docs as I was going through them. In specific:
- Update node failure examples to reflect the level
- Update CRD so it can be installed using Helm
- Document how to install using Helm
- Minor typo fixes

Motivation for change: Docs make devs happy :p

### Testing

### Checklist


examples/node_failure.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   name: node-failure
   namespace: chaos-demo
 spec:
+  level: node # note that any labels need to be applied on the nodes for this Disruption to work
```
ptnapoleon, line 12: Actually, this change (and the one in the other node_failure example) are not correct. Your comment alludes to it, but `level` here really only affects the targeting of `selector`. `level: pod` is totally valid for a `node-failure` disruption kind. If you pull these two changes, I'll merge the rest of the doc fixes (which are really great, thank you!)

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
```diff
@@ -1,5 +1,7 @@
-
----
+# Unless explicitly stated otherwise all files in this repository are licensed
+# under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2021 Datadog, Inc.
```
ptnapoleon, line 4: Seems that CI is sad when this isn't verbatim the same file generated by `controller-gen`. I think on our end it's fine to not include the copyright header here. Could you take this change out as well? Thanks!

examples/node_failure.yaml
nikos912000: Hm, maybe I misunderstood how this works. I also read [this](https://github.com/DataDog/chaos-controller/issues/223) issue.

My assumption was the following:
- When not specifying the level it defaults to pod
- If I add a label to the pod the expectation is it will disrupt the node where the pod is located

Unless I did something wrong this didn't work. I only managed to disrupt nodes after setting the level and labelling them.

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
nikos912000: I can revert that, sure.
Unfortunately a Helm install doesn't work without that though. [This](https://github.com/DataDog/chaos-controller/blob/91c4fca8c49a289ac5ca6b6412749b7ce0cf73a8/chart/templates/crds/chaos.datadoghq.com_disruptions.yaml#L2) line causes issues.

examples/node_failure.yaml
ptnapoleon: Both of those assumptions are correct. It's interesting that it didn't work. Do you have any logs from the controller manager at that time?

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
ptnapoleon: Hmm, let me see what I can do w/ the CI check then

ptnapoleon: @Devatoria What are your thoughts on ^ ?

examples/node_failure.yaml
nikos912000: Right, I tested that again and I missed something.
I deploy the controller in the `chaos-engineering` namespace and my applications in a different namespace.
I was setting the namespace to the one where the controller gets deployed but apparently it needs to match the target pod's one.

I've reverted the change.

Devatoria: Just to add some context, the node failure is the only injector which does not use anything related to a container so the level only impacts the selector as @ptnapoleon said: https://github.com/DataDog/chaos-controller/blob/main/injector/node_failure.go

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
Devatoria: This file is autogenerated. The CI re-generates it and compares it to the committed one to be sure there are no diff. Here, either removing the `---` or adding a header will make a diff. I don't think we have a hand on this automatically added first line `---` anyway. Let me have a look at Helm to see why it would fail on a standard YAML content separator.

examples/node_failure.yaml
nikos912000: Makes sense @Devatoria 👍

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
Devatoria: Could you prove us your `helm` version please? I can't reproduce the error (but I have a lot of other errors related to metadata though).

nikos912000: I don't know what to say. I tried many times today and I didn't manage to reproduce either ¯\_(ツ)_/¯
I remember the CRD either not getting deployed or Helm showing an error (not a warning). But as I cannot reproduce anymore I have reverted the change.

ptnapoleon (approved):

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
Devatoria: Alright, thanks for the updates! Feel free to open an issue if the error pops up again at some point.

Devatoria (approved):

PR #359, opened 2021-07-01
Add a --handler-enabled flag to the controller

Devatoria:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

A brief description of changes to implementation or controller behavior: it adds a `--handler-enabled` flag to enable (or disable) the handler. When it is disabled, the admission controller will reject any disruption created with the `onInit: true`.

Motivation for change: the `onInit` function is not available on clusters with a Kubernetes version lower than 1.15 because you can't filter on objects caught by the admission webhook. This flag allows users to disable the feature if they want to run the controller on old clusters by rejecting requests trying to create disruptions using the feature.

### Testing

### Checklist

### DIFF
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -21,13 +21,15 @@ var logger *zap.SugaredLogger
 var k8sClient client.Client
 var metricsSink metrics.Sink
 var deleteOnly bool
+var handlerEnabled bool
 
-func (r *Disruption) SetupWebhookWithManager(mgr ctrl.Manager, l *zap.SugaredLogger, ms metrics.Sink, deleteOnlyFlag bool) error {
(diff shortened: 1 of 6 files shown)


api/v1beta1/disruption_webhook.go
```diff
@@ -44,6 +46,11 @@ func (r *Disruption) ValidateCreate() error {
 		return errors.New("the controller is currently in delete-only mode, you can't create new disruptions for now")
 	}
 
+	// handle a disruption using the onInit feature without the handler being enabled
+	if !handlerEnabled && r.Spec.OnInit {
+		return errors.New("the chaos handler is disabled but the disruption onInit field is set to true, please enable the handler by specifying the --handler-enabled flag to the controller if you want to use the onInit feature")
```
ptnapoleon, line 51: I'd argue this error message should help the user understand that (almost certainly) the handler is disabled because the current cluster is too old of a k8s version. Maybe appending something like "if you are creating this disruption on a k8s cluster < 1.15, then this is disabled intentionally as it is not supported"

ptnapoleon (approved):

PR #361, opened 2021-07-09
CHAOS-121: Handle deleting chaos pods before they finish injecting

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug
- [ ] Improves Documentation

1. Make our injector's exitHandler signal channel buffered
2. Increase the TerminationGracePeriodSeconds of the chaos pod spec

Motivation for change:

1.  signals needs to be a buffered channel, so that we can receive a signal anytime during the injection process as an unbuffered channel will ignore signals sent before it begins listening. We can keep the buffer size to `1` here as no matter how many SIGINT or SIGTERMs we receive, we want to carry out the same action: clean and die
2.  The default TerminationGracePeriodSeconds is 30s. This can be too low for a chaos pod to finish cleaning. After TGPS passes, the signal sent to a pod becomes SIGKILL, which will interrupt any in-progress cleaning. Setting this to a generous 2 minutes in the pod spec itself, ensures that whether a chaos pod is deleted directly or by deleting a disruption, it will have time to finish cleaning up after itself.

### Testing

Proper testing would need a real cluster integration test. Locally, I have tested the following case before and after the patch, to ensure we are fixing the bug:

- Add a `time.Sleep(time.Minute)` immediately before `sig := <-signals` in `injector/main.go#injectAndWait` to ensure that I could definitely issue a delete before a chaos pod begins listening for signals
- Create a disruption
- After the disruption has started and begun injecting, but before all chaos pods are ready and listening for signals, delete the disruption
- Wait to see if the disruption is stuck on removal

### Checklist


controllers/disruption_controller.go
```diff
@@ -673,11 +673,17 @@ func (r *DisruptionReconciler) generatePod(instance *chaosv1beta1.Disruption, ta
 	hostPathDirectory := corev1.HostPathDirectory
 	hostPathFile := corev1.HostPathFile
 
+	// The default TerminationGracePeriodSeconds is 30s. This can be too low for a chaos pod to finish cleaning. After TGPS passes,
+	// the signal sent to a pod becomes SIGKILL, which will interrupt any in-progress cleaning. Setting this to a generous 2 minutes in the pod spec itself,
+	// ensures that whether a chaos pod is deleted directly or by deleting a disruption, it will have time to finish cleaning up after itself.
+	terminationGracePeriod := int64(120)
```
Devatoria, line 679: As discussed together, can we pass a context to the `Inject` method of injectors so we can cancel the context if we receive the signal during the injection and reduce that termination grace period?

injector/cpu_pressure.go
```diff
@@ -141,8 +143,20 @@ func (i cpuPressureInjector) Inject() error {
...
+
+	go func() {
+		wg.Wait()
+		close(waitCh)
+	}()
+
 	// wait for stress routines to finish initializing
-	wg.Wait()
+	select {
+	case <-ctx.Done():
```
ptnapoleon, line 155: Here we will exit either once all the stressors are initialized or once we're cancelled. I could also add a check in the loop where we create initializers

injector/disk_pressure.go
```diff
@@ -63,9 +64,9 @@ func NewDiskPressureInjector(spec v1beta1.DiskPressureSpec, config DiskPressureI
 	}, nil
 }
 
-func (i diskPressureInjector) Inject() error {
+func (i diskPressureInjector) Inject(ctx context.Context) error {
 	// add read throttle
-	if i.spec.Throttling.ReadBytesPerSec != nil {
+	if ctx.Err() != nil && i.spec.Throttling.ReadBytesPerSec != nil {
```
ptnapoleon, line 69: Only two operations so easy to just check if we're cancelled before both

injector/dns_disruption.go
```diff
@@ -60,7 +61,7 @@ func NewDNSDisruptionInjector(spec v1beta1.DNSDisruptionSpec, config DNSDisrupti
 }
 
 // Inject injects the given dns disruption into the given container
-func (i DNSDisruptionInjector) Inject() error {
+func (i DNSDisruptionInjector) Inject(ctx context.Context) error {
```
ptnapoleon, line 64: I could not find anywhere good to cancel in the middle of this.

injector/network_disruption.go
```diff
@@ -68,7 +68,7 @@ func NewNetworkDisruptionInjector(spec v1beta1.NetworkDisruptionSpec, config Net
 }
 
 // Inject injects the given network disruption into the given container
-func (i networkDisruptionInjector) Inject() error {
+func (i networkDisruptionInjector) Inject(ctx context.Context) error {
```
ptnapoleon, line 71: Wanted to seek your advice on where in `applyOperations` to check for cancellation and end early

injector/node_failure.go
```diff
@@ -57,14 +58,21 @@ func NewNodeFailureInjector(spec v1beta1.NodeFailureSpec, config NodeFailureInje
...
+func (i nodeFailureInjector) Inject(ctx context.Context) error {
 	var err error
 
 	i.config.Log.Infow("injecting a node failure by triggering a kernel panic",
 		"sysrq_path", i.sysrqPath,
 		"sysrq_trigger_path", i.sysrqTriggerPath,
 	)
 
+	select {
+	case <-ctx.Done():
```
ptnapoleon, line 70: This is inconsistent to everywhere else, lmk if I should stick with `select` everywhere or `if`. `select` appears to be more idiomatic (and I think clearer to read as "Done" is more sensible than "Err"), but the `if` is smaller.

ptnapoleon:
Other than needing feedback on our preferred "if cancelled; then return" syntax, and more manual testing, this is ready for review

ptnapoleon:
Since implementing cancel, the original bug has re-appeared, i'm investigating

controllers/disruption_controller.go
ptnapoleon: After implementation, I decided this was a huge amount of code+complexity for very little perf gain, so I'd like to compromise at just a 1 minute grace period

Devatoria (approved):

PR #363, opened 2021-07-16
Add a way to exclude hosts from network disruptions

Devatoria:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

A brief description of changes to implementation or controller behavior: it adds two ways to exclude hosts from a network disruption (from the disruption itself, or from the controller so it is applied for all network disruptions).

Motivation for change: we sometimes need to disrupt a whole CIDR except one or two IPs which is not doable currently.

### Testing

### Checklist

### DIFF
--- a/api/disruption_kind.go
+++ b/api/disruption_kind.go
@@ -20,8 +20,8 @@ type DisruptionKind interface {
 	Validate() error
 }
 
-// AppendCommonArgs is a helper function generating common args and appending them to the given args array
-func AppendCommonArgs(args []string, level chaostypes.DisruptionLevel, kind chaostypes.DisruptionKindName, containerIDs []string, sink string, dryRun bool,
+// AppendArgs is a helper function generating common and global args and appending them to the given args array
+func AppendArgs(args []string, level chaostypes.DisruptionLevel, kind chaostypes.DisruptionKindName, containerIDs []string, sink string, dryRun bool,
 	disruptionName string, disruptionNamespace string, targetName string, onInit bool, allowedHosts []string) []string {
 	args = append(args,
 		// basic args
--- a/api/disruption_kind.go
+++ b/api/disruption_kind.go
@@ -21,8 +21,8 @@ type DisruptionKind interface {
 }
 
 // AppendCommonArgs is a helper function generating common args and appending them to the given args array
-func AppendCommonArgs(args []string, level chaostypes.DisruptionLevel, containerIDs []string, sink string, dryRun bool,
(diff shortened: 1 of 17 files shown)


api/disruption_kind.go
```diff
@@ -45,5 +45,12 @@ func AppendCommonArgs(args []string, level chaostypes.DisruptionLevel, container
 		args = append(args, "--on-init")
 	}
 
+	// append allowed hosts for network disruptions
+	if kind == chaostypes.DisruptionKindNetworkDisruption {
```
ptnapoleon, line 49: There's a bit of a code smell putting a kind specific arg here under `AppendCommonArgs`. I see why you had to do it, but I would almost rather add `--allowed-hosts` as a flag to the injector root command, and then just pass this into all injection kinds, keeping this function true to its needs. I'll try to think of a better solution while I read the code

EDIT: After thinking I really couldn't find a better idea. my suggestion ^ isn't good, so perhaps we just rename this function and change its comment, leaving your implemented behavior?

examples/demo.yaml
```diff
@@ -60,6 +60,12 @@ spec:
       containers:
       - name: nginx
         image: nginx
+        livenessProbe:
```
ptnapoleon, line 63: Why add this livenessProbe?

injector/network_disruption.go
```diff
@@ -389,6 +341,11 @@ func (i *networkDisruptionInjector) applyOperations() error {
 		}
 	}
 
+	// add filters for allowed hosts
+	if err := i.addFiltersForHosts(interfaces, i.spec.AllowedHosts, "1:1"); err != nil {
+		return fmt.Errorf("error adding filter for alloed hosts: %w", err)
```
ptnapoleon, line 346:
```suggestion
		return fmt.Errorf("error adding filter for allowed hosts: %w", err)
```

ptnapoleon (approved):
+1 with a suggestion

examples/demo.yaml
Devatoria: I think it is helpful for local testing to ensure a disruption does not (or does if this is what's expected) disrupt liveness probes. A concrete example is applying a 100% drop network disruption on a pod should not drop packets sent for probing.

api/disruption_kind.go
Devatoria: I agree. I changed the function name and comment, let me know if it sounds ok now.

ptnapoleon (approved):

PR #364, opened 2021-07-16
Adding context feature to chaosli

Azoam:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug
- [X] Improves Documentation

Adds the context features to chaosli which will give users a bit of context in regards to their targets. It shares conditions/phases/statues of target resources to give users a better idea of the impact of the intended disruption.

Motivation for change:

Context is extremely important before running disruptions.  A user should be doing what this feature does by default before running any disruption, but this just makes it easier for users to get the jist of their environments.

### Testing

Tested with several target resources.

### Checklist


cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const MAXTARGETSHOW = 10
+const POD = "pod"
```
ptnapoleon, line 23: Rather than re-defining this, its better to import `chaostypes.DisruptionLevelPod` and use that for comparison. Same with `"node"` which is used at L34

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
+package cmd
+
+import (
+	"encoding/json"
```
ptnapoleon, line 4: imports need sorted

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+		printPodStatus(podsData)
+		printContainerStatus(podsData)
+	} else {
+		printNodeStatus(nodesData)
+	}
+
+}
+
+func init() {
+	contextCmd.Flags().String("path", "", "The path to the disruption file to be contextualized.")
```
ptnapoleon, line 392: The extended description and/or help text of this command should explain to the user that we're relying on them already having a working `kubectl` connection to a k8s cluster, and that the context we'll be producing is based on their current active `kubectx`

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+
+	fmt.Println("Let's look at your targets...")
+
+	s := spinner.New(spinner.CharSets[38], 100*time.Millisecond)
+	s.Start()
+	cmd := fmt.Sprintf("kubectl get %s -n %s -l %s | wc -l", level, podNamespaces, labels)
+	sizeString, err := exec.Command("bash", "-c", cmd).Output()
+
+	if err != nil {
+		return nil, fmt.Errorf("Could not count the number of pods correlating to target selector: %v", err)
```
ptnapoleon, line 46: I think this error message should suggest to users that they confirm that `kubectl` to a cluster is working for them, and to make sure their `kubectx` is set correctly

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	size, err := strconv.Atoi(strings.Trim(string(sizeString), "\n"))
+
+	if err != nil {
+		return nil, fmt.Errorf("Could not convert string to integer in context target size: %v", err)
```
ptnapoleon, line 51: My ide's linter is upset in a few places in this file because golang errors should not be capitalized at the beginning

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+		if level == POD {
+			newLevel = "node"
+		} else {
+			newLevel = POD
+		}
+		cmd := fmt.Sprintf("kubectl get %s -n %s -l %s | wc -l", newLevel, podNamespaces, labels)
+		sizeString, err := exec.Command("bash", "-c", cmd).Output()
+
+		if err != nil {
+			return nil, fmt.Errorf("Could not count the number of pods correlating to target selector: %v", err)
```
ptnapoleon, line 72: This isn't always pods being counted

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+	if size <= 0 {
+		errorString := fmt.Sprintf("\nThe label selectors chosen (%s) result in 0 targets, meaning this disruption would do nothing given the namespace/cluster/label combination.", labels)
+		newLevel := ""
+
+		if level == POD {
+			newLevel = "node"
+		} else {
+			newLevel = POD
+		}
+		cmd := fmt.Sprintf("kubectl get %s -n %s -l %s | wc -l", newLevel, podNamespaces, labels)
```
ptnapoleon, line 68: Looks like this code is repeated verbatim twice, it might be worth pulling it out into a function

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+
+		if err != nil {
+			return nil, fmt.Errorf("Could not convert string to integer in context target size: %v", err)
+		}
+
+		if size > 0 {
+			errorString = fmt.Sprintf("\nWe noticed that your target size is 0 for level %s given your label selectors. We checked to see if the %s level would give you results and we found %d %ss. Is this the level you wanted to use?", level, newLevel, size, newLevel)
+		}
+		fmt.Println(errorString)
+		return nil, fmt.Errorf("bad sized targets")
```
ptnapoleon, line 84: this error string could be improved

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+			targetsShow = append(targetsShow, targetsSplit[i])
+		}
+		if targetsSplit[i] == "NAME" || targetsSplit[i] == "" {
+			continue
+		}
+		targetsAll = append(targetsAll, targetsSplit[i])
+	}
+
+	fmt.Printf("\n🎯 There are %d %ss that will be targeted\n", size, level)
+	fmt.Println("📝 Here are a small set of those targets:")
```
ptnapoleon, line 111: Should this message be dynamic? Sometimes (probably often, even), we'll be showing all of the targets

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+	fullPath, err := filepath.Abs(path)
+
+	if err != nil {
+		log.Fatalf("Finding Absolute Path: %v", err)
+	}
+
+	disruptionPath, err := os.Open(filepath.Clean(fullPath))
+
+	if err != nil {
+		log.Fatalf("Openning Yam: %v", err)
```
ptnapoleon, line 343:
```suggestion
		log.Fatalf("could not open yaml: %v", err)
```

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+
+	err = disruption.Spec.Validate()
+
+	if err != nil {
+		log.Fatalf("There were some problems when validating your disruption: %v", err)
+	}
+
+	err = checkKubectl()
+	if err != nil {
+		log.Fatalf("Could not find/use kubectl command, make sure it is in your PATH variable and that all authorizations for the command are set (login to your authorization provider (e.g. AppGate).")
```
ptnapoleon, line 367: You'll want to change "AppGate" to VPN here, so this is generalizable to any user

cli/chaosli/cmd/context.go
ptnapoleon: EDIT: nvm I see we do `checkKubetctl` before this function is called

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+
+	fmt.Println(percentInfo)
+	PrintSeparator()
+}
+
+func checkKubectl() error {
+	cmd := exec.Command("kubectl", "get", "pods", "-n", "chaos-engineering")
+	_, err := cmd.Output()
+
+	if err != nil {
```
ptnapoleon, line 310: You can just directly `return err`

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		path, _ := cmd.Flags().GetString("path")
+		contextualize(path)
+	},
+}
+
+func contextualize(path string) {
+	var disruption v1beta1.Disruption
+
+	fullPath, err := filepath.Abs(path)
```
ptnapoleon, line 334: The "read file, unmarshal, validate" logic is now repeated thrice, consider extracting this into a function?

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,393 @@
...
+
+		for _, status := range node.Status.Conditions {
+			info = info +
+				"\t\t⭕️ Type: " + string(status.Type) + "\n" +
+				"\t\tℹ️  Status: " + string(status.Status) + "\n" +
+				"\t\t🤨 Reason: " + status.Reason + "\n\n"
+			percentCollectCondition[string(status.Type)+"/"+string(status.Status)] += 1.0 / float64(len(targetsInfo))
+		}
+
+		if i < MAXTARGETSHOW {
```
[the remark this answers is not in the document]
ptnapoleon: (ditto above where this is repeated)

ptnapoleon (changes requested):
Left some initial thoughts, I'm gonna experiment with running and using this after meetings today

cli/chaosli/cmd/context.go
Azoam: Can you explain, goland sorts and places the imports for me, unsure how to further sort (also tried to apply gofmt with no change).

cli/chaosli/cmd/context.go
Azoam: I have the actual error string printed before the error fires. I didnt want to put errorString in the Errorf because it becomes a bit unclear with all the error string nesting. Do you think "bad sized targets" should be changed considering errorString has more informative information?

cli/chaosli/cmd/context.go
Azoam: Dynamic how so?

cli/chaosli/cmd/context.go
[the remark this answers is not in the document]
Azoam: Let me know if I misunderstood; But restricting these loops would only get us a handful of targets to run percentages on, but we need to run the percentage map on every target, hence targetsInfo. `MAXTARGETSHOW` is only used for printing out the small set for users to checkout.

cli/chaosli/cmd/context.go
ptnapoleon: A user running the context cmd is going to see
```
$printed_errorstring
Could not grab context regarding size and names of pods: bad sized targets
```
Why print errorstring separately rather than just returning error string?

cli/chaosli/cmd/context.go
ptnapoleon: As in only printing this string if we really have truncated the list

cli/chaosli/cmd/context.go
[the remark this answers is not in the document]
ptnapoleon: Ahh wow im totally wrong, i missed this

cli/chaosli/cmd/context.go
ptnapoleon: This does look like the default `gofmt` rather than `goimports`, which appears to be the convention we use looking at `disruption_controller.go` where imports are split into three groups: stdlib, external packages, local packages

cli/chaosli/cmd/context.go
Azoam: Thats fair, ok I'll just return the errorString.

cli/chaosli/cmd/context.go
Azoam: Oh right, good point.

cli/chaosli/cmd/context.go
Azoam: got it, thanks!

cli/chaosli/cmd/root.go
```diff
@@ -86,34 +88,56 @@ func initConfig() {
...
+	}
+
+	return nil
+}
+
+func PrintSeparator() {
+	fmt.Println("=======================================================================================================================================")
+}
+
+func ReadUnmarshallValidate(path string) (v1beta1.Disruption) {
```
ptnapoleon, line 112: should be `Unmarshal`?

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,385 @@
...
+	v1 "k8s.io/api/core/v1"
+)
+
+const MAXTARGETSHOW = 10
+
+func contextTargetsSize(disruption v1beta1.Disruption) ([]string, error) {
+	spec := disruption.Spec
+	labels := spec.Selector.String()
+	level := spec.Level
+	// If node level, this \n will give us empty namespaces for node which is intended because nodes
```
ptnapoleon, line 26: Is "\n" what we'll get for `ObjectMeta.Namespace` if level is node? Sorry, I'm just confused by this comment

Azoam: this was an old comment when I was slicing out the namespace, it is now no longer a thing because I get the namespace from the disruption so I'm deleting this comment, thanks for the catch.

cli/chaosli/cmd/root.go
```diff
@@ -86,34 +88,56 @@ func initConfig() {
...
 
 func DisruptionFromFile(path string) (v1beta1.Disruption, error) {
-	yaml, err := os.Open(filepath.Clean(path))
+	parsedSpec := ReadUnmarshalValidate(path)
+
+	return parsedSpec, nil
+}
+
+func DisruptionToFile(path string, spec v1beta1.DisruptionSpec) error {
+	return nil
```
ptnapoleon, line 97: Could this return some sort of UnimplementedError?

ptnapoleon: Nvm, I see this is just entirely unused. Could it be removed then?

cli/chaosli/cmd/root.go
```diff
@@ -86,34 +88,56 @@ func initConfig() {
...
+}
+
+func validatePath(filePath string) error {
```
ptnapoleon, line 100: The only place this is used is to make sure the `path` flag is set for certain commands. cobra has a better way to do this. In the `init()` function of each command, you can do `contextCmd.MarkFlagRequired("path")` (or whatever the cmd variable is)

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+		printPodStatus(podsData)
+		printContainerStatus(podsData)
+	} else {
+		printNodeStatus(nodesData)
+	}
+}
+
+func init() {
+	contextCmd.Flags().String("path", "", "The path to the disruption file to be contextualized.")
+	contextCmd.Println("This flag requires that you are connected to a cluster and kubectl works as intended on that cluster. All the results of this flag will be based on your current kubectx.")
```
ptnapoleon, line 387:
```suggestion
	contextCmd.Println("This command requires that you are connected to a cluster and kubectl works as intended on that cluster. All the results of this command will be based on your current kubectx.")
```

This is referring to `chaosli context`, yes? I think "command" is a better noun to use then

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+
+func contextTargetsSize(disruption v1beta1.Disruption) ([]string, error) {
+	spec := disruption.Spec
+	labels := spec.Selector.String()
+	level := spec.Level
+	podNamespaces := disruption.ObjectMeta.Namespace
+	rowNames := 1
+
+	if level == types.DisruptionLevelNode {
+		rowNames = 1
```
ptnapoleon, line 35: Afaict this value is never set to anything other than `1`, so I think you can remove this `if` block

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	// Remove header NAME from consideration
+	size--
+	// If the size is 0, first check if changing the level will do anything, otherwise
+	// mention to the user that the labels they are using won't target anything
+
+	if size <= 0 {
+		errorString := fmt.Sprintf("\nThe label selectors chosen (%s) result in 0 targets, meaning this disruption would do nothing given the namespace/cluster/label combination.", labels)
+		newLevel := ""
```
ptnapoleon, line 56: `newLevel` never gets set to anything other than this empty string

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+	size, err := strconv.Atoi(strings.Trim(string(sizeString), "\n"))
+
+	if err != nil {
+		return -1, fmt.Errorf("could not convert string to integer in context target size: %v", err)
+	}
+
+	return size, nil
+}
+
+func grabDataForTargets(targets []string, disruption v1beta1.Disruption) ([]v1.Pod, []v1.Node, error) {
```
ptnapoleon, line 141: @Devatoria whats the idiomatic way (if you know it) in Go to return A XOR B in this fashion?

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+
+func printContainerStatus(targetInfo []v1.Pod) {
+	percentCollect := make(map[string]float64)
+
+	fmt.Println("\nLets take" +
+		" a look at the status of your Targeted Pod Containers...")
+
+	totalContainers := 0
+
+	for i := 0; i < len(targetInfo); i++ {
```
ptnapoleon, line 188: I think both these for loops can be shrunken slightly by using the `for idx, element := range $list` syntax

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+
+	PrintSeparator()
+
+	return targetsAll, nil
+}
+
+func getTargetSize(disruption v1beta1.Disruption) (int, error) {
+	level := disruption.Spec.Level
+	podNamespace := disruption.Namespace
+	labels := disruption.Labels
```
ptnapoleon, line 124: Small bug here, this should be the disruption's `Selector` (or some form of it). You're grabbing the labels on the disruption itself, which there aren't any. To test, I print the `cmd` we generated and see it is `kubectl get node -n datadog -l map[] | wc -l` for a disruption with 
```
  selector:
    service: zookeeper
```

ptnapoleon (changes requested):
Left a few comments, and one bug

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+		if targetsSplit[i] == "NAME" || targetsSplit[i] == "" {
+			continue
+		}
+
+		targetsAll = append(targetsAll, targetsSplit[i])
+	}
+
+	fmt.Printf("\n🎯 There are %d %ss that will be targeted\n", size, level)
+
+	if size >= MAXTARGETSHOW {
```
ptnapoleon, line 104:
```suggestion
	if size > MAXTARGETSHOW {
```

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,388 @@
...
+		fmt.Println("📝 Here are a small set of those targets:")
+	}
+
+	for _, target := range targetsShow {
+		fmt.Println(target)
+	}
+
+	if size >= MAXTARGETSHOW {
```
ptnapoleon, line 112:
```suggestion
	if size > MAXTARGETSHOW {
```

cli/chaosli/cmd/context.go
Azoam: yeah that sounds a lot better.

cli/chaosli/cmd/context.go
Azoam: yes another remnant of when I didnt have the namespace value on hand and the node and pod names were on different rows.

cli/chaosli/cmd/context.go
Azoam: yes another remnant of when I didnt have the namespace value on hand and the node and pod names were on different rows.

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,385 @@
...
+
+		if size > 0 {
+			errorString = fmt.Sprintf("\nWe noticed that your target size is 0 for level %s given your label selectors. We checked to see if the %s level would give you results and we found %d %ss. Is this the level you wanted to use?", level, disruption.Spec.Level, size, disruption.Spec.Level)
+		}
+
+		return nil, fmt.Errorf(errorString)
+	}
+
+	cmd := fmt.Sprintf("kubectl get %s -n %s -l %s | awk '{print $1}'", level, podNamespaces, labels)
+	targets, err := exec.Command("bash", "-c", cmd).Output()
```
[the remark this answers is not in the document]
ptnapoleon: oh i didn't realize the go client could do that. Does it handle oidc auth for us? if yes, then that is a much better solution

cli/chaosli/cmd/context.go
ptnapoleon: I think the current return value works fine then, though the function could benefit from a comment explaining that only one slice will be populated

cli/chaosli/cmd/explain.go
```diff
@@ -290,4 +246,9 @@ func explanation(path string) {
 
 func init() {
 	explainCmd.Flags().String("path", "", "The path to the disruption file to be explained.")
+	err := explainCmd.MarkFlagRequired("path")
+
+	if err != nil {
```
ptnapoleon, line 251: What kind of errors can happen here? Should a message be printed to the user?

Azoam: the error would be if the required path flag is not given. The MarkFlagRequired function sends out the error message on error with the help command as well, so it should be clear to the user what is wrong when that function errors.

cli/chaosli/cmd/root.go
```diff
@@ -86,34 +89,48 @@ func initConfig() {
...
 	if err != nil {
-		return v1beta1.Disruption{}, err
+		log.Fatalf("could not open yaml: %v", err)
 	}
 
 	yamlBytes, err := ioutil.ReadAll(yaml)
 	if err != nil {
-		return v1beta1.Disruption{}, err
+		log.Printf("disruption.Get err   #%v ", err)
+		os.Exit(1)
```
ptnapoleon, line 119: Why not use `log.Fatal` here as well? Is the exitcode for a fatal log not `1`?

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,423 @@
...
+		printContainerStatus(pods)
+	} else {
+		printNodeStatus(nodes)
+	}
+}
+
+func init() {
+	contextCmd.Flags().String("path", "", "The path to the disruption file to be contextualized.")
+	contextCmd.Flags().String("kubeconfig", "", "The path to your kube configuration directory (.../.kube/config). defaults to ~/.kube/config.")
+	contextCmd.Flags().Bool("verbose", false, "If set, will describe a small set of 10 of your targets. Otherwise it only describes percentages of the group of targets in total.")
```
ptnapoleon, line 413: The help text feels slightly confusing? I am surprised to hear that "verbose" leads to describing a small set of targets, though I do see how that works. It does seem that `10` here is wrong, and it's really the value of maxtargetshow?

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,423 @@
...
+		if size > 0 {
+			errorString = fmt.Sprintf("\nWe noticed that your target size is 0 for level %s given your label selectors. We checked to see if the %s level would give you results and we found %d %ss. Is this the level you wanted to use?", level, disruption.Spec.Level, size, disruption.Spec.Level)
+		}
+
+		return nil, nil, fmt.Errorf(errorString)
+	}
+
+	var allPods *[]v1.Pod
+
+	var allNodes *[]v1.Node
```
ptnapoleon, line 66: why use a pointer to a slice rather than just a slice? pointers default to nil, but slices default to empty, i think?

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,423 @@
...
+
+	if level == types.DisruptionLevelPod {
+		pods := getPods(disruption)
```
ptnapoleon, line 69: It looks like we're going to be querying kubernetes twice, minimum (here and above at getTargetSize). It would be more efficient to reduce that to just one call per level, and it prevents us from hitting the situation where the size changes

cli/chaosli/cmd/root.go
Azoam: log.Fatal should be fine here and good for consistency as well.

cli/chaosli/cmd/context.go
Azoam: Yeah my thought process behind verbose was that if people are running disruptions on tons of targets (which is usually the case), they don't really want to see individual targets information but what the cluster looks like as a whole. I'll change the comment to specify that 5 is the default which assumes you can change it. Let me know if there is any other confusion in the help text.

cli/chaosli/cmd/context.go
Azoam: I am using pointers so that I can return nil on error, I dont think I can return nil if it is a slice, but do you think it would just be better to use slices and return empty slices on error instead?

cli/chaosli/cmd/context.go
Azoam: great point!

cli/chaosli/cmd/context.go
ptnapoleon: i would think it's better to return an empty slice + error on error? just easier to let later code iterate through a slice w/o checking if it's nil

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	PrintSeparator()
+
+	return allPods, allNodes, nil
+}
+
+func showPods(pods v1.PodList) []v1.Pod {
+	targetsShow := []string{}
+
```
Devatoria, line 92:
```suggestion
```

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	options := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(disruption.Spec.Selector).String(),
+	}
```
Devatoria, line 175: Those lines are common to `getPods` and `getNodes` and should probably be part of the [command initialization](https://pkg.go.dev/github.com/spf13/cobra#OnInitialize).

We should also avoid `panic()` especially on errors which seems could be triggered easily (if the k8s configuration is not present, not correct, etc.) and should display a nice to read message instead so users are not lost in a stacktrace.

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,432 @@
...
+	pods, err := clientset.CoreV1().Pods(disruption.ObjectMeta.Namespace).List(context.TODO(), options)
+
```
Devatoria, line 177:
```suggestion
```

cli/chaosli/cmd/context.go
```diff
@@ -0,0 +1,432 @@
...
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	options := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(disruption.Spec.Selector).String(),
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(context.TODO(), options)
+
```
Devatoria, line 200:
```suggestion
```

Devatoria (changes requested):
A bunch of suggestions and things to change. Overall, I can see a lot of calls to `panic()` or `log.Fatal()`. Both of them will print a stacktrace of the error which is not user friendly at all to read. When working on CLI, we should avoid them as much as possible and display a nice-to-read message as much as possible so users can easily know what's happening.

cli/chaosli/cmd/context.go
Azoam: Im not sure if OnInitialize is what we want to use because its in the root.go file and Im not sure how to nicely bring kubeconfig into context.go. But I added a setKubeconfig function in the context.go file that aggregates all configuration setup for kubeconfig. Let me know what you think.

Devatoria (approved):

cli/chaosli/cmd/context.go
Devatoria: LGTM

PR #373, opened 2021-07-27
Allow advanced target selectors in disruptions

Devatoria:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

A brief description of changes to implementation or controller behavior: it adds the possibility to specify advanced label selectors which can select targets with something else than the regular key/value exact match. Possible operators are:
* `Exists` to select targets having a specific key
* `DoesNotExist` to select targets not having a specific key
* `In` to select targets having a key with a value matching a set of values
* `NotIn` to select targets having a key with a value not matching a set of values

Note that the advanced selector itself fills the current selector use case of matching an exact key/value but I decided to keep both fields for multiple reasons:
* the obvious one is for compatibility issues as format between both selectors are totally different
* the not obvious one is that specify an exact key/value match would be way more complex with advanced selectors (more complex than it should be at least)
  * as the exact key/value match is 99% of our use cases, I'd prefer to keep it simple and let advanced users using the advanced selectors when needed

Here's what advanced selectors look like:
```yaml
kind: Disruption
metadata:
  name: advanced-selector
  namespace: chaos-demo
spec:
  advancedSelector:
    - key: app
      operator: Exists
    - key: app
      operator: DoesNotExists
    - key: app
      operator: In
      values:
        - curl
    - key: app
      operator: NotIn
      values:
        - nginx
```

Motivation for change: we sometimes need to select targets having a specific key with potentially different values, or which does not have a specific key, etc. This feature allows more flexibility in target selection and mixes well with the existing selector.

### Testing

### Checklist

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -39,10 +39,12 @@ import (
 type DisruptionSpec struct {
 	// +kubebuilder:validation:Required
 	Count *intstr.IntOrString `json:"count"` // number of pods to target in either integer form or percent form appended with a %
-	// +kubebuilder:validation:Required
-	Selector labels.Set `json:"selector"`         // label selector
-	DryRun   bool       `json:"dryRun,omitempty"` // enable dry-run mode
-	OnInit   bool       `json:"onInit,omitempty"` // enable disruption on init
+	// +nullable
+	Selector labels.Set `json:"selector,omitempty"` // label selector
+	// +nullable
+	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
+	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
+	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
 	// +kubebuilder:validation:Enum=pod;node;""
 	Level      chaostypes.DisruptionLevel `json:"level,omitempty"`
 	Containers []string                   `json:"containers,omitempty"`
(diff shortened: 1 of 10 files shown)


api/v1beta1/disruption_types.go
```diff
@@ -39,10 +39,12 @@ import (
...
 	// +kubebuilder:validation:Required
 	Count *intstr.IntOrString `json:"count"` // number of pods to target in either integer form or percent form appended with a %
-	// +kubebuilder:validation:Required
-	Selector labels.Set `json:"selector"`         // label selector
-	DryRun   bool       `json:"dryRun,omitempty"` // enable dry-run mode
-	OnInit   bool       `json:"onInit,omitempty"` // enable disruption on init
+	// +nullable
+	Selector labels.Set `json:"selector,omitempty"` // label selector
+	// +nullable
+	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
```
ptnapoleon, line 45: ping for @nathan-tournant to think about if we can represent the relationship between `Selector` and `AdvancedSelector` as a ddmark

examples/complete.yaml
```diff
@@ -13,7 +13,20 @@ spec:
   level: pod # level the disruption should be injected at (can be either pod or node, defaults to pod)
   selector: # label selector[s] to target pods
     app: demo
-    team: developers
+    team: developer
+  advancedSelector: # advanced selectors can select targets on something else than an exact key/value match
+    - key: app
+      operator: Exists
+  # - key: app
+  #   operator: DoesNotExists
```
ptnapoleon, line 21: Should this be pluralized? It is not in the PR description. Do both options work?

ptnapoleon: Okay I looked in the code, this is a typo in both examples

ptnapoleon (approved):
Looks great! Just one typo in the examples (and of course the incoming tests)

examples/complete.yaml
Devatoria: Oops indeed, thanks! Copy/paste went bad.

Azoam (approved):
LGTM, really nice addition!

PR #389, opened 2021-08-11
Ignore targets being already affected by another disruption

Devatoria:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

A brief description of changes to implementation or controller behavior: already injected targets are now ignored to avoid stacking disruptions (which can lead to unexpected behaviours and errors).

Motivation for change: avoid to trigger strange errors when stacking disruptions (for instance network disruption can't be stacked or it would error).

### Testing

- Apply any disruption on the demo set
- Wait for injection to happen
- Apply any other disruption on the demo set

The second disruption should not be injected unless the first one is removed and an event telling that no eligible targets could be found should be recorded in it.

### Checklist


controllers/disruption_controller.go
```diff
@@ -611,18 +610,19 @@ func (r *DisruptionReconciler) selectTargets(instance *chaosv1beta1.Disruption)
...
+
+		return fmt.Errorf("error getting eligible targets: %w", err)
 	}
 
 	// if the asked targets count is greater than the amount of found targets, we take all of them
 	targetsCount = int(math.Min(float64(targetsCount), float64(len(eligibleTargets))))
 	if targetsCount < 1 {
 		r.log.Info("ignored targets has reached target count, skipping")
-		r.Recorder.Event(instance, "Warning", "NoTarget", "Ignored Targets is too large, we will no longer choose additional eligible targets for injection for this disruption")
+		r.Recorder.Event(instance, "Warning", "NoTarget", "No eligible targets found for injection for this disruption (either ignored or already targeted by another disruption)")
```
ptnapoleon, line 625: This message is a little misleading? It can absolutely be the  case that we find many eligible targets but still emit this event.

Consider the situation where there are 15 possible targets, and our `count` is set to 5. If ignoredTargets has reached 5 (due to previously injected but now deleted target pods), even if all 15 possible targets are eligible, we still won't select any, and we'll emit this event

Devatoria: I changed the message a bit, there's also a precision at the end of the message which could help understanding it? Otherwise we can split the logic in 2 different events, one when `ignored targets > count` and one when there's no eligible targets?

ptnapoleon (approved):

PR #392, opened 2021-08-16
[CHAOS-123] Fix bug related to delayJitter being in microseconds

Azoam:
## What does this PR do?

- [ ] Adds new functionality
- [X] Fixes a bug
- [ ] Improves Documentation

~~The main change is that when the parameters in tc.go are computed, delayJitter is specifically defined in milliseconds instead of using the default string formatting which may cause the value to read in microseconds.~~
After further testing, it would seem that there are specific formats of time that work (like 1s 50s 100s), but there are other formats that do not (like minutes and microseconds 1m 2m 30m). To keep everything consistent we are going to default to milliseconds for each value, even seconds.

We also saw issues down the road where the 10% delayJitter computation in tc.go was difficult to compute correctly because the function in network_disruption.go had already changed delayJitter into a milliseconds/duration format. Taking 10% of that value would not give us the value we wanted. Therefore we take the 10% with the value of delayJitter before its conversion to milliseconds/duration.

To combat the possibility that delayJitter may be smaller than 1ms, we add the Math.max functionality to guarantee that delayJitter will never be smaller than 1ms.

Motivation for change:

When delayJitter was smaller than 1 millisecond, the default string formatting would describe it in microseconds in the params of AddNetem in tc.go. Microseconds is not something that netem can handle as it expects milliseconds. Therefore we need to make sure that delayJitter is always in milliseconds.

### Testing

```
kind: Disruption
metadata:
  name: network-delay
  namespace: chaos-demo
spec:
  level: pod
  selector:
    app: demo-curl
  count: 1
  network:
    delay: 1000 # delay (in milliseconds) to add to outgoing packets, 10% of jitter will be added by default
    delayJitter: 5 # (optional) add X % (1-100) of delay as jitter to delay (+- X% ms to original delay), defaults to 10%
```
I used many variations of this disruption, playing with the delay and the delayJitter. Here are just a few values I've tested locally that work:

delay: 1, 50, 500, 1000
delayJitter: 0, 1, 5, 50

### Checklist


network/tc_test.go
```diff
@@ -87,7 +87,7 @@ var _ = Describe("Tc", func() {
 
 		Context("add 1s delay and 1s delayJitter to lo interface to the root parent without any handle", func() {
```
ptnapoleon, line 88: Would it be possible to add a test that confirms the default 10% is set appropriately? especially at or near the lower bounds of `delay`? Otherwise I'm +1 on the PR

Azoam: After looking at the code a bit, I noticed that adding the test would require more changes from the original code. The 10% computation is not done within tc.go, specifically the AddNetem function. It is done within injector/network_disruption.go. Therefore in tc_test.go we cannot easily test it because those tests dont go through the Inject code in injector/network_disruption.go. What are your thoughts?

Devatoria (approved):

PR #396, opened 2021-08-18
GRPC with PercentAffected

takakonishimura:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

We want to be able to mangle responses to gRPC requests

### Testing

- [X] I used existing unit tests
- [X] I will manually test in a canary deployment

examples/grpc.yaml

### Checklist


api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,128 @@
...
+func (s GRPCDisruptionSpec) Validate() error {
+	if s.Port == 0 {
+		return errors.New("the gRPC disruption does not specify the port which the target service exposes, but port must be specified")
+	}
+
+	if len(s.Endpoints) == 0 {
+		return errors.New("the gRPC disruption was selected with no endpoints specified, but endpoints must be specified")
+	}
+
+	queryPctByEndpoint := make(map[string]int)
```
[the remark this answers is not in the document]
takakonishimura:
```
>>  kubectl apply -f examples/grpc.yaml 
The Disruption "grpc" is invalid: spec.grpc.endpoints.query_pct: Invalid value: 101: spec.grpc.endpoints.query_pct in body should be less than or equal to 100
```
* When 1 endpoint has `101` (percent), it fails (above
* When 1 endpoint has `100`, it succeeds
* When multiple endpoints to `101`, it fails.
* When multiple endpoints add to `100`. it succeeds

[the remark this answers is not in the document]
takakonishimura: (this is with latest implementation)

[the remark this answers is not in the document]
takakonishimura: unit test spec by calling Validate!

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,118 @@
...
+	// +kubebuilder:validation:Maximum=65535
+	Port      int                  `json:"port,omitempty"`
+	Endpoints []EndpointAlteration `json:"endpoints,omitempty"`
+}
+
+// EndpointAlteration represents an endpoint to disrupt and the corresponding error to return
+type EndpointAlteration struct {
+	TargetEndpoint string `json:"endpoint,omitempty"`
+	// +kubebuilder:validation:Enum=OK;CANCELED;UNKNOWN;INVALID_ARGUMENT;DEADLINE_EXCEEDED;NOT_FOUND;ALREADY_EXISTS;PERMISSION_DENIED;RESOURCE_EXHAUSTED;FAILED_PRECONDITION;ABORTED;OUT_OF_RANGE;UNIMPLEMENTED;INTERNAL;UNAVAILABLE;DATA_LOSS;UNAUTHENTICATED
+	ErrorToReturn string `json:"error,omitempty"`
```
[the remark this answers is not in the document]
takakonishimura: ddmark?

injector/grpc_disruption.go
```diff
@@ -0,0 +1,96 @@
...
+
+	var opts []grpc.DialOption
+	opts = append(opts, grpc.WithInsecure())
+	opts = append(opts, grpc.WithBlock())
+
+	conn, err := grpc.Dial(i.serverAddr, opts...)
+	if err != nil {
+		i.config.Log.Fatalf("fail to dial: %v", err)
+		return errors.New("fail to dial: " + i.serverAddr)
+	}
```
takakonishimura, line 54: helper

injector/grpc_disruption.go
```diff
@@ -0,0 +1,96 @@
...
+		serverAddr: config.TargetPodIP + ":" + strconv.Itoa(spec.Port),
+	}, nil
+}
+
+// Inject injects the given dns disruption into the given container
+func (i GRPCDisruptionInjector) Inject() error {
+	i.config.Log.Infow("connecting to " + i.serverAddr + "...")
+
+	var opts []grpc.DialOption
+	opts = append(opts, grpc.WithInsecure())
```
takakonishimura, line 47: mark in doc that we are insecure

api/v1beta1/disruption_types.go
```diff
@@ -159,13 +161,8 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() error {
...
 	// Rule: at least one disruption field
-	if s.DNS == nil &&
-		s.CPUPressure == nil &&
-		s.Network == nil &&
-		s.NodeFailure == nil &&
-		s.ContainerFailure == nil &&
-		s.DiskPressure == nil {
-		return errors.New("cannot apply an empty disruption - at least one of Network, DNS, DiskPressure, NodeFailure, ContainerFailure, CPUPressure fields is needed")
+	if len(s.GetKindNames()) == 0 {
+		return errors.New("cannot apply an empty disruption - at least one disruption kind (e.g. 'network', 'grpc', 'cpuessure') required")
```
ptnapoleon, line 165:
```suggestion
		return errors.New("cannot apply an empty disruption - at least one disruption kind (e.g. 'network', 'grpc', 'cpuPressure') required")
```

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,132 @@
...
+		// check that endpoint is not already configured such that the sum of mangled queryPercents total to more than 100%
+		if totalQueryPercent, ok := queryPctByEndpoint[alteration.TargetEndpoint]; ok {
+			if alteration.QueryPercent > 0 {
+				if totalQueryPercent+alteration.QueryPercent > 100 {
+					return errors.New("total queryPercent of all alterations applied to endpoint %s is over 100%; modify them to so their total is 100% or less")
+				}
+				queryPctByEndpoint[alteration.TargetEndpoint] += alteration.QueryPercent
+			}
+		} else {
+			if alteration.QueryPercent > 100 {
```
ptnapoleon, line 72: Wouldn't we only get this message if someone specified >100% for a single alteration? The message says that the sum of alterations exceeded 100%

docs/grpc_disruption.md
```diff
@@ -0,0 +1,96 @@
+# GRPC disruption
+
+The `grpc` field offers a way to inject spoofed gRPC responses ont he server-side. To get this disruption to work, you must apply some code changes to the instantiation of your gRPC server.
+
+* `port` is the port exposed on target pods (the target pods are specified in `spec.selector`)
+* `endpoints` is a list of endpoints to alter (a spoof configuration is referred to as an `alteration`)
+  * `<endpoints[i]>.endpoint` indicates the fully qualified api endpoint to override (ex: `/<package>.<service>/<method>`)
+  * Exactly one of `<endpoints[i]>.error` or `<endpoints[i]>.override` should be defined per endpoint alteration, and the only override currently supported is `{}` which returnes `emptypb.Empty`
```
ptnapoleon, line 8:
```suggestion
  * Exactly one of `<endpoints[i]>.error` or `<endpoints[i]>.override` should be defined per endpoint alteration, and the only override currently supported is `{}` which returns `emptypb.Empty`
```

docs/grpc_disruption.md
```diff
@@ -0,0 +1,96 @@
...
+    <kbd>
+        <img src="../docs/img/grpc/request_good.png" height=240 width=500 />
+    </kbd>
+</p>
+
+Now consider this gRPC call which returns an error:
+* A gRPC client makes call to `/chaos_dogfood.ChaosDogfood/FoodOrder` over the gRPC connection with message `FoodOrder { animal: “cat” }`.
+* A gRPC server returns an error with code `INVALID_ARGUMENT`
+
+Here, it is also unclear from the outside if the error code returned is the correct error or if it's masking a an underlying issue. It is also not immediately clear which side produced the error.
```
ptnapoleon, line 49:
```suggestion
Here, it is also unclear from the outside if the error code returned is the correct error or if it's masking an underlying issue. It is also not immediately clear which side produced the error.
```

docs/grpc_disruption/interceptor_algorithm.md
```diff
@@ -0,0 +1,169 @@
+# gRPC Disruption Inceptor
+
+When the interceptor recognizes a query's endpoint as one which is actively getting disrupted, the interceptor generates a random integer from `0` to `100`, and consults an array caching the `PercentToAlteration` mapping. Provided the random integer, this hasmap identifies the disruption to apply to a query. The mapping represents user preference the proportion of queries affected by each alteration. The user cannot sepcify queryPercents for a single endpoint which sum to over `100%`. You can see an example below of an array that does not define all 100% of possible requests. You can see the next section for an example of a mapping which only has 25 entires.
```
ptnapoleon, line 3: Is this still true? I notice this is alternating between mentioning hashmap vs array

api/v1beta1/grpc_disruption.go
takakonishimura: Hm, it could technically be the first of 2 but it's just over 100 anyways?

Like
`200`
`50`

I'm down to be specific too

docs/grpc_disruption/interceptor_algorithm.md
takakonishimura: oof, yeah. I will call it a hashmap because that's more true to the design, and mention it's an array underneath the hood.

grpc/alteration_percent_test.go
```diff
@@ -0,0 +1,587 @@
...
+				{
+					ErrorToReturn:    "",
+					OverrideToReturn: "{}",
+					QueryPercent:     int32(50),
+				},
+			}
+
+			var err error
+			mapping, err = GetAlterationToPercentAffected(alterationSpecs, "endpointname")
+			Expect(err).To(BeNil())
```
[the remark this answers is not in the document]
takakonishimura: It would only return an error if the alterationSpecs are invalid. In this case, it does not return an error because `endpointname` is only used to print errors. This "smells" bad to pass a parameter just for the error state, but I don't really know how to error without the name of the endpoint, how can the user correct their mistake? Should I be handling the error outside of the function?

grpc/server.go
```diff
@@ -0,0 +1,288 @@
...
+// SendDisruption makes a gRPC call to DisruptionListener.
+func (d *DisruptionListener) SendDisruption(ctx context.Context, ds *pb.DisruptionSpec) (*emptypb.Empty, error) {
+	if ds == nil {
+		log.Error("Cannot execute SendDisruption when DisruptionSpec is nil")
+		return nil, status.Error(codes.InvalidArgument, "Cannot execute SendDisruption when DisruptionSpec is nil")
+	}
+
+	config := DisruptionConfiguration{}
+
+	for _, endpointSpec := range ds.Endpoints {
```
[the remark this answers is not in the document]
takakonishimura: Looking into it some more, there's a 3rd structure (the pb item). However, i think it's necessary to transfer the data over (and I can't reuse it with kubebuilder because that's tied to the the CRD. The final format with the percentages is also necessary with the current design. Let me know if you disagree I can wrestle with it some more!

api/v1beta1/grpc_disruption.go
ptnapoleon: I am pretty sure that the error message at L67 covers the situation where we have multiple alterations that total 100, and that we can only hit this error if an individual alteration is over 100, so I think this error message (L74) should be more specific just to be clearer to users

api/v1beta1/grpc_disruption_test.go
```diff
@@ -0,0 +1,260 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+/*
```
ptnapoleon, line 6: Seems to have an extra license header in here, you can remove the second

api/v1beta1/grpc_disruption_test.go
```diff
@@ -0,0 +1,260 @@
...
+					OverrideToReturn: "{}",
+					QueryPercent:     50,
+				},
+			},
+		}
+
+		Expect(spec.Validate()).ToNot(BeNil())
+	})
+
+	It("Good errors to return", func() {
```
ptnapoleon, line 179: Should the query percents in this test be > 0?

controllers/disruption_controller_test.go
```diff
@@ -255,33 +293,33 @@ var _ = Describe("Disruption Controller", func() {
...
-	// 		chaosPodKey := types.NamespacedName{Namespace: chaosPod.Namespace, Name: chaosPod.Name}
-
-	// 		By("Deleting one of the chaos pod")
-	// 		Expect(k8sClient.Delete(context.Background(), &chaosPod)).To(BeNil())
-
-	// 		By("Waiting for the chaos pod finalizer to be removed")
-	// 		Eventually(func() error { return k8sClient.Get(context.Background(), chaosPodKey, &chaosPod) }, timeout).Should(MatchError(fmt.Sprintf("Pod \"%s\" not found", chaosPod.Name)))
-	// 	})
-	// })
+	Context("manually delete a chaos pod", func() {
```
ptnapoleon, line 303: Seems like the merge from main had some unintended test changes

docs/grpc_disruption.md
```diff
@@ -0,0 +1,96 @@
+# GRPC disruption
+
+The `grpc` field offers a way to inject spoofed gRPC responses on the server-side. To get this disruption to work, you must apply some code changes to the instantiation of your gRPC server.
+
+* `port` is the port exposed on target pods (the target pods are specified in `spec.selector`)
```
ptnapoleon, line 5: I'm not sure this flows clearly from the previous line. We were just introducing the code changes to the grpc server needed, when we start explaining all of the spec options

docs/grpc_disruption.md
```diff
@@ -0,0 +1,96 @@
...
+* `endpoints` is a list of endpoints to alter (a spoof configuration is referred to as an `alteration`)
+  * `<endpoints[i]>.endpoint` indicates the fully qualified api endpoint to override (ex: `/<package>.<service>/<method>`)
+  * Exactly one of `<endpoints[i]>.error` or `<endpoints[i]>.override` should be defined per endpoint alteration, and the only override currently supported is `{}` which returns `emptypb.Empty`
+  * `<endpoints[i]>.query_pct` defines (out of 100) how frequently this alteration should occur; you may have multiple alterations per endpoint, but you cannot specify a sum total of more than 100 percent for any given endpoint
+
+Note: At this time, Chaos Controller does not support:
+* disrupting gRPC Streams
+* chaining the disruption interceptor on another interceptor you already have
```
ptnapoleon, line 13: while we haven't tested this, we aren't exactly sure it doesn't work either, so maybe mention this is more of a "caveat emptor" sort of unsupported, rather than a "definitely does not work" unsupported like grpc streams

docs/grpc_disruption/interceptor_algorithm.md
```diff
@@ -0,0 +1,169 @@
+# gRPC Disruption Inceptor
+
+When the interceptor recognizes a query's endpoint as one which is actively getting disrupted, the interceptor generates a random integer from `0` to `100`, and consults a hasmap for the `PercentToAlteration` mapping. The hashmap is represented as an array of 100 or less items, and provided the random integer, it identifies the disruption to apply to a query. The user cannot sepcify queryPercents for a single endpoint which sum to over `100%`. You can see an example below of a hashmap that does not define all 100% of possible requests below.
```
ptnapoleon, line 3: Is this still a hashmap in the code?

ptnapoleon:
```suggestion
When the interceptor recognizes a query's endpoint as one which is actively getting disrupted, the interceptor generates a random integer from `0` to `100`, and consults a hashmap for the `PercentToAlteration` mapping. The hashmap is represented as an array of 100 or less items, and provided the random integer, it identifies the disruption to apply to a query. The user cannot specify queryPercents for a single endpoint which sum to over `100%`. You can see an example below of a hashmap that does not define all 100% of possible requests below.
```

grpc/alteration_percent_test.go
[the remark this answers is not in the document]
ptnapoleon: I think I understand? But if this is a situation we expect to never be in due to validation, why test that it works?

grpc/server.go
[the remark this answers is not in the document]
ptnapoleon: Hmm, does seem like I was wrong then, though I do wish there were some way to repeat this validation less often.

grpc/percent_slot_test.go
```diff
@@ -0,0 +1,130 @@
...
+	Context("with three alterations", func() {
+		BeforeEach(func() {
+			mapping[altCfgs[0]] = PercentAffected(15)
+			mapping[altCfgs[1]] = PercentAffected(20)
+			mapping[altCfgs[2]] = PercentAffected(35)
+
+			percentToAlteration = GetPercentToAlteration(mapping)
+		})
+
+		It("should create 70 SlotPercents", func() {
```
ptnapoleon, line 69: what is a slotpercent, here?

grpc/server.go
```diff
@@ -0,0 +1,288 @@
...
+		// add endpoint to main configuration
+		config[targeted] = EndpointConfiguration{
+			TargetEndpoint: targeted,
+			Alterations:    alterationToPercentAffected,
+			AlterationHash: percentSlotToAlteration,
+		}
+	}
+
+	mutex.Lock()
+	d.Configuration = config
```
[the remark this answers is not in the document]
ptnapoleon: I was the one to ask to put the check at the beginning of the file, but could you actually move it down here inside the mutex, for race safety?

ptnapoleon (changes requested):
It appears the new tests don't run, afaict, because the suite_test.go file is missing. I'm seeing a few of them fail once I add that

api/v1beta1/grpc_disruption_test.go
takakonishimura: This should divide the errors evenly across 100 percent slots.

api/v1beta1/grpc_disruption_test.go
```diff
@@ -0,0 +1,260 @@
...
+					ErrorToReturn:    "UNIMPLEMENTED",
+					OverrideToReturn: "",
+					QueryPercent:     0,
+				},
+			},
+		}
+
+		Expect(spec.Validate()).To(BeNil())
+	})
+})
```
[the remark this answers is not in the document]
takakonishimura: good point, it's in `alteration_percent_test.go` which is collapsed, but it's not validated upfront.

api/v1beta1/grpc_disruption.go
takakonishimura: Hm, I reread a few times. Hear me out:

**Scenario 1:** 
If the input is: `50` then `50`, we would hit the `else` block, then on the second loop hit the `if` block, discover we reached `100` and error.

**Scenario 2:**
If the input is: `100` only, then we would hit the `else` block and in that case we'd need to catch that it was over 100% because we'd never hit the `if` block again.

**Scenario 3:**
Tricky one is this one. If we have `100` and then `50` for the same endpoint, it'll get flagged in the `else` block but it's not actually a single alteration. The sum is still `150` > `100`.

I tried to group `Scenario 2` and `Scenario 3` into the same if-statement to have less lines of code. I thought about treating lists with 1 item separately to make it clearer, but since we are iterating through and discovering as we go, that would require looping through them again after the tallies are taken.

Let me know if this sounds fishy, I think it is right.

api/v1beta1/grpc_disruption_test.go
takakonishimura: (and therefore be a valid test)

grpc/percent_slot_test.go
takakonishimura: tried to make it make more sense with the x out of 100

grpc/alteration_percent_test.go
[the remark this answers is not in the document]
takakonishimura: Tell me if this makes sense:

The (in)correctness of having over 100% is checked in `percent_to_alteration` file, this is just explicitly showing that it's valid but we will never access more than 100%. Few rows under:

> // intuition here as that the service will never pick a number larger than 100 randomly
> // so this return value never gets triggered, but the function does not error out

api/v1beta1/grpc_disruption.go
ptnapoleon: Hmm, two objections: (1) We only error on `> 100`, not `>= 100`, so I think technically both scenarios 1 and 2 are good and shouldn't error. (2) I've realized that because we are setting a max of 100 for any given queryPercent with ddmark, we can just drop the `if` check inside the `else`, as we should expect to literally never hit it?

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		// check that exactly one of ErrorToReturn or OverrideToReturn is configured
+		if alteration.ErrorToReturn != "" && alteration.OverrideToReturn != "" {
+			return fmt.Errorf("the gRPC disruption has ErrorToReturn and OverrideToReturn specified for endpoint %s, but it can only have one", alteration.TargetEndpoint)
+		} else if alteration.ErrorToReturn == "" && alteration.OverrideToReturn == "" {
+			return fmt.Errorf("the gRPC disruption must have either ErrorToReturn or OverrideToReturn specified for endpoint %s", alteration.TargetEndpoint)
+		}
+
+		if alteration.ErrorToReturn != "" && alteration.ErrorToReturn != "OK" &&
```
ptnapoleon, line 87: I think we should be using the error map `grpc/server.go` here, rather than re-listing every error type explicitly, wdyt?

api/v1beta1/grpc_disruption.go
takakonishimura: (oops yeah, should have been `51` and `50`, `101`)

Okay that's true I'll remove the if-statement!

api/v1beta1/grpc_disruption.go
takakonishimura: there is an import cycle 😭 
image

takakonishimura: oh I guess I could move it?

takakonishimura: I moved the map so there's no cycle

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,345 @@
...
+package api_test
+
+import (
+	"github.com/DataDog/chaos-controller/api/v1beta1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("GRPCDisruption Validation", func() {
```
ptnapoleon, line 15: Admittedly I dont know how ginkgo and gomega work very well, but do we not need any `Context` blocks?

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,345 @@
...
+					QueryPercent:     100,
+				},
+			},
+		}
+
+		err := spec.Validate()
+		Expect(err.Error()).To(Equal("the gRPC disruption has ErrorToReturn and OverrideToReturn specified for endpoint /chaos_dogfood.ChaosDogfood/order, but it can only have one"))
+	})
+
+	It("Error and override cannot both be defined", func() {
```
ptnapoleon, line 33:
```suggestion
	It("Error and override cannot both be undefined", func() {
```

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,345 @@
...
+					ErrorToReturn:    "CANCELED",
+					OverrideToReturn: "",
+					QueryPercent:     100,
+				},
+			}}
+
+		Expect(spec.Validate()).To(BeNil())
+	})
+
+	It("Query percent which sums to 100 on two endpoints does validate", func() {
```
ptnapoleon, line 103:
```suggestion
	It("Query percents which sum to 100 on one endpoint does validate", func() {
```

There's inconsistency in these test names. Sometimes "two endpoints" means two distinct endpoints, but other times it means two alterations for a single endpoint

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,345 @@
...
+					QueryPercent:     49,
+				},
+			}}
+
+		Expect(spec.Validate()).To(BeNil())
+	})
+
+	It("Query percent which sums to more than 100 on two endpoints does not validate", func() {
+		spec := v1beta1.GRPCDisruptionSpec{
+			Port: 50051,
```
ptnapoleon, line 147: You could use a `BeforeEach` block to create a spec object and define the Port in only one place, so you dont have to repeat this line.

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,345 @@
...
+					ErrorToReturn:    "ALREADY_EXISTS",
+					OverrideToReturn: "",
+					QueryPercent:     51,
+				},
+			}}
+
+		Expect(spec.Validate()).ToNot(BeNil())
+	})
+
+	It("Bad error to return", func() {
```
ptnapoleon, line 166:
```suggestion
	It("Errors not in the standard grpc errors do not validate", func() {
```

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,345 @@
...
+					OverrideToReturn: "",
+					QueryPercent:     50,
+				},
+			},
+		}
+
+		Expect(spec.Validate()).ToNot(BeNil())
+	})
+
+	It("Good errors to return", func() {
```
ptnapoleon, line 182:
```suggestion
	It("All standard grpc errors validate", func() {
```

grpc/alteration_percent_test.go
```diff
@@ -0,0 +1,582 @@
...
+			var err error
+			mapping, err = GetAlterationToPercentAffected(alterationSpecs)
+			Expect(err).To(BeNil())
+		})
+
+		It("should create a mapping with 3 elements", func() {
+			Expect(len(mapping)).To(Equal(3))
+		})
+
+		It("should create a mapping with correct configs", func() {
```
ptnapoleon, line 52: Please let me know if I'm wrong, but it seems like the BeforeEach is going to be called before both of these `It`s? I think that's likely not your intention, and definitely is spending extra time than needed

grpc/alteration_percent_test.go
```diff
@@ -0,0 +1,582 @@
...
+import (
+	"fmt"
+
+	. "github.com/DataDog/chaos-controller/grpc"
+	pb "github.com/DataDog/chaos-controller/grpc/disruption_listener"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("get mapping from randomly generated Percent to Alteration based on AlterationToPercentAffected", func() {
```
ptnapoleon, line 17: If they have the same `Describe`, how are `alteration_percent_test.go` and `percent_to_alteration_test.go` different?

ptnapoleon: okay I now see how they're different, but it does seem like at least one of them needs a different `Describe` text

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+	mapping := make(map[AlterationConfiguration]PercentAffected)
+
+	// unquantified is a holding variable used later to calculate and assign percentages to alterations which do not specify queryPercent
+	unquantifiedAlts := make([]AlterationConfiguration, 0)
+
+	pctClaimed := 0
+
+	for _, altSpec := range endpointSpecList {
+		if altSpec.ErrorToReturn == "" && altSpec.OverrideToReturn == "" {
+			return nil, status.Error(codes.InvalidArgument, "Cannot execute SendDisruption without specifying either ErrorToReturn or OverrideToReturn for all target endpoints")
```
ptnapoleon, line 199: I don't think these two error message should assume who is calling this function, and instead should just return a basic error, and let the caller handle wrapping that

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+			TargetEndpoint: TargetEndpoint(endpointSpec.TargetEndpoint),
+			Alterations:    alterationToPercentAffected,
+			AlterationHash: percentToAlteration,
+		}
+	}
+
+	mutex.Lock()
+	if len(d.Configuration) > 0 {
+		d.Logger.Error("Cannot apply new DisruptionSpec when DisruptionListener is already configured")
+		return nil, status.Error(codes.AlreadyExists, "annot apply new DisruptionSpec when DisruptionListener is already configured")
```
ptnapoleon, line 109:
```suggestion
		return nil, status.Error(codes.AlreadyExists, "Cannot apply new DisruptionSpec when DisruptionListener is already configured")
```

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+
+			d.Logger.Error("Endpoint %s should define either an ErrorToReturn or OverrideToReturn but does not", endptConfig.TargetEndpoint)
+		}
+	}
+
+	return handler(ctx, req)
+}
+
+// GetAlterationToPercentAffected takes a series of alterations configured for a Spec and maps them to the percentage of queries which should be affected
```
ptnapoleon, line 187: Should be specific that its the list of alterations for a single endpoint, not for an entire spec, right?

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+				pctClaimed += pctPerAlt
+			}
+		}
+	}
+
+	return mapping, nil
+}
+
+// GetPercentToAlteration Converts a mapping from alterationConfiguration to the percentage of requests which should return this altered response
+func GetPercentToAlteration(endpointSpecList map[AlterationConfiguration]PercentAffected, logger *zap.SugaredLogger) []AlterationConfiguration {
```
ptnapoleon, line 241: Is there anyway to rename this function and/or the one above? `GetPercentToAlterations` vs `GetAlterationToPercentAffected` is hard for me, and hopefully users too (and it isnt just me)

ptnapoleon: also the logger is unused here, so shouldnt be a parameter

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+			OverrideToReturn: altSpec.OverrideToReturn,
+		}
+
+		// Intuition:
+		// (1) add all endpoints where queryPercent is specified
+		// (2) track percentage of queries already claimed by an endpointConfiguration
+		// (3) track endpoints where queryPercent is missing and calculate them later
+		if altSpec.QueryPercent > 0 {
+			mapping[alterationConfig] = PercentAffected(altSpec.QueryPercent)
+			pctClaimed += int(altSpec.QueryPercent)
```
ptnapoleon, line 217: It would be, I think, very easy to check if we exceed 100 here and return an error. Why choose not to do so? That would help make some of the unit tests a little more clear, specifically the ones that check the certain error cases which we choose not to catch.

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+		}
+	}
+
+	return mapping, nil
+}
+
+// GetPercentToAlteration Converts a mapping from alterationConfiguration to the percentage of requests which should return this altered response
+func GetPercentToAlteration(endpointSpecList map[AlterationConfiguration]PercentAffected, logger *zap.SugaredLogger) []AlterationConfiguration {
+	hashMap := make([]AlterationConfiguration, 0, 100)
```
ptnapoleon, line 242: this isnt a hashmap and should be renamed?

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+
+// DisruptionConfiguration configures the DisruptionListener to chaos test endpoints of a gRPC server.
+type DisruptionConfiguration map[TargetEndpoint]EndpointConfiguration
+
+// EndpointConfiguration configures endpoints that the DisruptionListener chaos tests on a gRPC server.
+// The AlterationHash maps integers from 0 to 100 to alteration configurations.
+type EndpointConfiguration struct {
+	TargetEndpoint TargetEndpoint
+	Alterations    map[AlterationConfiguration]PercentAffected
+	AlterationHash []AlterationConfiguration
```
ptnapoleon, line 45: Not a hashmap, will be confusing for users?

grpc/server.go
```diff
@@ -0,0 +1,251 @@
...
+			d.Logger.Error("DisruptionSpec does not specify TargetEndpoint for at least one endpointAlteration")
+			return nil, status.Error(codes.InvalidArgument, "Cannot execute SendDisruption without specifying TargetEndpoint for all endpointAlterations")
+		}
+
+		alterationToPercentAffected, err := GetAlterationToPercentAffected(endpointSpec.Alterations)
+		if err != nil {
+			return nil, err
+		}
+
+		percentToAlteration := GetPercentToAlteration(alterationToPercentAffected, d.Logger)
```
ptnapoleon, line 96: If the output of `GetAlterationToPercentAffected` is sent directly into `GetPercentToAlteration` before being added to the config, why have two separate functions?

injector/grpc_disruption.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// GRPCDisruptionInjectorConfig contains all needed drivers to create a grpc disruption
+type GRPCDisruptionInjectorConfig struct {
+	Config
+}
+
+// NewGRPCDisruptionInjector creates a GRPCDisruptionInjector object with the given config,
+// missing fields are initialized with the defaults
+func NewGRPCDisruptionInjector(spec v1beta1.GRPCDisruptionSpec, config GRPCDisruptionInjectorConfig) (Injector, error) {
```
ptnapoleon, line 33: If we never return an error in the function, then error doesnt need to be in the returned values (simplifies the callsite)

injector/grpc_disruption.go
```diff
@@ -0,0 +1,105 @@
...
+	i.config.Log.Infow("connecting to " + i.serverAddr + "...")
+
+	if i.config.DryRun {
+		i.config.Log.Infow("adding dry run mode grpc disruption", "spec", i.spec)
+		return nil
+	}
+
+	conn, err := connectToServer(i.serverAddr)
+	if err != nil {
+		i.config.Log.Fatalf(err.Error())
```
ptnapoleon, line 54: I am pretty sure that a fatal log kills the process, and thus we wouldnt end up returning this error afterwards in the next line? Should this just be an error log?

injector/grpc_disruption_test.go
```diff
@@ -0,0 +1,134 @@
...
+	. "github.com/DataDog/chaos-controller/injector"
+
+	pb "github.com/DataDog/chaos-controller/grpc/disruption_listener"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/mock"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+func TestSendAndCleaDisruption(t *testing.T) {
```
ptnapoleon, line 34: This file should use `injector_suite_test.go` to match the other injector unit tests, rather than importing testing.T itself

injector/grpc_disruption_test.go
```diff
@@ -0,0 +1,134 @@
...
+				ErrorToReturn:    "",
+				OverrideToReturn: "{}",
+				QueryPercent:     0,
+			},
+		},
+	}
+
+	disruptionListenerClient := &DisruptionListenerClientMock{}
+
+	// define expectations
```
ptnapoleon, line 71: What does this file test?

grpc/percent_to_alteration_test.go
```diff
@@ -0,0 +1,172 @@
...
+			mapping[altCfgs[2]] = PercentAffected(35)
+
+			percentToAlteration = GetPercentToAlteration(mapping, logger)
+
+			Expect(len(percentToAlteration)).To(Equal(70))
+
+			position := 0
+			altCfg := percentToAlteration[position]
+
+			for i := 0; i < 3; i++ {
```
ptnapoleon, line 69: It _may_ be simpler to loop through every item in `percentToAlteration` and just count how many times each error is set? Then compare to the original percents specified?

grpc/generate_endpointspec_test.go
```diff
@@ -0,0 +1,133 @@
...
+					QueryPercent:     0,
+				},
+			}
+
+			var err error
+			endpointSpec = GenerateEndpointSpecs(endpointAlterations)
+			Expect(err).To(BeNil())
+		})
+
+		It("should create a list of endpointSpecs with 2 elements", func() {
```
ptnapoleon, line 62: double check that you arent using `It` where you should use `By`

grpc/generate_endpointspec_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Expect(len(endpointSpec)).To(Equal(2))
+		})
+
+		It("should create and endpointSpec for api_1 with 3 elements", func() {
+			var endpointSpec_1 *pb.EndpointSpec
+			if endpointSpec[0].TargetEndpoint == "service/api_1" {
```
ptnapoleon, line 68: Is this because the order is indeterminate? I think that's worth commenting here for future readers

ptnapoleon (changes requested):
At least for me locally, `go mod tidy`, `go mod vendor`, and `inv license-check` all have updates

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		// check that endpoint is not already configured such that the sum of mangled queryPercents total to more than 100%
+		if alteration.QueryPercent == 0 {
+			if count, ok := unquantifiedAlts[alteration.TargetEndpoint]; ok {
+				unquantifiedAlts[alteration.TargetEndpoint] = count + 1
+
+				pctClaimed := 100 - queryPctByEndpoint[alteration.TargetEndpoint]
+				if pctClaimed < count+1 {
+					return fmt.Errorf("%s will never return some alterations because alterations exceed 100%% of possible queries", alteration.TargetEndpoint)
```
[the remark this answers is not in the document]
takakonishimura: I *THINK* `%%` prints `%` https://pkg.go.dev/fmt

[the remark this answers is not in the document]
takakonishimura: What do you think of `Alterations must have at least 1%% chance of occurring;`

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,173 @@
...
+				}
+			} else {
+				queryPctByEndpoint[alteration.TargetEndpoint] = alteration.QueryPercent
+			}
+		}
+
+		// check that exactly one of ErrorToReturn or OverrideToReturn is configured
+		if alteration.ErrorToReturn != "" {
+			if alteration.OverrideToReturn == "" {
+				if _, ok := ErrorMap[alteration.ErrorToReturn]; !ok {
```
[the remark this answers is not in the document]
takakonishimura: Context here:

* Originally, I felt good about not having `ErrorMap` here because I validate `gRPC` strings.
* However, I do not think we unit-test `ddmark`.
* In order to make sure that the errors being request will actually work on the server side, we need to make sure the server is using string mappings which work with `ddmark`. And in fact if they're not in sync, then `chaos-controller` should change `ddmark` rather than have our users update their DisruptionListener services (for example with `CANCELED` vs `CANCELLED`.
* By having the `ErrorMap` mapping here and importing it in the server code, we can unit test that the mapping is the same, meaning the only thing that can go wrong is the ddmark.

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,173 @@
...
+	OverrideToReturn string `json:"override,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=100
+	// +ddmark:validation:Minimum=1
+	// +ddmark:validation:Maximum=100
+	QueryPercent int `json:"query_pct,omitempty"`
+}
+
+// Validate validates that there are no missing hostnames or records for the given grpc disruption spec
+func (s GRPCDisruptionSpec) Validate() error {
```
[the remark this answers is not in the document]
takakonishimura: added to `api/v1beta1/disruption_types.go`

api/v1beta1/grpc_disruption.go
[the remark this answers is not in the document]
ptnapoleon: ohhhhh duh duh duh my bad!

api/v1beta1/grpc_disruption.go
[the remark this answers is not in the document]
ptnapoleon: We could add unit testing for the ddmark inside this test https://github.com/DataDog/chaos-controller/pull/396/files#diff-808195d7ed62fe71cd613f1261479ea92b280d2026e5a44e307a7f79ac58b4e9R182 ? That unit test could pull directly from ErrorMap, accomplishing the same goal but without validating twice?

api/grpc_disruption_test.go
takakonishimura: Okay, I did some reading and it _*appears*_ that `Describe` and `Context` are just two mechanisms of grouping tests (`Context` is a nested scope in the case that you want to test different types of outcomes for one topic).

`By` is also a way of nesting multiple behaviors into one `It`.

injector/grpc_disruption_test.go
takakonishimura: hm, so I need the variable `t` to run `disruptionListenerClient.AssertExpectations(t)` at the bottom of the file so this is using a slightly different testing approach than the other files by leveraging `AssertExpectations` (there's no equivalent in the `Expect` universe that allows you to express nuances in function calls that the `matchedBy` allows (since we aren't sure which of two maps actually gets returned).

injector/grpc_disruption_test.go
takakonishimura: I want to test that a fairly common script calls `SendDisruption` with the expected parameters using gRPC without actually making the call. This is not encapsulated by our current `e2e` test which only checks that the pod lifecycles are as expected. As our `e2e` gets more sophisticated, we likely won't need this test anymore. I'm also okay to remove this file if we feel ready to do so!

injector/grpc_disruption_test.go
takakonishimura: This is what I get when I make it a simple `Describe` test `injector/grpc_disruption_test.go:130:48: undefined: t`

api/v1beta1/disruption_types.go
```diff
@@ -186,6 +184,10 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() error {
 		}
 	}
 
+	if s.GRPC != nil && s.Level != chaostypes.DisruptionLevelPod && s.Level != chaostypes.DisruptionLevelUnspecified {
```
ptnapoleon, line 187:
```suggestion
	if s.GRPC != nil && s.Level != chaostypes.DisruptionLevelPod {
```

takakonishimura: @ptnapoleon I put this back after `e2e-test` failed because our `e2e-test` does not specify a level for the specs, I think unspecified should default to pod level like what I believe `init-container` is doing. We could also break up the end-to-end test file, but it seems excessive to me?

api/v1beta1/grpc_disruption.go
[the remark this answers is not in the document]
takakonishimura: Need to figure out what file path is, my test fails for this

examples/complete.yaml
```diff
@@ -68,3 +68,15 @@ spec:
       record:
         type: CNAME # return a CNAME record
         value: google.com # hostname to return
+  grpc: # disrupt gRPC responses by faking results
+    port: 50051 # port to disrupt
+    endpoints:
+      - endpoint: /chaos_dogfood.ChaosDogfood/getCatalog # gRPC service endpoint to disrupt
+        error: NOT_FOUND # gRPC error code to return instead computed response
+        query_pct: 10 # percentage to affect (1-100); multiple alterations allowed for single endpoint, but sum should not exceed 100%
```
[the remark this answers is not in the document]
ptnapoleon: Given that everything else uses camelCase, I agree it's preferable to switch to queryPercent over query_pct

docs/grpc_disruption/interceptor_algorithm.md
```diff
@@ -0,0 +1,180 @@
...
+    .. -> Error: ...
+    22 -> Error: PERMISSION_DENIED
+    23 -> Error: PERMISSION_DENIED
+    24 -> Error: PERMISSION_DENIED
+}
+```
+
+In this case, we would return an Override with empty results for 5% of queries, a `NOT_FOUND` error for 5% of queries, and return `PERMISSION_DENIED` for 15% of queries.
+
+:warning: If you are a maintainer looking at the source code, the "mapping" is represented as a golang slice with length <= 100.
```
Azoam, line 58: It seems strange to see documentation here aimed at the maintainer. Maybe you can move this explanation somewhere like a comment in your code (which I assume you already have) or maybe a clever way of doing it is referencing the code somewhere here where a comment saying this is located.

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,356 @@
...
+
+	BeforeEach(func() {
+		spec = v1beta1.GRPCDisruptionSpec{
+			Port:      50051,
+			Endpoints: []v1beta1.EndpointAlteration{},
+		}
+	})
+
+	Context("Error and override are both defined", func() {
+		It("errors becuase exactly one of error or override must be defined for an alteration", func() {
```
ptnapoleon, line 27:
```suggestion
		It("errors because exactly one of error or override must be defined for an alteration", func() {
```

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,356 @@
...
+				},
+			}
+
+			err := spec.Validate()
+			Expect(err.Error()).To(Equal("the gRPC disruption has ErrorToReturn and OverrideToReturn specified for endpoint /chaos_dogfood.ChaosDogfood/order, but it can only have one"))
+		})
+	})
+
+	Context("Error and override are both undefined", func() {
+		It("errors becuase exactly one of error or override must be defined for an alteration", func() {
```
ptnapoleon, line 43:
```suggestion
		It("errors because exactly one of error or override must be defined for an alteration", func() {
```

api/v1beta1/grpc_disruption.go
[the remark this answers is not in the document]
takakonishimura: updated!

controllers/disruption_controller_test.go
```diff
@@ -4,13 +4,10 @@
 // Copyright 2021 Datadog, Inc.
 
 /*
```
takakonishimura, line 6: I am not sure why this is in some files and not others

api/v1beta1/grpc_disruption.go
```diff
@@ -0,0 +1,167 @@
...
+
+	queryPctByEndpoint := make(map[string]int)
+	unquantifiedAlts := make(map[string]int)
+
+	for _, alteration := range s.Endpoints {
+		if alteration.TargetEndpoint == "" {
+			return errors.New("some list items in gRPC disruption are missing endpoints; specify a TargetEndpoint for each item in the list")
+		}
+
+		// check that endpoint is not already configured such that the sum of mangled queryPercents total to more than 100%
```
ptnapoleon, line 88: mangled means, like, "mutilated", right? Why use it here?

docs/grpc_disruption/interceptor_algorithm.md
```diff
@@ -0,0 +1,180 @@
...
+    99  -> Error: PERMISSION_DENIED
+    100 -> Error: PERMISSION_DENIED
+}
+```
+
+## Design Implications
+
+### Setting 0 as query percentage
+
+Users should avoid explicitly setting `query_pct: 0`. Because the Kubebuilder infers 0 for omitted `int`s, this configuration is synonymous for the chaos-controller to not setting a `query_pct` at all. This means that the alteration with this `query_pct` setting would actually be applied to the remaining queries that have not yet be claimed by another alteration. There isn't a straight forward way to chaos-controller to discriminate an unset `query_pct` from one explicitly set to `0`, so users may be surprised by inputting the value `0` and receving many of errors with that input.
```
ptnapoleon, line 157:
```suggestion
Users should avoid explicitly setting `query_pct: 0`. Because the Kubebuilder infers 0 for omitted `int`s, this configuration is synonymous for the chaos-controller to not setting a `query_pct` at all. This means that the alteration with this `query_pct` setting would actually be applied to the remaining queries that have not yet be claimed by another alteration. There isn't a straight forward way to chaos-controller to discriminate an unset `query_pct` from one explicitly set to `0`, so users may be surprised by inputting the value `0` and receiving many of errors with that input.
```

"many of errors" could flow more clearly.

What are we hoping to warn people against here? I don't know why anyone would ever explicitly set a query percentage of 0 and think it would never happen, because why even add that alteration then? What's here is perfectly fine, but we can probably also just tell people that `0` is a special value and should expect it to be treated that way

Azoam (commented):
Read through the docs, LGTM, really clear from both a user's perspective and from a developers perspective.

grpc/calculations/get_altconfig_to_querypct_test.go
```diff
@@ -0,0 +1,481 @@
...
+				var err error
+				QueryPercentByAltConfig, err = ConvertAltSpecToQueryPercentByAltConfig(alterationSpecs)
+				Expect(err).To(BeNil())
+			})
+
+			By("returning 3 elements", func() {
+				Expect(len(QueryPercentByAltConfig)).To(Equal(3))
+			})
+
+			By("by assigning a query percentage of 20 to CANCELLED error", func() {
```
ptnapoleon, line 51:
```suggestion
			By("by assigning a query percentage of 20 to CANCELED error", func() {
```

grpc/calculations/get_alterationmap_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	Context("with three alterations that add to 100", func() {
+		It("should create 100 (out of 100 potential) slots with alteration configurations", func() {
+			alterationConfigToQueryPercent[altCfgs[0]] = QueryPercent(40)
+			alterationConfigToQueryPercent[altCfgs[1]] = QueryPercent(40)
+			alterationConfigToQueryPercent[altCfgs[2]] = QueryPercent(20)
+
+			alterationMap = ConvertQueryPercentByAltConfigToAlterationMap(alterationConfigToQueryPercent)
+
+			By("having 70 entries", func() {
```
ptnapoleon, line 104:
```suggestion
			By("having 100 entries", func() {
```

grpc/calculations/get_alterationmap_test.go
```diff
@@ -0,0 +1,134 @@
...
+				Expect(len(alterationMap)).To(Equal(100))
+			})
+
+			By("having three alteration types each with the righ number of configurations", func() {
```
ptnapoleon, line 108:
```suggestion
			By("having three alteration types each with the right number of configurations", func() {
```

grpc/calculations/get_alterationmap_test.go
```diff
@@ -0,0 +1,134 @@
...
+			alterationConfigToQueryPercent[altCfgs[1]] = QueryPercent(20)
+			alterationConfigToQueryPercent[altCfgs[2]] = QueryPercent(35)
+
+			alterationMap = ConvertQueryPercentByAltConfigToAlterationMap(alterationConfigToQueryPercent)
+
+			By("having 70 entries", func() {
+				Expect(len(alterationMap)).To(Equal(70))
+			})
+
+			By("having three alteration types each with the righ number of configurations", func() {
```
ptnapoleon, line 69:
```suggestion
			By("having three alteration types each with the right number of configurations", func() {
```

grpc/client.go
```diff
@@ -0,0 +1,78 @@
...
+
+	chaosv1beta1 "github.com/DataDog/chaos-controller/api/v1beta1"
+	pb "github.com/DataDog/chaos-controller/grpc/disruption_listener"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// ExecuteSendDisruption takes in a CRD specification for GRPC disruptions and
+// executes a SendDisruption call on the provided DisruptionListenerClient
+func ExecuteSendDisruption(client pb.DisruptionListenerClient, spec chaosv1beta1.GRPCDisruptionSpec, logger *zap.SugaredLogger) {
```
ptnapoleon, line 20: Personally, I would suggest neither this function nor `ExecuteCleanDisruption` takes a logger. For both functions, they are only doing so in order to log an error. I think it's preferable for them to return an error instead, and have the caller deal with logging

grpc/client.go
```diff
@@ -0,0 +1,78 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	_, err := client.CleanDisruption(ctx, &emptypb.Empty{})
+	if err != nil {
+		logger.Error("Received an error: %v", err)
+	}
+}
+
+// GenerateEndpointSpecs converts an EndpointAlteration into a list of EndpointSpecs which
```
ptnapoleon, line 43:
```suggestion
// GenerateEndpointSpecs converts a slice of EndpointAlterations into a list of EndpointSpecs which
```

grpc/generate_endpointspec_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+var _ = Describe("construct DisruptionListener query for configuring disruptions from api spec", func() {
+	var (
+		endpointAlterations []chaosv1beta1.EndpointAlteration
+		endpointSpec        []*pb.EndpointSpec
+	)
+
+	Context("with five alterations which add up to less than 100", func() {
+		BeforeEach(func() {
```
ptnapoleon, line 23: Why run this before each `It` rather than just once?

grpc/generate_endpointspec_test.go
```diff
@@ -0,0 +1,135 @@
...
+				}
+			}
+
+			Expect(canceled_found).To(BeTrue())
+			Expect(already_exists_found).To(BeTrue())
+			Expect(empty_response_found).To(BeTrue())
+		})
+
+		It("should create and endpointSpec for api_2 with 2 elements", func() {
+			// handling that the results of `endpointSpec` is indeterminate
```
ptnapoleon, line 105:
```suggestion
			// handling that the order of `endpointSpec` is indeterminate
```
nit

grpc/server.go
```diff
@@ -0,0 +1,136 @@
...
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+/*
+ * When a random integer from 0 to 100 is randomly selected, the PercentToAlteration mapping is referenced to
+ * identify the disruption to apply to a query. The mapping represents user preference the proportion of queries
+ * affected by each alteration. See docs/grpc_disruption/interceptor_algorithm.md for examples.
+ */
+
+// DisruptionListener is a gRPC Service that can disrupt endpoints of a gRPC server.
```
ptnapoleon, line 31: The package structure under `grpc` is a bit confusing to me. Why isn't `DisruptionListener` in the `disruption_listener` package?

grpc/server.go
```diff
@@ -0,0 +1,136 @@
...
+type DisruptionListener struct {
+	pb.UnimplementedDisruptionListenerServer
+	Configuration grpc_calcapi.DisruptionConfiguration
+	Logger        *zap.SugaredLogger
+}
+
+var mutex sync.Mutex
+
+// NewDisruptionListener creates a new DisruptionListener Service with the logger instantiated and DisruptionConfiguration set to be empty
+func NewDisruptionListener() (*DisruptionListener, error) {
```
ptnapoleon, line 41: According to my IDE, this is unused. shouldn't it be?

grpc/server.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/DataDog/chaos-controller/log"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+/*
+ * When a random integer from 0 to 100 is randomly selected, the PercentToAlteration mapping is referenced to
```
ptnapoleon, line 26: Is this where this comment should be?

grpc/server.go
```diff
@@ -0,0 +1,136 @@
...
+
+	var err error
+	d.Logger, err = log.NewZapLogger()
+
+	d.Configuration = grpc_calcapi.DisruptionConfiguration{}
+
+	return &d, err
+}
+
+// SendDisruption receives a disruption specfication and configures the interceptor to spoof responses to specified endpoints.
```
ptnapoleon, line 52:
```suggestion
// SendDisruption receives a disruption specification and configures the interceptor to spoof responses to specified endpoints.
```

grpc/server.go
```diff
@@ -0,0 +1,136 @@
...
+func (d *DisruptionListener) SendDisruption(ctx context.Context, ds *pb.DisruptionSpec) (*emptypb.Empty, error) {
```
ptnapoleon, line 53: It feels wrong to never check this `ctx`. It might be good to check it hasn't timed out or been canceled inside the mutex where we actually save the configuration

grpc/server.go
ptnapoleon: This is for users to call, right? In that case, we should definitely take in the logger as an argument, as people will want to configure the zap logger, rather than just taking the default from `NewZapLogger()`

ptnapoleon: And that gets us out of returning an `err` here.

injector/grpc_disruption.go
```diff
@@ -0,0 +1,105 @@
...
+	return GRPCDisruptionInjector{
+		spec:       spec,
+		config:     config,
+		serverAddr: config.TargetPodIP + ":" + strconv.Itoa(spec.Port),
+	}
+}
+
+// Inject injects the given dns disruption into the given container
+func (i GRPCDisruptionInjector) Inject() error {
+	i.config.Log.Infow(fmt.Sprintf("%t", i.config.DryRun))
```
ptnapoleon, line 43: This seems to duplicate the log we're about to do in L48?

api/grpc_disruption_test.go
```diff
@@ -0,0 +1,377 @@
...
+					)
+				}
+
+				Expect(len(spec.Endpoints)).To(Equal(17))
+
+				err := ddmark.ValidateStruct(spec, "grpc_test_suite", "github.com/DataDog/chaos-controller/api/v1beta1")
+				Expect(len(err)).To(Equal(0))
+			})
+
+			It("Do validate", func() {
```
ptnapoleon, line 216: nit: `Do validate` and `do not validate` are a little terse, maybe `passes validation` or `fails validation`?

ptnapoleon:
I can understand wanting to merge this w/o adding this, given the size of the PR, but I do think we need a convenient way for maintainers to test changes to this disruption kind. How are you testing this locally? Can you add scripts, docs, or yaml manifests that set up the grpc server and client, so that we can disrupt them?

ptnapoleon:
> Would you want a script that checks that these error rates are within a threshold of the expected error rate, or is it okay as long as I leave a documentation on how to run the service end-to-end and visually check the logs?

I'm not really worried about the error percentages being correct, I think it's fine to handle this in a follow-up PR, as discussed

PR #400, opened 2021-08-31
CHAOS-31: Support durationSeconds in disruption spec

ptnapoleon:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

- Adds a new field to the disruption spec: `durationSeconds`
- Adds a new mutatingadmissionwebhook for setting spec defaults
- Adds a new controller configuration option: `expired-disruption-gc-delay`
- Adds a new disruption injection status: `PreviouslyInjected`
- Rolls back the minikube kubernetes version to 1.19.latest

Disruptions now have a `spec.durationSeconds` value, that defaults to `3600` (one hour). When chaos pods are created, their `activeDeadlineSeconds` is set so that they expire `spec.durationSeconds` after the disruption was first created (not the chaos pod). After a disruption's duration has expired, at least`expired-disruption-gc-delay` seconds later, the disruption will be deleted and all of the chaos pods removed.

Motivation for change:
https://github.com/DataDog/chaos-controller/discussions/379

### Testing

### Checklist

My documentation is not up to date, but I wanted to get feedback before I finished the PR

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"time"
 
 	chaosapi "github.com/DataDog/chaos-controller/api"
 	chaostypes "github.com/DataDog/chaos-controller/types"
@@ -49,7 +50,7 @@ type DisruptionSpec struct {
 	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
 	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
 	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
-	DurationSeconds  int64                             `json:"durationSeconds"`            // time from disruption creation until chaos pods are deleted and no more are created
+	Duration         time.Duration                     `json:"duration,omitempty"`         // time from disruption creation until chaos pods are deleted and no more are created
 	// +kubebuilder:validation:Enum=pod;node;""
 	// +ddmark:validation:Enum=pod;node;""
 	Level      chaostypes.DisruptionLevel `json:"level,omitempty"`
@@ -158,11 +159,6 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() error {
 	if s.ContainerFailure != nil && s.Level == chaostypes.DisruptionLevelNode {
 		return errors.New("cannot execute a container failure because the level configuration is set to node")
 	}
-
-	// Rule: must have a durationSeconds // Can't implement until we are a v1 CRD because you can't default in v1beta1
-	// if s.DurationSeconds == 0 {
-	//	 return errors.New("must have durationSeconds set in your disruption spec")
-	// }
 
 	// Rule: at least one disruption field
 	if s.DNS == nil &&
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -49,6 +49,7 @@ type DisruptionSpec struct {
 	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
 	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
 	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
+	DurationSeconds  int64                             `json:"durationSeconds"`            // time from disruption creation until chaos pods are deleted and no more are created
 	// +kubebuilder:validation:Enum=pod;node;""
 	// +ddmark:validation:Enum=pod;node;""
 	Level      chaostypes.DisruptionLevel `json:"level,omitempty"`
@@ -157,6 +158,11 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() error {
 	if s.ContainerFailure != nil && s.Level == chaostypes.DisruptionLevelNode {
 		return errors.New("cannot execute a container failure because the level configuration is set to node")
 	}
+
+	// Rule: must have a durationSeconds // Cant implement until we are a v1 CRD because you cant default in v1beta1
+	//if s.DurationSeconds == 0 {
+	//	return errors.New("must have durationSeconds set in your disruption spec")
+	//}
 
 	// Rule: at least one disruption field
 	if s.DNS == nil &&
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -159,10 +159,10 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() error {
 		return errors.New("cannot execute a container failure because the level configuration is set to node")
 	}
 
-	// Rule: must have a durationSeconds // Cant implement until we are a v1 CRD because you cant default in v1beta1
-	//if s.DurationSeconds == 0 {
-	//	return errors.New("must have durationSeconds set in your disruption spec")
-	//}
+	// Rule: must have a durationSeconds // Can't implement until we are a v1 CRD because you can't default in v1beta1
+	// if s.DurationSeconds == 0 {
+	//	 return errors.New("must have durationSeconds set in your disruption spec")
+	// }
 
 	// Rule: at least one disruption field
 	if s.DNS == nil &&
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"time"
 
 	chaosapi "github.com/DataDog/chaos-controller/api"
 	chaostypes "github.com/DataDog/chaos-controller/types"
@@ -50,7 +49,7 @@ type DisruptionSpec struct {
 	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
 	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
 	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
-	Duration         time.Duration                     `json:"duration,omitempty"`         // time from disruption creation until chaos pods are deleted and no more are created
+	DurationSeconds  int64                             `json:"durationSeconds,omitempty"`  // seconds from disruption creation until chaos pods are deleted and no more are created
 	// +kubebuilder:validation:Enum=pod;node;""
 	// +ddmark:validation:Enum=pod;node;""
 	Level      chaostypes.DisruptionLevel `json:"level,omitempty"`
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -120,7 +120,7 @@ minikube-start:
 		--container-runtime=containerd \
 		--memory=${minikube-memory} \
 		--cpus=4 \
-		--kubernetes-version=1.21.1 \
+		--kubernetes-version=1.19.14 \
```
ptnapoleon, line 123: The behavior of pods changed between 1.19 and 1.21. In the former, pods that exceed their deadline become Failed with a reason of "DeadlineExceeded" (which we catch). In the latter, pods simply become Completed, and aren't in a failed state

api/v1beta1/disruption_types.go
```diff
@@ -49,6 +49,7 @@ type DisruptionSpec struct {
 	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
 	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
 	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
+	DurationSeconds  int64                             `json:"durationSeconds"`            // time from disruption creation until chaos pods are deleted and no more are created
```
ptnapoleon, line 52: `int64` chosen here so very large seconds values can be used.

webhook/defaults.go
```diff
@@ -0,0 +1,63 @@
...
+	decoder *admission.Decoder
+}
+
+func (m *DefaultMutator) InjectDecoder(d *admission.Decoder) error {
+	m.decoder = d
+
+	return nil
+}
+
+func (m *DefaultMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
```
ptnapoleon, line 32: While this only mutates the duration for now, this is a good way to handle all defaults until we move to v1 CRD

controllers/disruption_controller_test.go
```diff
@@ -190,6 +191,27 @@ var _ = Describe("Disruption Controller", func() {
 		})
 	})
 
+	Context("disruption expires naturally", func() {
+		BeforeEach(func() {
+			disruption.Spec.Count = &intstr.IntOrString{Type: intstr.String, StrVal: "100%"}
+			disruption.Spec.DurationSeconds = int64(timeout.Seconds() + 5)
```
ptnapoleon, line 197: This test adds like ~25 seconds to the unit tests. Worth it?

controllers/disruption_controller.go
```diff
@@ -197,13 +208,19 @@ func (r *DisruptionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+		requeueDelay := time.Duration(math.Max(float64(calculateRemainingDurationSeconds(*instance)), float64(r.ExpiredDisruptionGCDelay))) * time.Second
 
+		r.log.Infow("requeuing disruption", "requeueDelay", requeueDelay.String())
+
+		return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: requeueDelay,
+			},
+			r.Update(context.Background(), instance)
+	}
```
Devatoria, line 223:
```suggestion
	}

```

controllers/disruption_controller.go
```diff
@@ -342,7 +360,7 @@ func (r *DisruptionReconciler) startInjection(instance *chaosv1beta1.Disruption)
 
 				// send metrics and events
 				r.Recorder.Event(instance, "Normal", "Created", fmt.Sprintf("Created disruption injection pod for \"%s\"", instance.Name))
-				r.recordEventOnTarget(instance, target, "Warning", "Disrupted", fmt.Sprintf("Pod %s from disruption %s targeted this resourcer for injection", chaosPod.Name, instance.Name))
+				r.recordEventOnTarget(instance, target, "Warning", "Disrupted", fmt.Sprintf("Pod %s from disruption %s targeted this resource for injection", chaosPod.Name, instance.Name))
```
Devatoria, line 363: Good catch :')

controllers/disruption_controller_test.go
Devatoria: You can let it here, it'll be modified with e2e tests (which will take more time anyway as they really go through injections and cleanups).

controllers/disruption_controller.go
```diff
@@ -74,6 +74,7 @@ type DisruptionReconciler struct {
 	log                                   *zap.SugaredLogger
 	InjectorServiceAccountNamespace       string
 	InjectorNetworkDisruptionAllowedHosts []string
+	ExpiredDisruptionGCDelay              int
```
Devatoria, line 77: Same question as for the duration field, why not using a `time.Duration` here? If we keep an `int`, could we eventually specify the unit in the same as you did for the duration field? Like `ExpiredDisruptionGCSeconds` or something like this.

main.go
```diff
@@ -118,6 +119,9 @@ func main() {
 	pflag.StringVar(&cfg.Controller.ImagePullSecrets, "image-pull-secrets", "", "Secrets used for pulling the Docker image from a private registry")
 	handleFatalError(viper.BindPFlag("controller.imagePullSecrets", pflag.Lookup("image-pull-secrets")))
 
+	pflag.IntVar(&cfg.Controller.ExpiredDisruptionGCDelay, "expired-disruption-gc-delay", 300, "Seconds after a disruption expires before being automatically deleted")
```
Devatoria, line 122: Can you also add this configuration to the [configmap](https://github.com/DataDog/chaos-controller/blob/main/chart/templates/configmap.yaml) and [values.yaml](https://github.com/DataDog/chaos-controller/blob/main/chart/values.yaml) files of the chart so it can be configured through it please?

webhook/defaults.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// decode object
+	if err := m.decoder.Decode(req, dis); err != nil {
+		m.Log.Errorw("error decoding disruption object", "error", err, "name", req.Name, "namespace", req.Namespace)
+
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if dis.Spec.DurationSeconds == 0 {
+		defaultDuration := int64(3600)
```
Devatoria, line 50: Can you add a way to configure this value from the controller configuration please? It looks like a magic number to have a hard coded default value here and one would like to change this value.

api/v1beta1/disruption_types.go
Devatoria: The field has no `omitempty` json tag here, meaning that it'll be flagged as required, automatically rejecting the disruption if the field is not set. Though we have a mutating webhook setting a default if the field is not set. Should we set this field as optional so it does not force people to set it but we still have the default value from the webhook?

Makefile
ptnapoleon: Both are working, we handle both behaviors

api/v1beta1/disruption_types.go
ptnapoleon: Didn't even realize we could take a complex type like that here, I'll switch to Duration.  I am fairly sure that the mutating webhook fires first, so even if people don't set it, it's fine. I want to insist on setting this explicitly as much as possible though?

ptnapoleon: I'll confirm the behavior

api/v1beta1/disruption_types.go
```diff
@@ -158,6 +159,11 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() error {
 		return errors.New("cannot execute a container failure because the level configuration is set to node")
 	}
 
+	// Rule: must have a durationSeconds // Can't implement until we are a v1 CRD because you can't default in v1beta1
```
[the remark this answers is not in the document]
ptnapoleon: No

api/v1beta1/disruption_types.go
Devatoria: > I am fairly sure that the mutating webhook fires first

Yes, mutating webhooks are triggered before validating ones

> I want to insist on setting this explicitly as much as possible though?

My main concern is that adding a required field to an existing CRD would break any kind of automation or would break any "saved" disruptions (users keeping a previously working disruption somewhere) while we already have what's needed to apply a default value if not set. Though I understand why you would like it to be explicit, I think adding new fields as optional with a default value is a best practice to avoid breaking changes.

chart/values.yaml
```diff
@@ -13,6 +13,8 @@ images: # images and tag to pull for each component of the stack
 controller:
   deleteOnly: false # enable delete-only mode
   metricsSink: noop # metrics driver (noop or datadog)
+  defaultDuration: 1h # default spec.duration for a disruption with none specified
```
Devatoria, line 16: Good to keep in mind that changing this delay will NOT affect existing disruptions (and it is the expected behavior anyway).

Devatoria (approved):
LGTM

Devatoria (changes requested):
Just realized that some doc is missing, could you just add a line about the duration feature and its default value somewhere [in the features doc](https://github.com/DataDog/chaos-controller/blob/main/docs/features.md) please?

Devatoria (approved):

chart/templates/webhook.yaml
```diff
@@ -107,6 +107,37 @@ webhooks:
     resources:
     - disruptions
 ---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
```
nikos912000, line 111: What is the purpose of this webhook btw?
We are trying to limit the number of mutating webhooks internally as they could impact the entire cluster.

chart/templates/configmap.yaml
```diff
@@ -16,6 +16,8 @@ data:
       metricsSink: {{ .Values.controller.metricsSink | quote }}
       deleteOnly: {{ .Values.controller.deleteOnly }}
       imagePullSecrets: {{ .Values.images.pullSecrets }}
+      defaultDuration: {{ .Values.controller.defaultDuration }}
+      expiredDisruptionGCDelay: {{ .Values.controller.expiredDisruptionGCDelay }}
```
nikos912000, line 20: I was curious about the need for this. Can't we just rely on the `duration` field and delete the CR once the time-window is exceeded?

chart/templates/webhook.yaml
nikos912000: Ah, the project uses v1beta1 for CRDs which doesn't support default values.

ptnapoleon: Indeed, that's why. We'll be moving off a mutating webhook once we're on v1, which should be by eoy

chart/templates/configmap.yaml
ptnapoleon: Personally, yes, I think that would work and would be fine, but concern has been expressed that users might find it surprising for their disruption CRDs to suddenly disappear. What do you think?

nikos912000: Thanks for the context. Although it adds some complexity it sounds sensible :)

ptnapoleon:
Confirmed this worked in testing in a live cluster

Devatoria (commented):
Can you add a specific examples for this feature like `examples/timed_disruption.yaml` just showing the feature in addition to the `complet.yaml` one please? It completes the set of examples we have.

Otherwise LGTM

PR #401, opened 2021-09-02
End-to-end/integration tests running against a real Kubernetes cluster

Devatoria:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug
- [ ] Improves Documentation

A brief description of changes to implementation or controller behavior: it implements end-to-end/integration tests running against a real Kubernetes cluster from the CI (running in Minikube).

Motivation for change: running tests against a real cluster helps to spot potential bugs we would not see in unit tests because of all the mocking.

Here's the implementation:

- we do not spawn a controller from the test suite anymore but we are using the [envtest USE_EXISTING_CLUSTER feature](https://book.kubebuilder.io/reference/envtest.html#environment-variables) instead
- we split unit tests (`make test`) from integration tests (`make e2e-test`) because of the difference of configuration to run them
- in the CI, we now spin up a Minikube cluster running in a VM with the `docker` driver (but still using the `containerd` container engine, both being independent) which will be used to run the controller and run the tests
- still in the CI, once the Minikube cluster is up and running, we deploy the controller as a user would do in their local environment (install cert-manager, `make minikube-build` to build and load docker images into Minikube, `make install` to render and apply the chart)

Pros:

- we now have real tests (and we were able to spot some bugs)
- the `Makefile` is simplified with some `minikube` commands to build and load images
- the `Makefile` is now using `minikube kubectl` instead of `kubectl` directly; not only it removes a dependency on `kubectl`, but it also prevents one from mistakenly running those commands against another cluster than the minikube one

Cons:

- of course, because of the Minikube setup in the CI, tests take longer to terminate (around 10 minutes today but it'll increase when adding more tests)

### Testing

### Checklist


.circleci/config.yml
```diff
@@ -183,5 +220,11 @@ workflows:
       - test:
           requires:
             - prepare-env
+      - test:
```
ptnapoleon, line 223: Seems wrong to have `test` here twice?

controllers/suite_test.go
```diff
@@ -236,110 +138,73 @@ var _ = BeforeEach(func() {
...
 
-	By("Creating target pod")
-	err := k8sClient.Create(context.Background(), targetPodA)
-	if apierrors.IsInvalid(err) {
-		logf.Log.Error(err, "failed to create object, got an invalid object error")
-		return
-	}
-	Expect(err).NotTo(HaveOccurred())
+	By("Creating target pods")
+	Expect(k8sClient.Create(context.Background(), targetPodA)).ToNot(HaveOccurred())
```
ptnapoleon, line 149: I don't understand, what are we expecting to have not occurred here?

ptnapoleon: nvm, it's the err returned by Create

controllers/suite_test.go
```diff
@@ -236,110 +138,73 @@ var _ = BeforeEach(func() {
...
+		if err := k8sClient.Get(context.Background(), types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, &p); err != nil {
+			return false, fmt.Errorf("error getting pod: %w", err)
+		}
+
+		// check the pod phase
+		if p.Status.Phase != corev1.PodRunning {
+			return false, nil
+		}
+
+		// check the pod containers statuses (phase can be running while all pods are not running)
```
ptnapoleon, line 194: should this say "while all containers are not running"?

controllers/suite_test.go
```diff
@@ -236,110 +138,73 @@ var _ = BeforeEach(func() {
...
+		running := false
+		for _, status := range p.Status.ContainerStatuses {
+			if status.State.Running != nil {
+				running = true
+
+				break
```
ptnapoleon, line 200: Won't we return true as long as any containers are running? Is that the goal?

.circleci/config.yml
Devatoria: Indeed, copy paste fail

controllers/suite_test.go
Devatoria: Yup this assertion is misleading. I think `.To(BeNil())` would have the same behavior and eventually be more explicit?

controllers/suite_test.go
Devatoria: Indeed, thanks

controllers/suite_test.go
Devatoria: Definitely my bad, I inverted the whole condition 😅 fixed

controllers/suite_test.go
ptnapoleon: Yes, that's what I'm used to from our tests, and I think is more explicit, but maybe that's because i'm more familiar with it? This (to me) reads like we expect the Create to have not occurred, even though it really means the error

CONTRIBUTING.md
```diff
@@ -106,4 +118,4 @@ See [minikube_image.md](docs/minikube_image.md)
 
 ## 3rd-party licenses
 
-3rd-party references and licenses are kept in the [LICENSE-3rdparty.csv](LICENSE-3rdparty.csv) file. This file has been generated by the [tasks/3rdparty.py](tasks/3rdparty.py) script. If any vendor is updated, added or removed, this file must be updated as well.
\ No newline at end of file
+3rd-party references and licenses are kept in the [LICENSE-3rdparty.csv](LICENSE-3rdparty.csv) file. This file has been generated by the [tasks/3rdparty.py](tasks/3rdparty.py) script. If any vendor is updated, added or removed, this file must be updated as well.
```
ptnapoleon, line 121:
```suggestion
3rd-party references and licenses are kept in the [LICENSE-3rdparty.csv](LICENSE-3rdparty.csv) file. This file has been generated by the [tasks/thirdparty.py](tasks/thirdparty.py) script. If any vendor is updated, added or removed, this file must be updated as well.
```

i cant tell what actually changed in this line, but a good time to fix this typo

ptnapoleon (approved):

PR #402, opened 2021-09-07
CHAOS[102] Bug Fix - Only Target Ctns with Relevant Paths @ Disk Disruption

Azoam:
## What does this PR do?

There is a bug in the current state of the chaos controller which follows the following user story:
A user tries to apply a disk disruption to a target pod with 3 containers. 2 of which have the corresponding `/mnt/foo` which is described in the disk disruption configuration, while the last container has a different volume entirely. When the user attempts to run the disruption they are met with an error claiming the path does not exist even though it does in 2 of their containers.

This disruption filters out containers which do not have the corresponding path specified in the disk disruption configuration. This will allow users to run their disruptions on their specific containers even if they have other containers with different paths.

The implementation of this bug fix opens up the possibility that in the future we will be filtering out containers based on other factors besides the disk disruption. 

- [] Adds new functionality
- [X] Fixes a bug
- [X] Improves Documentation

Motivation for change:

Without this bug fix, users would not be able to run a disk disruption if their containers had different volume's setup on them.

### Testing

- [X] I used existing unit tests

I used the current `demo.yaml` file found under `/examples`. All I did was change `demo-curl` by adding the following to the list of containers:
```
...
- name: novolumehere
        image: ubuntu:bionic
        command: ["/bin/bash"]
        args: ["-c", "while true; do sleep 1; done"]
```
this is the only container to not have a volume in that deployment and before the bug fix, running a disk disruption on the volume `/mnt/data` would have errored. After the bug fix, no error occurs as the `novolumehere` container is successfully ignored.

### Checklist


docs/disk_pressure.md
```diff
@@ -18,6 +18,11 @@ To apply the throttle, the injector will:
 
 **Note: the throttle will be applied to the whole device (for the pod only) and not only to the partition handling the path.**
 
+### Container Filtering
+
+There may be the scenario where a user wants to run a disk disruption but only on specific containers which have specific volume's set up, leaving the other containers to be unaffected.
+This is possible at no additional expense to the user as the disk disruption will filter out all containers which do not carry the volume's specified in the user's disk disruption configuration.
```
ptnapoleon, line 24:
```suggestion
This is possible with no additional configuration for the user as the disk disruption will filter out all containers which do not carry the volumes specified in the user's disk disruption configuration.
```

docs/disk_pressure.md
```diff
@@ -18,6 +18,11 @@ To apply the throttle, the injector will:
 
 **Note: the throttle will be applied to the whole device (for the pod only) and not only to the partition handling the path.**
 
+### Container Filtering
+
+There may be the scenario where a user wants to run a disk disruption but only on specific containers which have specific volume's set up, leaving the other containers to be unaffected.
```
ptnapoleon, line 23:
```suggestion
There may be a scenario where a user wants to run a disk disruption but only on specific containers which have specific volumes set up, leaving the other containers to be unaffected.
```

controllers/helpers.go
```diff
@@ -32,6 +32,74 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// filterContainerIDs filters out any containers that may not be intended for disruption in a multi-container disruption
```
ptnapoleon, line 35: Could this comment explain the function of the three returned string slices?

controllers/disruption_controller.go
```diff
@@ -294,12 +294,24 @@ func (r *DisruptionReconciler) startInjection(instance *chaosv1beta1.Disruption)
...
 			targetNodeName = pod.Spec.NodeName
-
 			// get IDs of targeted containers or all containers
 			containerIDs, err = getContainerIDs(&pod, instance.Spec.Containers)
 			if err != nil {
 				return fmt.Errorf("error getting target pod container ID: %w", err)
 			}
+
+			// filter containers
+			var removed []string
```
ptnapoleon, line 304: Could you add an e2e-test for this new feature? it should be possible to work it into an existing one, and just change the number of expected injectors?

controllers/helpers.go
```diff
@@ -32,6 +32,78 @@ import (
...
+// this function returns 3 string slices:
+// containers: The updated list of containers with the filtered out containers removed
+// removed: The list of containers that were removed from the list containers, used later for logging purposes
+// reasoning: A list of strings used later for logging the reason as to why the corresponding container in removed was removed
```
ptnapoleon, line 39: purely optional, but since `removed` and `reasoning` are so inextricably tied together, you could return a slice of a new struct that represents a removed/reasoning tuple, and reduce the return values to just two slices

ptnapoleon: It looks like `reasoning` is always the same string, right? In that case, why not remove this return value and just use that string directly at the point of logging?

controllers/helpers.go
```diff
@@ -32,6 +32,78 @@ import (
...
+}
+
+// removeCtnsFromConsideration removes containers from the removal list that do not comply with multi-container
+// disruption rules. Returns the updated list of containers
+func removeCtnsFromConsideration(pod *corev1.Pod, removal []string, containers []string) []string {
+	if len(removal) == 0 {
+		return containers
+	}
+
+	for _, ctn := range pod.Status.ContainerStatuses {
```
ptnapoleon, line 60: I also cannot find a way to avoid this triple embedded for loop, given that volume mounts are in pod.Spec but containerID is in pod.Status

controllers/helpers.go
Azoam: Reasoning may be different in the future, for now there is only one value for it, but in the future if we filter out containers for other reasons, that reasoning will most likely change, thats why I've set that string to its own set. I do like the idea of putting these in a struct and returning a slice.

controllers/helpers.go
Azoam: I am not a fan of it either :/

controllers/helpers.go
```diff
@@ -32,6 +32,81 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// filtering container meta data containing a list of filtered out containers and the reasoning for the
+// the filter.
+type filteringContMeta struct {
```
ptnapoleon, line 37: nit: this name is a little terse, could it just be `filteredContainers`? Also, could `removed` be clear if it's ids or names?

controllers/helpers.go
ptnapoleon: This comment needs updated now, but 👍

controllers/helpers.go
ptnapoleon: Also, I might suggest that you use a slice of this struct, rather than a struct of slices? That way the removed/reasoning tuples are more closely packaged, rather than relying on all users of this struct to make sure the two inner slices have matching orders

Azoam: I've added to the comments that removed is specifically the names of the containers, let me know if there was any other place to clarify that I missed.

controllers/helpers.go
```diff
@@ -32,6 +32,86 @@ import (
...
 
+// filtering container meta data containing a list of filtered out containers (by name) and the reasoning for
+// the filter.
+type filteredContainer struct {
+	removed   string
+	reasoning string
+}
+
+// filterContainerIDs filters out any containers that may not be intended for disruption in a multi-container disruption
+// this function returns 3 string slices:
```
ptnapoleon, line 43: This part of the comment needs changed, now that it isn't 3 string slices.

ptnapoleon (dismissed):
One comment change request, and it's optional but I do still feel we don't need to keep a slice of the `reasoning` string given it's static for now, but +1

examples/demo.yaml
```diff
@@ -102,6 +102,10 @@ spec:
         volumeMounts:
         - mountPath: /mnt/data
           name: data
+      - name: novolumehere
```
Devatoria, line 105: Can you remove the mount from one of the existing container (for example [the dummy one](https://github.com/DataDog/chaos-controller/blob/a2eb1c5b32178a965b508eaf17c48ea1e8b92360/examples/demo.yaml#L102)) instead of creating a new one? We already have 2 containers in the `demo-curl` pod and none are really using the mount.

Devatoria (commented):
Overall, the PR looks very complex for a specific use case of a specific disruption. Would it not be simpler to just handle that in the disk informer itself?
* the [FromPath](https://github.com/DataDog/chaos-controller/blob/main/disk/disk.go#L29) method could return `(nil, nil)` if the given path is not found
* we can catch this special return case [in the disk injector](https://github.com/DataDog/chaos-controller/blob/main/injector/disk_pressure.go#L52)
It would avoid all that logic of removing a container from the list of containers especially because it does not work well with the case of a disruption applying multiple kinds of disruptions against the same target (eg. a network + disk disruption against a container not mounting the specific path... it would not apply the network disruption while it could just because the container would be removed from the list of containers)

Devatoria:
Should we close this one?

PR #403, opened 2021-09-09
CHAOS-142: Dont rely on containerd reported cgroup path

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug
- [ ] Improves Documentation

The injector no longer solely relies on the cgroup path for a container reported by containerd. Instead, it also scrapes all of the cgroup paths (logic courtesy of the datadog agent), and will fall back to the detected paths on disk when the reported path does not exist.

Motivation for change:
we were experiencing an issue in some hosts where containerd would return a cgroup path that was very incorrect, and did not exist.

### Testing

### Checklist


cgroup/cgroup_detect.go
```diff
@@ -0,0 +1,192 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+package cgroup
+
+// This entire file mostly copied from the datadog-agent
```
ptnapoleon, line 8: Importing this package from the agent was a little difficult because most of these functions arent exported or intended for external use, so I copied them (and modified slightly to remove dd-agent specific config stuff)

cgroup/cgroup.go
```diff
@@ -82,7 +89,26 @@ func (m manager) write(path, data string) error {
...
+	if generatedPath != foundPath {
+		m.log.Infow("containerd generated path does not match found cgroup path",
+			"generatedPath", generatedPath,
+			"foundPath", foundPath,
+		)
+	}
+
+	if _, err := os.Stat(generatedPath); os.IsNotExist(err) {
+		m.log.Infow("containerd generated path does not exist, using found path instead",
+			"generatedPath", generatedPath,
```
takakonishimura, line 104: Do we want this line?

cgroup/cgroup.go
```diff
@@ -82,7 +89,26 @@ func (m manager) write(path, data string) error {
 
 // generatePath generates a path within the cgroup like /<mount>/<kind>/<path (kubepods)>
 func (m manager) generatePath(kind string) string {
-	return fmt.Sprintf("%s%s/%s", m.mount, kind, m.path)
+	generatedPath := fmt.Sprintf("%s%s/%s", m.mount, kind, m.path)
+	foundPath := fmt.Sprintf("%s%s/%s", m.mount, kind, m.cgroup.Paths[kind])
```
takakonishimura, line 93: do we need to check that `m.cgroup.Paths[kind]` exists?

cli/injector/main.go
```diff
@@ -187,9 +187,27 @@ func initConfig() {
...
+
+		return
+	}
+
 	// create cgroup managers
-	for _, cgroupPath := range cgroupPaths {
-		cgroupMgr, err := cgroup.NewManager(dryRun, cgroupPath)
+	for i, cgroupPath := range cgroupPaths {
+
+		rawID := strings.Split(containerIDs[i], "://") // This is "guaranteed" to work because it was already validated when we called container.New(containerID) earlier
```
takakonishimura, line 201: can we call this `rawContainerID`

cgroup/cgroup.go
ptnapoleon: Yep, we do! Good eye.

cgroup/cgroup.go
ptnapoleon: I think so! I am pretty sure I want to know when these paths aren't the same, whether or not the generated path exists

cgroup/cgroup.go
ptnapoleon: Okay, so I've checked the behavior, and I don't think we need to. this is our fallback path, and if it doesn't exist we are gonna error immediately after this is called anyway. I've checked what happens if you get a string from a map[string]string if the key doesn't exist, and it will just be an empty string, so we'll log the warning appropriately and the error will occur in the expected place

takakonishimura (approved):

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
ptnapoleon, line 1: This is due to go 1.17, and will get auto_generated for anyone on that go version who compiles

cgroup/cgroup.go
```diff
@@ -59,6 +65,7 @@ func (m manager) read(path string) (string, error) {
 // write appends the given data to the given cgroup file path
 // NOTE: depending on the cgroup file, the append will result in an overwrite
 func (m manager) write(path, data string) error {
+	m.log.Infow("writing to cgroup file", "path", path, "data", data)
```
ptnapoleon, line 68: I am happy to turn these down to debug, if desired?

takakonishimura (approved):
lgtm!

Azoam (approved):

cgroup/cgroup.go
nathantournant: I think it'd be okay to merge it as it is, and deactivate it once we're more confortable with the new cgroup pathing

nathantournant (approved):

PR #416, opened 2021-09-23
Modify ddmark ExclusiveFields to allow for specifying a one-to-many relationship

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- In order to specify that neither `NodeFailure` nor `ContainerFailure` are appropriate to set alongside other disruption kinds, I wanted to use ddmark's `ExclusiveField` feature. However, because this has previously only been used to handle pairs of fields, assumed a many-to-many relationship: if any specified field is set, none of the others can. I needed a one-to-many relationship, if the first field is set, the rest cannot, but the rest don't necessarily have an interdependent relationship. This change enables that, without breaking any of the existing uses (at least ours)

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/ddmark/README.md
+++ b/ddmark/README.md
@@ -30,7 +30,7 @@ It also allows us to define custom rules to apply to our structures, and focus t
 - ExclusiveFields:
   - `// +ddmark:validation:ExclusiveFields={<fieldName1>,<fieldName2>,...}`
   - Applies to: `<struct>` type
-  - Asserts only one of the given fieldnames isn’t `nil`
+  - Asserts that if `<fieldname1>` can only be non-nil iff all of the other fields are `nil`
 - Maximum:
   - `// +ddmark:validation:Maximum=<int/uint>`
   - Applies to: `int/uint` field
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -37,6 +37,8 @@ import (
 
 // DisruptionSpec defines the desired state of Disruption
 // +ddmark:validation:ExclusiveFields={Network,DNS}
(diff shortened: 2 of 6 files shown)


ddmark/README.md
```diff
@@ -30,7 +30,7 @@ It also allows us to define custom rules to apply to our structures, and focus t
 - ExclusiveFields:
   - `// +ddmark:validation:ExclusiveFields={<fieldName1>,<fieldName2>,...}`
   - Applies to: `<struct>` type
-  - Asserts only one of the given fieldnames isn’t `nil`
+  - Asserts that if `<fieldname1>` can only be non-nil iff all of the other fields are `nil`
```
Devatoria, line 33:
```suggestion
  - Asserts that if `<fieldname1>` can only be non-nil if all of the other fields are `nil`
```

Devatoria (commented):
Can we add some extra unit tests to validate the new behavior of exclusive fields?

ddmark/README.md
ptnapoleon: Intentional! Meant "if and only if" here. Is that shorthand unfamiliar? I can remove it then

nathantournant: Sugg: - Asserts that `<fieldname1>` can only be non-nil if all of the other fields are `nil`

Alternative (seems more readable to me): - Asserts that `<fieldname1>` can be non-nil only if all of the other fields are `nil`
(that's one too many ifs?

Devatoria: TIL `iff` :D sorry

Devatoria (approved):

PR #417, opened 2021-09-23
[CHAOS-102] Ignoring containers without specified path for disk disruptions

Azoam:
## What does this PR do?

- [ ] Adds new functionality
- [X] Fixes a bug
- [X] Improves documentation or testing

One could want to apply a disk disruption on a multi containers pod without specifying a container because only one of them has the expected volume. However, doing so triggers a fatal error on the injector because the volume can’t be found.

Instead of erroring, we simply continue with nil values until we get to the point where the injector is added to the list of injectors. Now the failing injector will be skipped over as if it did not exist on the pod and will not be attempted to inject the disruption and of course clean.

## Code Quality Checklist

## Testing

    - `x`
    - [X] locally.

### DIFF
--- a/cli/injector/disk_pressure.go
+++ b/cli/injector/disk_pressure.go
@@ -47,7 +47,7 @@ var diskPressureCmd = &cobra.Command{
 			}
 
 			if inj == nil {
-				log.Infow("Skipping current injector...")
+				// skipping this injector due because path cannot be found on specified container
 				continue
 			}
 
--- a/cli/injector/disk_pressure.go
+++ b/cli/injector/disk_pressure.go
@@ -47,7 +47,7 @@ var diskPressureCmd = &cobra.Command{
 			}
 
 			if inj == nil {
-				// skipping this injector due because path cannot be found on specified container
+				// skipping this injector because path cannot be found on specified container
 				continue
 			}
 
--- a/cli/injector/disk_pressure.go
+++ b/cli/injector/disk_pressure.go
@@ -46,6 +46,11 @@ var diskPressureCmd = &cobra.Command{
 				log.Fatalw("error initializing the disk pressure injector", "error", err)
 			}
 
+			if inj == nil {
+				log.Infow("Skipping current injector...")
+				continue
+			}
+
 			injectors = append(injectors, inj)
 		}
 	},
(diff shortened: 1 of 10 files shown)


injector/disk_pressure.go
```diff
@@ -46,6 +46,11 @@ func NewDiskPressureInjector(spec v1beta1.DiskPressureSpec, config DiskPressureI
 		if err != nil {
 			return nil, fmt.Errorf("error initializing disk informer: %w", err)
 		}
+
+		if err == nil && len(path) == 0 {
+			config.Log.Infow(fmt.Sprintf("Could not apply injector on container: %s; %s not found on this targeted container.", config.Container.Name(), spec.Path))
```
Devatoria, line 51:
```suggestion
			config.Log.Warnf("could not apply injector on container: %s; %s not found on this targeted container.", config.Container.Name(), spec.Path)
```

Switched to warn level + format signature instead of `Warnw + fmt.Sprintf`

cli/injector/disk_pressure.go
```diff
@@ -46,6 +46,11 @@ var diskPressureCmd = &cobra.Command{
 				log.Fatalw("error initializing the disk pressure injector", "error", err)
 			}
 
+			if inj == nil {
+				log.Infow("Skipping current injector...")
```
Devatoria, line 50:
```suggestion
```

I don't think this log line is useful as it does not give any interesting context (the warn log on injector creation should be enough)

Devatoria (changes requested):

cli/injector/disk_pressure.go
Azoam: I've added a comment instead for future context as to why we continue here.

cli/injector/disk_pressure.go
```diff
@@ -46,6 +46,11 @@ var diskPressureCmd = &cobra.Command{
 				log.Fatalw("error initializing the disk pressure injector", "error", err)
 			}
 
+			if inj == nil {
+				// skipping this injector due because path cannot be found on specified container
```
ptnapoleon, line 50: should be either `due to` or `because`, no?

ptnapoleon:
Worth adding in docs somewhere? Possibly also modifying an e2e-test?

ptnapoleon (approved):

PR #419, opened 2021-09-28
[Chaos-20] Notify on Gitlab Job Failures Correctly

Azoam:
## What does this PR do?

- [ ] Adds new functionality
- [X] Fixes a bug

We added a way to notify on certain job failures in our gitlab ci. Turns out `on_failure` is not stage specific, but means "on_failure of ANY previous stage" not the current stage and previously believed. Therefore the bug occurring was when the build stage would fail (which often times happens with new branches), the pre-release and release notifiers would go off because build is a previous stage to them. Pre-release and Release notifier were requested to go off no matter what branch they were run in, therefore all build jobs would be notified which is not intended.

The solution here is to create 5 notifiers in their own notify stage towards the end of the pipeline. Each notifier corresponds to a specific job under the big three umbrellas Build, Pre-Release, and Release. Release contains 3 jobs of importance therefore we have 5 jobs if you include 1 job for both Build and Pre-Release. The strategy we are using here which will hopefully solve this issue is using the `needs` keyword. This keyword should make it so that notifiers will ONLY go off if their specified job fails. Once that specified job fails, the notifier in next on the list to run. If the job never runs, neither does the notifier.

We are also adding that all manual jobs do NOT ignore failure. When ignoring failures with `allow_failures` it makes it difficult to notify on those failures. This should be fine as the manual jobs (pre-release, release)  make sense to fail the pipeline if one of them fails. If pre-release fails, then release shouldn't be run on that pipeline without fixing whatever caused pre-release to fail.

## Code Quality Checklist

## Testing

I tested manually by commiting a few test changes to the branch to see the changes were correct. Please check the corresponding gitlab pipelines for each testing commit to see how this would work in practice.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,7 +28,7 @@ stages:
   script:
     - BUILD_URL="$CI_PROJECT_URL/pipelines/$CI_PIPELINE_ID"
     - 'MESSAGE_TEXT=":siren: | A Job in the $CI_PIPELINE_NAME Pipeline Failed | [ $CI_PROJECT_NAME ][ $CI_COMMIT_REF_NAME ] [ <$BUILD_URL|$CI_PIPELINE_ID> ]   :siren:"'
-    - postmessage "#test-chaosnotifier2" "$MESSAGE_TEXT"
+    - postmessage "#chaos-ops" "$MESSAGE_TEXT"
 
 # CI image
 .docker-runner: &docker-runner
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -203,23 +206,38 @@ release-docker-hub-tag:
   variables:
     TAG: "${CI_COMMIT_TAG}"
 
+# notify slack when the pre-release job fails
 slack-notifier-pre-release.on-failure:
   extends: .slack-notifier-base
   variables:
     CI_PIPELINE_NAME: "Pre-Release"
```
Devatoria, line 213: Can't we use the predefined variable `CI_JOB_STAGE` instead of setting the pipeline name here?

Azoam: CI_JOB_STAGE would be equivalent to notify because the notifications happen at the notify stage, so it wouldnt work as intended.

PR #421, opened 2021-09-30
Controller Warning Slack Notifier

nathantournant:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug

## Please briefly describe your changes as well as the motivation behind them:
This PR creates a Notifier interface to send kubernetes events to the Slack API, with a useful format. I allows to leverage kubernetes events creation mechanisms (anti-spam, lifecycle integration, tooling, etc.) as a way to notify users about unexpected behaviors in their disruption, directly as a Slack DM.
Currently, users have little to no information when a disruption goes wrong in a controlled way (stuck in removal, no target found, etc.). With this PR, any warning event from the controller (which we generate ourselves and have total control over) will be sent to the user as a Slack DM, along with information about the disruption (where to find it, number of targets, etc.) so they know quickly they have to dig and know where to start.

## Code Quality Checklist

## Testing

    - starting disruptions with no target will generate a warning event. On a cluster, you will receive a Slack DM for it. On minikube, you can take a look at the logs and search for `NOOP: warning` to assert the warning was sent to the console.
    - Any configuration mistake will generate benign, easy to debug errors at controller launch (no token found, sink doesn't exist, slack auth failed, etc.)
    - Test version will be deployed to `plain1b.us1.staging.dog` at the release of this PR so you can try it there. There will be a dedicated branch in chaos-k8s-ressources to observe the differences in deployment files.
    - [X] locally.
    - [X] as a canary deployment to a cluster.


nathantournant:
Noted bug: on live clusters the cluster name doesn't appear on the slack DM, will be fixed very soon

nathantournant:
Test image deployed on `plain1b.us1.staging.dog`, you may test numb disruptions (eg. with no valid target) to check out the DM !
On k8s-chaos-ressources, the branch event-slack-notifier was deployed using the command `bzl run //k8s/chaos-controller:plain1b.us1.staging.dog -- apply`

chart/install.yaml
```diff
@@ -44,6 +44,7 @@ data:
       metricsAddr: "127.0.0.1:8080"
       leaderElection: true
       metricsSink: "noop"
+      slackTokenFilePath: "/etc/chaos-slack-token"
```
Devatoria, line 47: This should be autogenerated from the chart but I can't see anything specifying this in the chart

chart/templates/configmap.yaml
```diff
@@ -14,6 +14,7 @@ data:
       metricsAddr: "127.0.0.1:8080"
       leaderElection: true
       metricsSink: {{ .Values.controller.metricsSink | quote }}
+      notifierSinks: {{ .Values.controller.notifierSinks }}
```
Devatoria, line 17: Could you add a per-notifier configuration here?
I would expect a notifier, for instance Slack, to be configurable like:

```yaml
controller:
  notifiers:
    noop:
      enabled: false
    slack:
      enabled: true
      tokenFilePath: /etc/chaos-controller/slack_token
      context:
        clusterName: foo
```

or with common context values

```yaml
controller:
  notifiers:
    context:
      clusterName: foo
    noop:
      enabled: false
    slack:
      enabled: true
      tokenFilePath: /etc/chaos-controller/slack_token
```

chart/values.yaml
```diff
@@ -13,6 +13,7 @@ images: # images and tag to pull for each component of the stack
 controller:
   deleteOnly: false # enable delete-only mode
   metricsSink: noop # metrics driver (noop or datadog)
+  notifierSinks: noop # notifier sinks list (slack, noop, both) - ex: "slack,noop"
```
Devatoria, line 16: Following my comment on the configmap file, you should be able to customize those values from there.

eventbroadcaster/eventbroadcaster.go
```diff
@@ -0,0 +1,28 @@
...
+
+func eventMessageAggregator(event *v1.Event) string {
+	return "Agg: " + event.Message
+}
+
+func EventBroadcaster() record.EventBroadcaster {
+	correlator := record.CorrelatorOptions{
+		MaxEvents:            2,
+		MaxIntervalInSeconds: 60,
+		// BurstSize:            10,
```
Devatoria, line 21: Expected to keep that commented?

chart/install.yaml
Devatoria: Also I would probably not expect any notifiers to be enabled by default or eventually the noop one.

eventbroadcaster/notifiersink.go
```diff
@@ -0,0 +1,109 @@
...
+			broadcaster.StartRecordingToSink(&NotifierSink{client: client, notifier: notifier})
+		}
+	}
+
+	return resError
+}
+
+func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
+	dis, err := s.getDisruption(event)
+
```
Devatoria, line 54:
```suggestion
```

eventbroadcaster/notifiersink.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return event, nil
+	}
+
+	err = s.parseEventToNotifier(event, dis)
+
+	if err != nil {
```
Devatoria, line 61:
```suggestion
	if err := s.parseEventToNotifier(event, dis); err != nil {
```

eventnotifier/slack/slack.go
```diff
@@ -0,0 +1,103 @@
...
+type Notifier struct {
+	client slack.Client
+}
+
+// New Slack Notifier
+func New(tokenFilePath string) (*Notifier, error) {
+	not := &Notifier{}
+	tokenfile, err := os.Open(filepath.Clean(tokenFilePath))
+
+	if err != nil {
```
Devatoria, line 31:
```suggestion
	not := &Notifier{}
	
	if tokenfile, err := os.Open(filepath.Clean(tokenFilePath)); err != il {
```

Devatoria: Same suggestion for other error handlings please

Devatoria (changes requested):

chart/templates/configmap.yaml
nathantournant: Done

chart/templates/configmap.yaml
```diff
@@ -14,6 +14,14 @@ data:
       metricsAddr: "127.0.0.1:8080"
       leaderElection: true
       metricsSink: {{ .Values.controller.metricsSink | quote }}
+      notifiers:
+        common:
+          clustername: {{ .Values.controller.notifiers.common.clusterName | quote }}
```
Devatoria, line 19:
```suggestion
          clusterName: {{ .Values.controller.notifiers.common.clusterName | quote }}
```

It can sound as a detail but better keeping the same naming everywhere

main.go
```diff
@@ -136,6 +140,18 @@ func main() {
 	pflag.StringVar(&cfg.Controller.MetricsSink, "metrics-sink", "noop", "Metrics sink (datadog, or noop)")
 	handleFatalError(viper.BindPFlag("controller.metricsSink", pflag.Lookup("metrics-sink")))
 
+	pflag.StringVar(&cfg.Controller.Notifiers.Common.ClusterName, "notifiers-common-clustername", "", "Cluster Name for notifiers output")
+	handleFatalError(viper.BindPFlag("controller.notifiers.common.clustername", pflag.Lookup("notifiers-common-clustername")))
```
Devatoria, line 144:
```suggestion
	handleFatalError(viper.BindPFlag("controller.notifiers.common.clusterName", pflag.Lookup("notifiers-common-clustername")))
```

eventbroadcaster/notifiersink.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
+	dis, err := s.getDisruption(event)
+
+	if err != nil {
+		return event, nil
+	}
+
+	if err = s.parseEventToNotifier(event, dis); err != nil {
+		fmt.Println(err)
```
Devatoria, line 51: Expected? It'll print an error on standard output with no formatting while I would expect more something handled by a logger here.

eventbroadcaster/notifiersink.go
```diff
@@ -0,0 +1,97 @@
...
+	notifiers, err := eventnotifier.GetNotifiers(notifiersConfig)
+
+	for _, notifier := range notifiers {
+		broadcaster.StartRecordingToSink(&NotifierSink{client: client, notifier: notifier})
+	}
+
+	return
+}
+
+func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
```
Devatoria, line 43: What happens when this function returns an error? Does it retry or stop here? Because we'll likely reach this case if the username stored in the disruption status is invalid (for instance when a disruption is created by something else than a human, like a workflow).

Devatoria (changes requested):

eventbroadcaster/notifiersink.go
nathantournant: Nothing happens: the error is generated in the slack notifier, gets back to the NotifierSink.Create function, which just absorbs the error (the error from parseEventToNotifier is just ignored, as it would actually generate a "Event Rejected" error log)

nikos912000:
This feature is very useful, thanks for that!

I was wondering if you are planning to add support for HTTP sinks. I implemented something similar in [kube-monkey](https://github.com/asobti/kube-monkey#notifications) back in time which works for any HTTP collectors, including Slack, as this can be treated as a simple REST call.

PR #439, opened 2021-11-17
CHAOS-186: Chaos pods should remove their own finalizer

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Chaos Injector Pods now attempt to remove their own finalizer if they are sure they've properly cleaned up
- Contains two unrelated changes: (1) Fixes some typos and (2) Puts all the args for chaosapi.AppendArgs into a single struct, as the header for that method was getting ridiculous

## Code Quality Checklist

## Testing

I've tested locally, but still have to test in a larger, live kubernetes environment.


controllers/disruption_controller.go
```diff
@@ -533,7 +524,7 @@ func (r *DisruptionReconciler) handleChaosPodsTermination(instance *chaosv1beta1
 		if removeFinalizer || ignoreStatus {
 			r.log.Infow("chaos pod completed, removing finalizer", "target", target, "chaosPod", chaosPod.Name)
 
-			controllerutil.RemoveFinalizer(&chaosPod, chaosPodFinalizer)
+			controllerutil.RemoveFinalizer(&chaosPod, chaostypes.ChaosPodFinalizer)
```
takakonishimura, line 527: Are you still doing leaving this line in for the case where the pod didn't clean itself up?

chart/templates/injector_role.yaml
```diff
@@ -14,6 +14,8 @@ rules:
   - pods
   verbs:
   - list
+  - get
+  - update
```
takakonishimura, line 18: Just to confirm, we need `get` and `update` in order to find the pod and update it once we remove its finalizer correct?

ptnapoleon: Exactly, yes.

controllers/disruption_controller.go
ptnapoleon: Essentially, yes. When would that be? I can't really think of a time. However, since we got here, we're "sure" this is safe, so why not do or try it? E.g., maybe somebody sigkilled a chaos pod, but its target is gone now, so we're sure it's safe to clean up. 

I wanted to supplement this check, not replace it.

takakonishimura: And if the finalizer was removed by the pod already, most of the time it won't even be picked up by this loop.

I follow up with this unlikely scenario:

- If unless it wasn't cleaned up by the time this loop hits, so it was erroneously picked up despite its finalizer already having been removed by the new injector code, then would this code path gracefully give up?
- How about the other direction? Could this code path start removing the pod's finalizer before the new injector code removes the finalizer? Would that cause a bad outcome on the injector code side?

ptnapoleon: So in both cases, I would expect to get error logs (from whichever of the two updated second), but that we will eventually, successfully remove the finalizer in the controller on a later reconciliation. I will give some thought to how we can log at WARN instead, but without running into a scenario where we never actually error and leave it stuck

takakonishimura: Okay sounds good.

takakonishimura:
I feel pretty confident about this PR logically and have no issues with the design. Slightly worried the double remove-finalizer will have some weird behavior but not quite sure how to try to reproduce that concern. :)

controllers/disruption_controller.go
Devatoria: Yes I think it is a good thing to garbage collect chaos pods in case they couldn't remove their own finalizer which can happen in a lot of cases (a pending chaos pod, a sigkilled chaos pod as Philip said, etc.). We could eventually simplify the conditions above this code by removing the `PodSucceeded` case so we do not have a collision with what's happening in the chaos pod (since it removes its finalizer only when it's succeeded anyway).

env/env.go
```diff
@@ -14,4 +14,5 @@ const (
 	InjectorMountSysrqTrigger = "CHAOS_INJECTOR_MOUNT_SYSRQ_TRIGGER"
 	InjectorTargetPodHostIP   = "TARGET_POD_HOST_IP"
 	InjectorChaosPodIP        = "CHAOS_POD_IP"
+	Hostname                  = "HOSTNAME"
```
Devatoria, line 17: Is it a default environment variables for pods? I could not find any information about that in the Kubernetes documentation. Otherwise, or if we are not sure, we can still create the chaos pods with an environment variable like:
```yaml
      env:
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```
[as suggested in the Kubernetes documentation about environment variables](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables).

cli/injector/main.go
```diff
@@ -354,6 +360,18 @@ func cleanAndExit(cmd *cobra.Command, args []string) {
 		log.Fatalw(fmt.Sprintf("disruption cleanup failed on %d injectors (comma separated errors)", len(errs)), "errors", combined.String())
 	}
 
+	pod, err := configs[0].K8sClient.CoreV1().Pods(chaosNamespace).Get(context.Background(), os.Getenv(env.Hostname), metav1.GetOptions{})
+	if err != nil {
+		log.Errorw("couldn't GET this pod in order to remove its finalizer", "pod", os.Getenv(env.Hostname), "err", err)
```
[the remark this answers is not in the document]
ptnapoleon: Isn't it the opposite though? If this fails, we know that we cleaned up, we just weren't able to remove the finalizer, so we are happy that the controller sees the exit code as 0 and removes the finalizer instead?

env/env.go
ptnapoleon: oh, hmm, good point, i will change

api/disruption_kind.go
```diff
@@ -20,41 +20,58 @@ type DisruptionKind interface {
...
 		// basic args
-		"--metrics-sink", sink,
-		"--level", string(level),
-		"--target-container-ids", strings.Join(targetContainerIDs, ","),
-		"--target-pod-ip", targetPodIP,
+		"--metrics-sink", xargs.MetricsSink,
+		"--level", string(xargs.Level),
+		"--target-container-ids", strings.Join(xargs.TargetContainerIDs, ","),
+		"--target-pod-ip", xargs.TargetPodIP,
+		"--chaos-namespace", xargs.InjectorServiceAccountNamespace,
```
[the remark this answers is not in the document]
ptnapoleon: Let's discuss this outside of this PR? `namespace` is too generic, as we have three different namespaces to think of: the controller's, the injector's (usually the controller's but doesnt have to be) and the disruption's.

cli/injector/main.go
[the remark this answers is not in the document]
Devatoria: There is a (maybe small but still) possibility that the pod fails to update the pod with the removed finalizer (let's say because of a resource version mismatch), exits properly but still triggers the k8s 1.20 bug with the pod exiting 137 and so the controller will not clean this, right?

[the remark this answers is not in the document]
Devatoria: Also if we decide to keep an error log here we should exit, otherwise we'll continue with the pod finalizer update with a nil object as error does not exit.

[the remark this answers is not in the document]
ptnapoleon: Does the new backoff logic change your thoughts? We now "exit" early if the GET fails, we retry 3 times total. Admittedly, it is still possible to fail all of those, exit properly, get the 137, and then we don't get cleaned up, but I dont see how to eliminate that chance, and this is still an improvement?

env/env.go
```diff
@@ -14,4 +14,5 @@ const (
 	InjectorMountSysrqTrigger = "CHAOS_INJECTOR_MOUNT_SYSRQ_TRIGGER"
 	InjectorTargetPodHostIP   = "TARGET_POD_HOST_IP"
 	InjectorChaosPodIP        = "CHAOS_POD_IP"
+	MyPodName                 = "MY_POD_NAME"
```
Devatoria, line 17:
```suggestion
	ControllerPodName                 = "CONTROLLER_POD_NAME"
```
If that's ok for you, because I added almost the same variable for node name and it is named `CONTROLLER_NODE_NAME`, so we keep the same format?

ptnapoleon: But this definitely isn't the Controller Pod Name though?

ptnapoleon: Changed to Injector

PR #445, opened 2021-11-23
Add a safeguard to exclude the controller node from targeted nodes wh…

Devatoria:
…en enabled

## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- The PR adds a safeguard that excludes the node running the controller from the list of targeted nodes with it is enabled so we do not apply a disruption to the controller node by mistake (which could lead to an expected behavior like not being able to delete the related disruption)
- Safeguards are enabled by default, but disabled locally through `make install`

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


ptnapoleon (approved):
Looks very suitable to me! This doesnt prevent targeting pods on the controller's node though, so we could still nodeFailure the controller, is that alright?

Devatoria:
> 



> Looks very suitable to me! This doesnt prevent targeting pods on the controller's node though, so we could still nodeFailure the controller, is that alright?

Good catch, we could handle that easily as well by ensuring that the targeted pod node name is not the controller node name (maybe only focused to node failures though?).

PR #447, opened 2021-11-29
Chaosli: Embedding and versioning the Disruption API

nathantournant:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Major QOL changes for `chaosli`, making it *actually* usable on any machine (or at least DD's)
- In the current state, the `chaosli` is incapable of versioning other than through `brew`. This PR adds the `chaosli version` command, from the build `-ldflags` inclusion, to the added cobra command
- More importantly, in its current state, the `chaosli` fetches the locally installed chaos-controller code to get the library, and runs validation against it (through `ddmark`). This poses important problems as the code in `$GOPATH/src/github.com/DataDog/chaos-controller` can be invalid, on a different version, or just not on disk. Somehow the API absolutely needs to be strictly versioned with the binary.
- The most efficient way to go that was found is to package the library in the binary with `pkger` and have the program copy the API someplace practical/unique in the `$GOPATH` of the local machine. Unless you dive into deep configuration, the go `package` package loads from either `$GOROOT` (which usually requires `sudo`) or `$GOPATH`.
- This PR makes `chaosli` copy the API versioned by the binary to the path `$GOPATH/src/chaosli-api-lib/<version>/v1beta1` on every run, and tries to cleanup `$GOPATH/src/chaosli-api-lib` after every run that doesn't `log.Fatal()`. This behavior is completely questionnable and may be discussed here, although I believe it's the most effective to assuredly get the required result.

## Code Quality Checklist

## Testing

    - [X] by adding new `unit` tests or `end-to-end` tests.
    - run the docker file with `make chaosli-test` to assert the library is copied and found on a fresh go environment
    - [X] locally.


cli/chaosli/README.md
```diff
@@ -13,6 +13,9 @@ The Chaos Controller CLI, `chaosli`, is meant to help create a more user friendl
 
 Download the `chaosli` binary for your operating system and architecture [from the latest release](https://github.com/DataDog/chaos-controller/releases/latest).
 
+##### Requirements (important):
+- Have `go` installed (v. 1.11 or greater) and the local `$GOPATH` environment variable setup
```
ptnapoleon, line 17: Can we just require 1.16 or higher here? Why support a lower version for this CLI than we want to support for the controller?

cli/chaosli/README.md
```diff
@@ -13,6 +13,9 @@ The Chaos Controller CLI, `chaosli`, is meant to help create a more user friendl
 
 Download the `chaosli` binary for your operating system and architecture [from the latest release](https://github.com/DataDog/chaos-controller/releases/latest).
 
+##### Requirements (important):
```
ptnapoleon, line 16: nit: I think `(important)` is redundant after "Requirements"

cli/chaosli/cmd/root.go
```diff
@@ -22,6 +24,8 @@ import (
 	"github.com/spf13/viper"
 )
 
+var VERSION string
+var APILIBPATH string = fmt.Sprintf("chaosli-api-lib/%v/v1beta1", VERSION)
```
ptnapoleon, line 28: A few questions here:
- `API_LIB_PATH` please?
- How is VERSION set?
- Depending on how VERSION is set, aren't we interpolating it into APILIBPATH too soon?

cli/chaosli/README.md
nathantournant: Fair enough haha

cli/chaosli/cmd/root.go
nathantournant: - Will edit the variable name
- VERSION is set through the -ldflags, and in practice will be done from the k8s-chaos-ressources Makefile command you commented on earlier, where the release tag is set as VERSION
- very good question, I did not think of kubebuilder versioning/naming while doing this - would love to know what's coming and talk about it very soon

cli/chaosli/README.md
nathantournant: I see the CLI as an executable that shouldn't require the chaos-controller code/tech nor its requirements to run - so I set it as the lowest possible value 
But yes we could 100% ask for 1.16 that's definitely future-proof

ptnapoleon: I agree that's a very sensible point, but it's more that I don't want to install or think about any golang versions that I didn't already have to? 1.11 is EOL

cli/chaosli/cmd/root.go
ptnapoleon: The -ldflags can directly set a variable like this? I didn't know that, that's magic and cool, please comment that above VERSION then

cli/chaosli/README.md
nathantournant: Done

cli/chaosli/cmd/root.go
nathantournant: Fixed

cli/chaosli/cmd/root.go
```diff
@@ -22,6 +24,8 @@ import (
 	"github.com/spf13/viper"
 )
 
+var VERSION string
+var API_LIB_PATH string = fmt.Sprintf("chaosli-api-lib/v1beta1/%v", VERSION)
 var cfgFile string
```
Devatoria, line 29: Is it expected to have variable names which do not respect naming standards (I'd have expected `Version` and `APILibPath` or `ApiLibPath` here?

cli/chaosli/cmd/root.go
```diff
@@ -88,6 +95,71 @@ func initConfig() {
...
 
+// initLibrary copies the binary-embedded disruption API into a custom folder in GOPATH
+func initLibrary() {
+	if _, isGoInstalled := os.LookupEnv("GOPATH"); !isGoInstalled {
+		log.Fatal("Setup error: please make sure go (1.11 or higher) is installed and the GOPATH is set")
+	}
+
+	err := os.Setenv("GO111MODULE", "off")
+
+	if err != nil {
```
Devatoria, line 106:
```suggestion
	if err := os.Setenv("GO111MODULE", "off"); err != nil {
```

cli/chaosli/cmd/root.go
```diff
@@ -88,6 +95,71 @@ func initConfig() {
...
+		log.Fatal(err)
+	}
+
+	folderPath := fmt.Sprintf("%v/src/%v/", os.Getenv("GOPATH"), API_LIB_PATH)
+	err = os.MkdirAll(folderPath, 0750)
+
+	if err != nil {
```
Devatoria, line 113:
```suggestion
	if err := os.MkdirAll(folderPath, 0750); err != nil {
```

cli/chaosli/cmd/root.go
```diff
@@ -88,6 +95,71 @@ func initConfig() {
...
+		log.Fatal(err)
+	}
+
+	err = pkger.Walk("github.com/DataDog/chaos-controller:/api/v1beta1", func(path string, info os.FileInfo, err error) error {
```
Devatoria, line 117: Maybe a small comment on what this anonymous function does would be helpful for further maintenance

Devatoria (changes requested):

cli/chaosli/cmd/root.go
nathantournant: I'll conform to the naming standards then

nathantournant: Done, variable is now APILibPath

cli/chaosli/cmd/root.go
```diff
@@ -88,6 +96,70 @@ func initConfig() {
 	}
 }
 
+// initLibrary copies the binary-embedded disruption API into a custom folder in GOPATH
+func initLibrary() {
+	if _, isGoInstalled := os.LookupEnv("GOPATH"); !isGoInstalled {
+		log.Fatal("Setup error: please make sure go (1.11 or higher) is installed and the GOPATH is set")
```
ptnapoleon, line 102:
```suggestion
		log.Fatal("Setup error: please make sure go (1.16 or higher) is installed and the GOPATH is set")
```

ptnapoleon (approved):

Devatoria (approved):

PR #453, opened 2021-12-03
CHAOS-181: Try to recover better from stuck nodeFailure disruptions

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- First change is to always log the target node in the injector. This can be redundant in node level disruptions, but is very useful when debugging pod level disruptions
- Second change is to check, in the case of a nodeFailure disruption, if the node hosting the target pod still exists. I don't think this change is sufficient to fix all of the failure cases we've seen, but it does seem to catch the most common one.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


LICENSE-3rdparty.csv
```diff
@@ -185,8 +185,6 @@ github.com/markbates/pkger,github.com/markbates/pkger,MIT
 github.com/markbates/pkger,github.com/markbates/pkger/here,MIT
 github.com/markbates/pkger,github.com/markbates/pkger/internal/maps,MIT
 github.com/markbates/pkger,github.com/markbates/pkger/pkging,MIT
-github.com/markbates/pkger,github.com/markbates/pkger/pkging/embed,MIT
```
ptnapoleon, line 188: `make deps` removed these, not sure why?

targetselector/running_target_selector.go
```diff
@@ -153,6 +154,16 @@ func (r runningTargetSelector) TargetIsHealthy(target string, c client.Client, i
 		if p.Status.Phase != corev1.PodRunning {
 			return errors.New("pod is not Running")
 		}
+
+		// check if pod's node is gone in the case that this was a node failure
+		if instance.Spec.NodeFailure != nil {
+			var n corev1.Node
+			if err := c.Get(context.Background(), client.ObjectKey{Name: p.Spec.NodeName}, &n); err != nil {
+				if apierrors.IsNotFound(err) {
```
ptnapoleon, line 162: Put this in as a safeguard

targetselector/running_target_selector.go
```diff
@@ -153,6 +154,16 @@ func (r runningTargetSelector) TargetIsHealthy(target string, c client.Client, i
...
+					return err
+				}
```
takakonishimura, line 164: What does it mean to have an `err` that is not the `NotFound` error?

Devatoria: It can be any kind of error while trying to get the resource (like apiserver timeout for instance).

targetselector/running_target_selector.go
Devatoria: I'm afraid about the behavior of get returning an error which is not a not found error? We seem to skip it intentionally, is it not dangerous?

ptnapoleon: I felt it would be more dangerous to return those? Returning an error here means that we can ignore the status of the injector pod and directly clean up. I wouldn't want to do that just because of a timeout or other transient error. What do you think?

Devatoria: I agree with that, though we do it [here](https://github.com/DataDog/chaos-controller/pull/453/files/1c4849634e8b287e8803b86f32295db7f80c489b#diff-fbedc5033a333cbdab5a7469df68499adef8dd7b1e7e3eca5824be728c0eb268R149) so I guess we should decide to do either one or the other but not to mix behaviors (not telling that the already existing one is good)

ptnapoleon: I changed my mind here? The sole callsite of this function already handles the different error types correctly, so I don't think extra filtering inside the function is helping us at all. I'm gonna switch this to `return err` as well to stay consistent as you'd like

PR #470, opened 2022-01-10
CHAOS-40: Delete all orphaned chaos pods

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- If we have an orphan, something went wrong, so we always log these at warn, and emit a metric that we can alert on. We end up getting a lot of node failure orphans. If the disruption itself is gone, I think we can go ahead and just try to delete these (i can make it a force delete too). changing handleChaosPodsTermination wasn't very smooth, as it updates the instance too often

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


controllers/disruption_controller.go
```diff
@@ -431,6 +434,40 @@ func (r *DisruptionReconciler) cleanDisruption(instance *chaosv1beta1.Disruption
...
+
+			continue
+		}
+
+		if err := r.Client.Delete(context.Background(), &chaosPod); err != nil {
+			r.log.Errorw("error deleting orphaned chaos pod", "error", err, "chaosPod", chaosPod.Name)
+
+			continue
+		}
+		r.handleMetricSinkError(r.MetricsSink.MetricOrphanDeleted([]string{"name:" + req.Name, "namespace:" + req.Namespace}))
```
ptnapoleon, line 465: Right now we're incrementing one metric for each found orphan, and one for each deleted orphan. We could switch to one increment every time we reconcile loop for orphans (rather than for each orphan)? Also possible to increment for every attempted delete, rather than every successful

controllers/disruption_controller.go
```diff
@@ -431,6 +434,40 @@ func (r *DisruptionReconciler) cleanDisruption(instance *chaosv1beta1.Disruption
...
+
+	chaosPods, err := r.getChaosPods(nil, ls)
+	if err != nil {
+		return err
+	}
+
+	for _, chaosPod := range chaosPods {
+		r.handleMetricSinkError(r.MetricsSink.MetricOrphanFound([]string{"name:" + req.Name, "namespace:" + req.Namespace}))
+
+		r.log.Warnw("orphaned chaos pod detected, will attempt to delete", "chaosPod", chaosPod.Name)
```
ptnapoleon, line 451: Error log gives a stack trace which is unhelpful here. We can rely on the metric to detect issues, rather than needing this to be error instead of warn

controllers/disruption_controller.go
```diff
@@ -97,6 +97,9 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}()()
 
 	if err := r.Get(context.Background(), req.NamespacedName, instance); err != nil {
+		if client.IgnoreNotFound(err) == nil {
```
ptnapoleon, line 100: Do we want a comment here explaining why this means we must have orphans?

controllers/disruption_controller.go
```diff
@@ -431,6 +437,41 @@ func (r *DisruptionReconciler) cleanDisruption(instance *chaosv1beta1.Disruption
...
+	chaosPods, err := r.getChaosPods(nil, ls)
+	if err != nil {
+		return err
+	}
+
+	for _, chaosPod := range chaosPods {
+		r.handleMetricSinkError(r.MetricsSink.MetricOrphanFound([]string{"name:" + req.Name, "namespace:" + req.Namespace}))
+
+		r.log.Warnw("orphaned chaos pod detected, will attempt to delete", "chaosPod", chaosPod.Name)
+		controllerutil.RemoveFinalizer(&chaosPod, chaostypes.ChaosPodFinalizer)
```
Devatoria, line 455: What if the orphan pod fails to clean up the injected disruption? So far, it'll be deleted and removed without us knowing it. I feel like the lifecycle for those pods should (but probably can't?) be the same as for chaos pods with an attached disruption (deleting them, checking their status for a proper removal, etc).

controllers/disruption_controller.go
Devatoria: I'd either go with a metric for each found orphan or a metric for each call to this function if orphans are found (whatever the amount of orphans).

ptnapoleon: Are you suggesting only one metric total, (in which case I'd prefer one per found orphan), and thus no metric for deleted orphans?

controllers/disruption_controller.go
ptnapoleon: Hmm, great point. Let's reason about the situations in which we even have orphans:

1. Worst Case:
- The controller goes down. Someone then deletes a disruption resource and removes its finalizer. Without the controller to reconcile, the chaos pods are left uncleaned and orphaned.
2. Current Observed Case:
- We have a node failure disruption. After running, the node is gone, but chaos pods remain due to their finalizer. Somehow? `cleanDisruption` must be returning `isCleaned == true`, because the disruption itself cleans up.
3. "Other" Case:
- Overlap of (2), but `cleanDisruption` returns `isCleaned == true`, even though that's false, in which case the disruption has its finalizer removed.


It seems that theoretically we definitely can't just remove the finalizer from these orphans. In practice though, they've already been attempted to be deleted, and are stuck in Terminating, and just have the finalizer stuck for some reason. Maybe we should try deleting again before removing?

controllers/disruption_controller.go
Devatoria: Yup exactly

controllers/disruption_controller.go
```diff
@@ -152,6 +158,11 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		// the injection is being created or modified, apply needed actions
 		controllerutil.AddFinalizer(instance, chaostypes.DisruptionFinalizer)
+		if err := r.Update(context.Background(), instance); err != nil {
```
ptnapoleon, line 161: this was the source of the orphans

Devatoria: Should we keep the `handleOrphanedChaosPods` function now?

examples/pulse_drop.yaml
```diff
@@ -0,0 +1,20 @@
+# Unless explicitly stated otherwise all files in this repository are licensed
```
Devatoria: Not expected I think

controllers/disruption_controller.go
```diff
@@ -428,6 +439,49 @@ func (r *DisruptionReconciler) cleanDisruption(instance *chaosv1beta1.Disruption
...
+	for _, chaosPod := range chaosPods {
+		r.handleMetricSinkError(r.MetricsSink.MetricOrphanFound([]string{"disruption:" + req.Name, "chaosPod:" + chaosPod.Name, "namespace:" + req.Namespace}))
+		target := chaosPod.Labels[chaostypes.TargetLabel]
+		var p corev1.Pod
+
+		r.log.Infow("checking if we can clean up orphaned chaos pod", "chaosPod", chaosPod.Name, "target", target)
+
+		// if target doesn't exist, we can try to clean up the chaos pod
+		if err := r.Client.Get(context.Background(), types.NamespacedName{Name: target, Namespace: req.Namespace}, &p); errors.IsNotFound(err) {
+
```
Devatoria, line 462:
```suggestion
```

controllers/disruption_controller.go
Devatoria: Though it means that there's still the case of someone editing the disruption to manually remove the finalizer and deleting the disruption while there are running chaos pods, right?
If yes, what do you think to implement (in another PR of course, eventually later) a mechanism where the admission webhook would reject a disruption update removing the finalizer if there are existing chaos pods for the related disruption? It would prevent such a scenario + it would probably be helpful as it forces the user to check for failed pods in stuck on removal cases?

ptnapoleon: > If yes, what do you think to implement (in another PR of course, eventually later) a mechanism where the admission webhook would reject a disruption update removing the finalizer if there are existing chaos pods for the related disruption? It would prevent such a scenario + it would probably be helpful as it forces the user to check for failed pods in stuck on removal cases?

I think that's great and I'll file a ticket for it

Devatoria (approved):

PR #473, opened 2022-01-13
Claire/pulsing disruptions

clairecng:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- `x`

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


examples/complete.yaml
```diff
@@ -31,6 +31,9 @@ spec:
     - demo
     - demo2
   count: 1 # number of pods to target or a percentage (1% - 100%)
+  pulse: # optional, activate pulsing disruptions. Available for any disruptions except nodeFailure and containerFailure
```
ptnapoleon, line 34: In addition to modifying complete.yaml, can you please add a distinct example for this feature?

injector/cpu_pressure_test.go
```diff
@@ -71,6 +72,12 @@ var _ = Describe("Failure", func() {
 		inj = NewCPUPressureInjector(spec, config)
 	})
 
+	Describe("getDisruptionKind", func() {
+		It("should return cpu-pressure", func() {
+			Expect(string(inj.GetDisruptionKind())).To(Equal(types.DisruptionKindCPUPressure))
```
ptnapoleon, line 77: I think these tests aren't worth having?

api/v1beta1/disruption_types.go
```diff
@@ -216,6 +224,18 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() (retErr error) {
 		}
 	}
 
+	// Rule: pulsing duration
+	if s.Pulse != nil {
+		// Pulsing disruptions should not be pulsing less than every 10 milliseconds
```
ptnapoleon, line 229: Given we're doing the entire Inject or Clean operation between these intervals, I think we should have a much larger floor, how about 500ms, or at least 250?

cli/chaosli/cmd/create.go
```diff
@@ -560,6 +574,49 @@ func getCount() *intstr.IntOrString {
...
+				return fmt.Errorf("duration must be greather than 10ms")
+			}
+
+			return nil
+		}
+
+		return fmt.Errorf("expected a string response, rather than type %v", reflect.TypeOf(val).Name())
+	}
+
+	if !confirmOption("Would you like your disruptions to be pulsing?", "The default is non pulsing disruptions.") {
```
ptnapoleon, line 596: Could we have a little more help text here, to explain what pulsing is?

cli/injector/main.go
```diff
@@ -329,6 +339,50 @@ func injectAndWait(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if len(pulsingInjectors) > 0 {
+		isInjected := true
+		lastOperationTime := time.Now()
+
+		for { // This loop will inject, wait, clean, wait until a signal is received
+			// Inject all the pulsing disruptions
+			if time.Since(lastOperationTime) >= pulseDormantDuration && !isInjected {
```
ptnapoleon, line 348: I think (either at the end or depending of this function) we can add a small sleep, as we should be able to calculate approximately the next time we need to either re-activate or re-dormant. Currently, we will just run these `if` checks extremely quickly. I think, to avoid an issue with catching the exit signal, it may be possible to put a sleep in the select conditions?

Something like:
```
select {
case <- signals:
case sleepTimerRanOut:
}
```

cli/injector/main.go
```diff
@@ -329,6 +339,50 @@ func injectAndWait(cmd *cobra.Command, args []string) {
...
+				}
+
+				isInjected = !isInjected
+				lastOperationTime = time.Now()
+
+				log.Info("pulsing disruption(s) are active")
+			} else if time.Since(lastOperationTime) >= pulseActiveDuration && isInjected { // Clean all the pulsing disruptions
+				for _, inj := range pulsingInjectors {
+					// start cleanup which is retried up to 3 times using an exponential backoff algorithm
+					if err := backoff.RetryNotify(inj.Clean, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 3), retryNotifyHandler); err != nil {
```
ptnapoleon, line 362: So far in testing I haven't run into an issue where we delete the chaos pod while it's dormant (And thus we'd be cleaning a cleaned target). 

I am a little concerned that we do these 3 retries for each (Inj and Clean), and after that just log an Error and move on. Should we maybe fatal instead? @Devatoria

cli/injector/main.go
clairecng: Wouldn't it prevent the signal to be caught early on? Is it ok to wait for the sleep to finish to be able to remove the disruption?

ptnapoleon: I don't think so? If the sleep is set up as a channel, then we're just blocking until we get either the exit signal or the wake up signal. Am I wrong? This way we're not running code for no reason

cli/injector/main.go
Devatoria: I agree, we should probably exit here to catch the error earlier. What about injection though? Should we do the same thing?

cli/injector/main.go
```diff
@@ -329,6 +339,50 @@ func injectAndWait(cmd *cobra.Command, args []string) {
...
+
+					log.Info("pulsing disruption(s) are in dormant mode")
+
+					sleepDuration = pulseDormantDuration
+				}
+
+				isInjected = !isInjected
+			}
+		}
+	} else if pulseActiveDuration > 0 && pulseDormantDuration > 0 && !errOnInject {
```
Devatoria, line 382: Same as my comment above, not sure that this condition is meaningful if it is prevented by the controller already?

examples/pulse.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: pulse
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 1
+  pulse: # optional, activate pulsing disruptions. Available for any disruptions except nodeFailure and containerFailure
+    activeDuration: 60s # this is the duration of the disruption in an active state, must be a valid time.Duration string, e.g. (300s, 15m25s, 4h) and must be greater than 250ms
+    dormantDuration: 20s # this is the duration of the disruption in a dormant state, must be a valid time.Duration string, e.g. (300s, 15m25s, 4h) and must be greated than 250ms
```
ptnapoleon, line 18: These say must be greater than 250ms, but the code says 500ms
```suggestion
    dormantDuration: 20s # this is the duration of the disruption in a dormant state, must be a valid time.Duration string, e.g. (300s, 15m25s, 4h) and must be greater than 500ms
```

docs/features.md
```diff
@@ -31,6 +31,21 @@ If a `duration` is not specified, then a disruption will receive the default dur
 After a disruption's duration expires, the disruption resource will live in k8s for a default of 15 minutes. This can be configured by altering 
 `controller.expiredDisruptionGCDelay` in the controller's config map.
 
+## Pulse
+
+The `Disruption` spec takes a `pulse` field. It activates the pulsing mode of the disruptions of type `cpu_pressure`, `disk_pressure`, `dns_disruption`, `grpc_disruption` or `network_disruption`.
```
ptnapoleon, line 36:
```suggestion
The `Disruption` spec takes a `pulse` field. It activates the pulsing mode of the disruptions of type `cpu_pressure`, `disk_pressure`, `dns_disruption`, `grpc_disruption` or `network_disruption`. A "pulsing" disruption is one that alternates between an active injected state, and an inactive dormant state. Previously, one would need to manage the Disruption lifecycle by continually re-creating and deleting a Disruption to achieve the same effect
```

What do you think about this? It's "obvious", but I felt it wouldn't hurt to be explicit?

cli/chaosli/cmd/create.go
```diff
@@ -83,6 +84,19 @@ func createSpec() (v1beta1.DisruptionSpec, error) {
 	spec.Selector = getSelectors()
 	spec.Count = getCount()
 
+	isPulsingCompatible := true
+
+	for _, disruptionKind := range spec.GetKindNames() {
+		if disruptionKind == types.DisruptionKindContainerFailure || disruptionKind == types.DisruptionKindNodeFailure {
```
ptnapoleon, line 90: This validation is logically different from what we're performing in https://github.com/DataDog/chaos-controller/pull/473/files#diff-972800d1b1e32baefe4ea8e8b7e3cba2845eac77ce8f9ea9b026d168361d9181R229

Consider the following odd, but valid disruption spec:
```
spec:
  containerFailure:
  network:
    drop: 5
```

If we added a `pulse` field to this, the api would accept it, because we are using one of the pulsing compatible fields. But chaosli would not accept it, because we are using one of the pulsing incompatible fields as well. I'm fairly indifferent as to which direction you'd choose to support, but I do think we need to be consistent

cli/chaosli/cmd/create.go
```diff
@@ -560,6 +574,49 @@ func getCount() *intstr.IntOrString {
...
+func getPulse() *v1beta1.DisruptionPulse {
+	validator := func(val interface{}) error {
+		if str, ok := val.(string); ok {
+			_, err := time.ParseDuration(str)
+			if err != nil {
+				return err
+			}
+
+			duration := v1beta1.DisruptionDuration(str)
+			if duration.Duration().Milliseconds() < 500 {
```
ptnapoleon, line 586: We repeat `500` four times in this file. Could that be pulled out to a constant somewhere and referred to? That would make altering it in the future easier

cli/injector/main.go
```diff
@@ -329,6 +339,50 @@ func injectAndWait(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if len(pulsingInjectors) > 0 {
+		isInjected := true
+		sleepDuration := pulseActiveDuration
+
+		for { // This loop will inject, wait, clean, wait until a signal is received
```
ptnapoleon, line 346: This loop says it begins by injecting. However, haven't we _already_ injected by the time we get here? It looks like we start all the injectors, even the pulsing ones, at https://github.com/DataDog/chaos-controller/pull/473/files#diff-7a828d3d2d4e9739314e9b20ad4cde74c8b9fdcef19e29facad2ecb7f17e1b78R296

In that case, shouldn't we sleep for the dormant duration initially? I think the `if` condition correctly handles the clean/Inj lifecycle, we just start with the wrong duration, afaict

cli/injector/main.go
```diff
@@ -271,6 +275,73 @@ func initExitSignalsHandler() {
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 }
 
+// inject inject all the disruptions using the list of injectors
+func inject(kind string, sendToMetrics bool) (int, string) {
```
Devatoria, line 279: We are already logging errors in this function, can we just return a boolean to tell if the injection (or cleanup) was successful instead?
```suggestion
# returns true if injection succeeded, false otherwise
func inject(kind string, sendToMetrics bool) bool {
```

cli/injector/main.go
```diff
@@ -271,6 +275,73 @@ func initExitSignalsHandler() {
...
+	for _, err := range errs {
+		combined.WriteString(err.Error())
+		combined.WriteString(",")
+	}
+
+	return len(errs), combined.String()
+}
+
+// clean clean all the disruptions using the list of injectors
+func clean(kind string, sendToMetrics bool) (int, string) {
```
Devatoria, line 313: Same comment as above for the boolean instead of errors

cli/injector/main.go
```diff
@@ -283,29 +354,17 @@ func injectAndWait(cmd *cobra.Command, args []string) {
...
-	if !errOnInject {
+	if nbErrorsOnInject == 0 {
+		log.Infof("disruption(s) injected, now waiting for an exit signal")
+
 		if err := ioutil.WriteFile(readinessProbeFile, []byte("1"), 0400); err != nil {
 			log.Errorw("error writing readiness probe file", "error", err)
 		}
 	} else {
-		log.Error("an injector could not inject the disruption successfully, please look at the logs above for more details")
+		log.Errorf("disruption injection failed on %d injectors (comma separated errors)", nbErrorsOnInject, "errors", reason)
```
Devatoria, line 367: We can keep this line if we use the boolean as commented above

cli/injector/main.go
```diff
@@ -337,36 +430,18 @@ func injectAndWait(cmd *cobra.Command, args []string) {
...
-	if len(errs) != 0 {
-		var combined strings.Builder
-
-		for _, err := range errs {
-			combined.WriteString(err.Error())
-			combined.WriteString(",")
-		}
-
-		log.Fatalw(fmt.Sprintf("disruption cleanup failed on %d injectors (comma separated errors)", len(errs)), "errors", combined.String())
+	if nbErrorsOnClean > 0 {
```
Devatoria, line 436: Same thing here, I think that using a simple boolean in the clean function return is enough since we already log errors in that function. We can keep the last fatal log that will exit (non 0) the injector if cleanup failed so we can catch that later.

cli/chaosli/cmd/create.go
```diff
@@ -599,14 +599,14 @@ func getPulse() *v1beta1.DisruptionPulse {
 
 	activeDuration := v1beta1.DisruptionDuration(getInput(
 		"What would be the duration of the disruption in an active state during the pulse? This can be a golang's time.Duration.",
-		"Please specify a golang's time.Duration's >500ms, e.g., \"45s\", \"15m30s\", \"4h30m\".",
+		fmt.Sprintf("Please specify a golang's time.Duration's >%dms, e.g., \"45s\", \"15m30s\", \"4h30m\".", types.PulsingDisruptionMinimumDuration.Milliseconds()),
```
ptnapoleon, line 602: Rather then putting `duration.Milliseconds()` into `%dms`, you can just put `duration` into `%s`, as the duration already includes its type (or it should)

docs/features.md
clairecng: I 100% agree, thanks for the suggestion!

cli/chaosli/cmd/create.go
clairecng: After some talk with Joris, we decided to just not allow the creation of disruptions with the pulse field if one of them is a container failure or a node failure. It also removes some of the complexity in the injector

Devatoria: Yup to me it was very confusing that, in the given example, network disruption would be pulsing but not the container failure one because it is not compatible, and this without warning the user. Rejecting disruptions with pulsing on at least one incompatible failure ensures the user knows that it is not possible.

Devatoria (approved):
LGTM, nice job. I'd wait for one more approval before merging given the PR size though.

cli/injector/main.go
```diff
@@ -329,44 +381,54 @@ func injectAndWait(cmd *cobra.Command, args []string) {
...
+			case <-time.After(sleepDuration):
+				if !isInjected {
+					action = inject
+					sleepDuration = pulseDormantDuration
+				} else {
+					action = clean
+					sleepDuration = pulseActiveDuration
+				}
+
+				if errOnClean := action(cmd.Name(), false); errOnClean {
```
ptnapoleon, line 405: I'd like to make a small complaint. `errOnClean` is too misleading to use here, as we are alternating between Inject and Clean

cli/injector/main.go
```diff
@@ -291,21 +285,79 @@ func injectAndWait(cmd *cobra.Command, args []string) {
...
+
+	if errOnInject {
+		log.Errorf("an injector could not inject the disruption successfully, please look at the logs above for more details")
+	}
+
+	return errOnInject
+}
+
+// clean clean all the disruptions using the list of injectors
+// returns true if cleanup succeeded, false otherwise
```
ptnapoleon, line 310: This comment is backwards, we return true if cleanup had an error, false otherwise. Would it be better to return an error here (or a multi error), rather than a bool?

cli/injector/main.go
```diff
@@ -271,18 +274,9 @@ func initExitSignalsHandler() {
...
-	if !readyToInject {
-		log.Error("an injector could not be configured successfully during initialization, aborting the injection now")
-
-		return
-	}
-
-	log.Infow("injecting the disruption", "kind", cmd.Name())
-
+// inject inject all the disruptions using the list of injectors
+// returns true if injection succeeded, false otherwise
```
ptnapoleon, line 278: This comment is backwards, we return true if cleanup had an error, false otherwise. Would it be better to return an error here (or a multi error), rather than a bool?

ptnapoleon (changes requested):

cli/injector/main.go
clairecng: I fixed the function return value to match the comment. The inject / clean functions were previously returning the list of errors but it was unused in the code after as we're already logging every error inside of the functions.

PR #475, opened 2022-01-14
Safemode [CHAOS-25]

Azoam:
## What does this PR do?

Adds a framework for adding safety nets (modules used to stop a disruption in its track when I known issue/concern will occur when a disruption is run) to the chaos controller while also adding a few safety nets as examples.

Here is an example of a safety net and what the controller's intended response would be:
Currently we have implemented a `SporadicTargets` safety net which attempts to identify environments where the targets in question are constantly being restarted due to some underlying issue (bad configurations, sharp load increasing on startup of a target on a pod, etc.). In a sporadic environment like this, its ideal to first resolve those underlying issues before introducing additional problems, such as starting a disruption. In the case where this safety net is not ignored, when I user attempts to run any disruption in this environment, the controller's response is to simply ignore the disruption and send events to the CRD and the chaos-controller explaining the situation. If the user is confident that the sporadic behavior of the environment is acceptable with additional disruptions, they can simply ignore this safety net and run the disruption again.

- [X] Adds new functionality
- [ ] Fixes a bug

- Added a  generic `safemode` module which represents an interface which other disruptions can build off of.
- Added a `safemode` module for each currently available disruption where safety nets can be defined and utilized.
- Added `Safemode` parameter to the disruption CRD which contains boolean values for ignoring certain safety nets and a larger scope ignoring of all safety nets.
- Added documentation on how to use this new safemode feature
- Added several safety nets for different disruptions that are necessary but also serve as examples for safety nets moving forwards, all of these can be found in `docs/safemode.md` with their description and how to ignore them.

## Code Quality Checklist

## Testing

    - `x`
    - [X] locally.


docs/safemode.md
```diff
@@ -0,0 +1,58 @@
+# Safemode
+
+Safemode represents a number of safety nets we have implemented into the chaos controller to help new and experienced users feel more confident deploying new disruptions to their environments.
+
+The Chaos Controller can be scary to use in production environments, but we learn a lot more running chaos experiments in our production environments. Considering this, having safety nets in place makes the entire process of running chaos experiments in high value environments a little safer.
+
+The disruption config has a parameter appropriately named `safeMode` which represents a spec containing several boolean fields.
+One of the fields, `IgnoreAll` overrides the rest if set to true. It will ignore every safety net.
+We recommend that `IgnoreAll` be set on the cluster level by default, so maintainers of the controller can turn it on by default in production and maybe off in staging environments.
```
ptnapoleon, line 9: Sorry that I'm getting so easily confused by double negatives, but we have a field `IgnoreAll`. We recommend it be ON (so true?) in prod. When set to true, we ignore all safety nets? Isn't that the opposite of what we want in prod?

api/v1beta1/safemode.go
```diff
@@ -0,0 +1,18 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+package v1beta1
+
+// SafemodeSpec represents a spec with a switch turning safemode on and all safety net switches to turn off
+// IgnoreAll, by default, is defined by the cluster the controller is running on
+// If the use is running on a cluster that has IgnoreAll as false, they will have to manually ignore the other safety nets they do not want considered
+// If IgnoreAll is true by default, the user can change its value
+type SafemodeSpec struct {
```
ptnapoleon, line 12: Are these all safe modes that exist by default, and setting these values to `true` disables them? Could we call this `unsafe` in the yaml spec, then, rather than safe?

docs/safemode.md
```diff
@@ -0,0 +1,58 @@
...
+Therefore the controller allows for you to ignore specific safety nets in the Safemode Spec. Checkout out example below to see how to remove certain safety nets.
+Remember that these fields to turn off specific safety nets is only accessible to the user if the `IgnoreAll` field is false. Otherwise it overrides all other safety nets and turns them off.
+Keep in mind that all safety nets are turned on by default when safemode is on (`IgnoreAll` set to False), so all that is necessary to ignore safety nets.
+
+### Safety Nets
+
+| Safety Net  | Type | Description | IgnoreName |
+| ----------- | ----------- | ----------- | ----------- |
+| Namespace-wide Targeting                                  | Generic |  Using generical label selectors (e.x. X) that selects a majority of pods/nodes in a namespace as a target to inject a disruption into                     | ignoreCountToLarge        |
+| Sporadic Targets                                          | Generic |  In a volatile environment where targets are being terminated and created sporadically, disruptions should not be allowed to continue disrupting           | ignoreSporadicTargets     |
```
ptnapoleon, line 25: I feel this is too situational and shouldn't be enabled by default

docs/safemode.md
```diff
@@ -0,0 +1,58 @@
...
+metadata:
+  name: network-ingress
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  safeMode:
+    ignoreAll: false
+    ignoreNoPortOrHost: true
```
ptnapoleon, line 44: Why do we need to ignore this if we aren't violating it in this example?

api/v1beta1/safemode.go
```diff
@@ -0,0 +1,18 @@
...
+	IgnoreAll                 bool `json:"ignoreAll,omitempty"`
+	IgnoreCountNotTooLarge    bool `json:"ignoreCountNotToLarge,omitempty"`
```
ptnapoleon, line 14: Typo in `TooLarge` in the json here

docs/safemode.md
```diff
@@ -0,0 +1,58 @@
...
+Because the list of safety nets to be implemented will grow in the future, there will surely be overlap with safety nets which will make it difficult for a user who is confident a specific safety net is not necessary but unsure if others will be.
+Therefore the controller allows for you to ignore specific safety nets in the Safemode Spec. Checkout out example below to see how to remove certain safety nets.
+Remember that these fields to turn off specific safety nets is only accessible to the user if the `IgnoreAll` field is false. Otherwise it overrides all other safety nets and turns them off.
+Keep in mind that all safety nets are turned on by default when safemode is on (`IgnoreAll` set to False), so all that is necessary to ignore safety nets.
+
+### Safety Nets
+
+| Safety Net  | Type | Description | IgnoreName |
+| ----------- | ----------- | ----------- | ----------- |
+| Namespace-wide Targeting                                  | Generic |  Using generical label selectors (e.x. X) that selects a majority of pods/nodes in a namespace as a target to inject a disruption into                     | ignoreCountToLarge        |
```
ptnapoleon, line 24: The json value here doesn't match what's in the spec

safemode/safemode.go
```diff
@@ -0,0 +1,255 @@
...
+
+type Safemode interface {
+	// CreationSafetyNets will look at all the specific safety nets corresponding to the creation of a new disruption
+	CreationSafetyNets() ([]string, error)
+	// GetTypeSpec will grab the necessary spec depending on the type (e.g. SafemodeNetwork will grab NetworkDisruptionSpec)
+	// and grab the disruption itself for data such as the kubernetes namespace the disruption is running on
+	GetTypeSpec(disruption v1beta1.Disruption)
+	// GetKubeClient grab the kube client for functions that require state information from k8s system
+	GetKubeClient(client client.Client)
+	// Init will run GetTypeSpec and GetKubeClient together to cimplify further code
```
ptnapoleon, line 29:
```suggestion
	// Init will run GetTypeSpec and GetKubeClient together to simplify further code
```

safemode/safemode.go
```diff
@@ -0,0 +1,255 @@
...
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// safetyNetCountNotTooLarge is the safety net regarding the count of targets
+// it will check against the number of targets being targeted and the number of targets in the k8s system
+// > 66% of the k8s system being targeted warrants a safety check if we assume each of our targets are replicated
+// at least twice. > 80% in a namespace also warrants a safety check as typically namespaces are shared between services.
```
ptnapoleon, line 188:
```suggestion
// at least twice. > 80% in a namespace also warrants a safety check as typically namespaces are shared between services.
```
> 80% in a namespace also warrants a safety check as typically namespaces are shared between services.

Is this claim true?

safemode/safemode_dns.go
```diff
@@ -0,0 +1,39 @@
...
+// Copyright 2021 Datadog, Inc.
+
+package safemode
+
+import (
+	"github.com/DataDog/chaos-controller/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type DNS struct {
```
ptnapoleon, line 13: Why add several no-ops safemodes? Do we need to have them?

safemode/safemode_network.go
```diff
@@ -0,0 +1,60 @@
...
+// Init Refer to safemode.Safemode interface for documentation
+func (sm *Network) Init(disruption v1beta1.Disruption, client client.Client) {
+	sm.GetTypeSpec(disruption)
+	sm.GetKubeClient(client)
+}
+
+// safetyNetNoHostNoPort is the safety net regarding missing host or port values.
+// it will check against all defined hosts in the network disruption spec to see if any of them have a host or
+// port missing. The more generic a hosts tuple is (Omitting fields such as port), the bigger the blast radius.
+func (sm *Network) safetyNetNoHostNoPort() bool {
```
ptnapoleon, line 48: I think using only port happens fairly often. Could this be implemented as neitherHostNorPort, rather than NotHostAndPort?

safemode/safemode.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/DataDog/chaos-controller/api/v1beta1"
+	chaostypes "github.com/DataDog/chaos-controller/types"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Safemode interface {
+	// CreationSafetyNets will look at all the specific safety nets corresponding to the creation of a new disruption
+	CreationSafetyNets() ([]string, error)
```
ptnapoleon, line 23: This is currently the only place where we are checking and enforcing safety for now, yes? Then I assume the name is to differentiate it from as-yet-unimplemented RuntimeSafetyNets that are checked in an ongoing fashion? 🚲  🏠 , but I'd prefer something like `CheckSpecSafety`, as this method appears to be a verb doing the checking, rather than creating safety nets that will later be checked

safemode/safemode.go
```diff
@@ -0,0 +1,255 @@
...
+	// GetTypeSpec will grab the necessary spec depending on the type (e.g. SafemodeNetwork will grab NetworkDisruptionSpec)
+	// and grab the disruption itself for data such as the kubernetes namespace the disruption is running on
+	GetTypeSpec(disruption v1beta1.Disruption)
+	// GetKubeClient grab the kube client for functions that require state information from k8s system
+	GetKubeClient(client client.Client)
```
ptnapoleon, line 28: These two functions appear to be setters, not getters? I think you're fine just putting `Init` in the interface in that case

controllers/disruption_controller.go
```diff
@@ -150,6 +152,35 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			if len(r.SafetyNets) == 0 {
+				r.SafetyNets = []safemode.Safemode{}
+				r.SafetyNets = safemode.AddAllSafemodeObjects(*instance, r.Client)
+			}
+
+			responses := []string{}
+
+			for _, safetyNet := range r.SafetyNets {
+				// safety nets may occur throughout the reconciler, the safety nets used at creation are done here
+				response, err := safetyNet.CreationSafetyNets()
```
ptnapoleon, line 167: I'd argue either these need to be moved into validate, or they should be checked more often than just the first Reconcile loop

docs/safemode.md
Azoam: IgnoreAll would be set to false in prod. In our global configurations, we would have ignoreAll set automatically depending on environment, so in prod it would be false. So we recommend it to be False in prod.

api/v1beta1/safemode.go
Azoam: I think its easier from a developers POV to keep it as `safemode`. I can see how it could be confusing from a user's POV. The idea is that everything related to safemode would be in that spec, so if a user wants to turn of a safety net, they would go to the safemode spec where it is currently turned on. If you think it still wouldnt make sense this way, maybe we can reorganize the SafemodeSpec's parameters so it makes more sense?

docs/safemode.md
Azoam: I think there are two options here; We could have disabled by default but it is unlikely any one will ever use it, the safety nets are all enabled by default because the idea is that they are unforeseen events. And if they are seen, they can be ignored. We can also just remove it if it brings no real value. I'd rather the later as I dont like the idea of having specific safety nets disabled by default, they should all be enabled if `IgnoreAll` is false, and then users can start disabling once they realize which ones are not needed.

docs/safemode.md
Azoam: It is a violation, as there is no  `host` parameter in the single item in the `hosts` list.

safemode/safemode.go
Azoam: Probably not typical, but I have seen services sharing a namespace.

safemode/safemode_dns.go
Azoam: For now, we do not, but Im guessing for the future when more safety nets are considered for each disruption, they would be implemented in these places. Also I wanted to reference them in `safemode.go` but in order for me to reference them they must exist. Its essentially boilerplate for future safety nets, what are your thoughts?

safemode/safemode_network.go
Azoam: Yes, I think that is a fair change.

docs/safemode.md
ptnapoleon: https://github.com/DataDog/chaos-controller/pull/475#discussion_r785069424

Sorry, with the context of ^, this should be changed too

docs/safemode.md
ptnapoleon: > maintainers of the controller can turn it on by default in production

> IgnoreAll would be set to false in prod

So the comment is backwards?

api/v1beta1/safemode.go
ptnapoleon: developer == us, here yes? we can call the structs and feature whatever we want, but I think as far as the yaml spec is concerned, we should try to make it very clear to users that they're being unsafe

docs/safemode.md
Azoam: Changing this format in the next commit, so resolving this conversation in favor for the new `Unsafemode` schema.

api/v1beta1/safemode.go
Azoam: Noted, new commit will make clear that the spec contain unsafe measures rather than safe measures. Resolving for next commit.

safemode/safemode.go
Azoam: This is a fair comments, as the method doesnt create safety nets to be used later. Though the checks that are being done in `CreationSafetyNets` are ONLY checked during the start of the disruption rather than during the runtime of the disruption. Do you think `CheckInitialSafety` with its counterpart `CheckRuntimeSafety` would be more appropriate?

safemode/safemode.go
Azoam: Good Point, adding their functionality to init and removing the setters.

controllers/disruption_controller.go
Azoam: I see, this goes back to `CheckSpecSafety` as a method that checks continuously rather than just on creation. There are definitely some of the safety nets that would be futile to check more than once (`SpecificContainDisk`), but I can understand wanting to continuously check a safety net such as `CountTooLarge` because if a user has a percentage our check may change as the disruption continues. Do you think its ok for these safety nets to run more than once if they are only required once to run?

ptnapoleon: It depends on how long they take, I think? If it's under 1-2s, I think we can just run them every loop. I'd prefer to move them into validate if they literally only make sense once at a time (e.g. specificcontaindisk)

Azoam: There are some that take a fairly long time, for example the countTooLarge makes a kube api call for many targets. Maybe this is something we can discuss as a team in regards to which one of these should move to validate.

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+    hosts:
+      drop: 100
+```
+
+
+
+
+
+
+
```
ptnapoleon, line 55: err_too_many_new_lines

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+| Namespace-wide Targeting                                  | Generic | Using generical label selectors (e.x. X) that selects a majority of pods/nodes in a namespace as a target to inject a disruption into                     | DisableCountTooLarge       |
+| Sporadic Targets                                          | Generic | In a volatile environment where targets are being terminated and created sporadically, disruptions should not be allowed to continue disrupting           | DisableSporadicTargets     |
+| No Port and No Host Specified                             | Network | Running a network disruption without specifying a port and a host                                                                                         | DisableNeitherHostNorPort  |
+| Specific Container Disk Disruption on Multi Container Pod | Disk    | The disk disruption is a pod-wide disruption, if a user tries to specify a specific container, they may be unware they are affecting all other containers | DisableSpecificContainDisk |
```
ptnapoleon, line 24:
```suggestion
| Specific Container Disk Disruption on Multi Container Pod | Disk    | The disk disruption is a pod-wide disruption, if a user tries to specify a specific container, they may be unaware they are affecting all other containers | DisableSpecificContainDisk |
```

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+Therefore the controller allows for you to disable specific safety nets in the Safemode Spec. Checkout out example below to see how to remove certain safety nets.
+Keep in mind that all safety nets are turned on by default.
+
+### Safety Nets
+
+| Safety Net                                                | Type | Description                                                                                                                                               | IgnoreName                 |
+|-----------------------------------------------------------| ----------- |-----------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------|
+| Namespace-wide Targeting                                  | Generic | Using generical label selectors (e.x. X) that selects a majority of pods/nodes in a namespace as a target to inject a disruption into                     | DisableCountTooLarge       |
```
ptnapoleon, line 21:
```suggestion
| Namespace-wide Targeting                                  | Generic | Using generic label selectors that select a majority of pods/nodes in a namespace as a target to inject a disruption into                     | DisableCountTooLarge       |
```

Could we specify the exact percentage, rather than `majority`? (unless it is literally 50+%, then this is fine)

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Ignoring Safety Nets
+
+Because the list of safety nets to be implemented will grow in the future, there will surely be overlap with safety nets which will make it difficult for a user who is confident a specific safety net is not necessary but unsure if others will be.
+Therefore the controller allows for you to disable specific safety nets in the Safemode Spec. Checkout out example below to see how to remove certain safety nets.
+Keep in mind that all safety nets are turned on by default.
+
+### Safety Nets
+
+| Safety Net                                                | Type | Description                                                                                                                                               | IgnoreName                 |
```
ptnapoleon, line 19: Could the language under `Description` be a bit more consistent for each?

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
+# Safemode
+
+Safemode represents a number of safety nets we have implemented into the chaos controller to help new and experienced users feel more confident deploying new disruptions to their environments.
+
+The Chaos Controller can be scary to use in production environments, but we learn a lot more running chaos experiments in our production environments. Considering this, having safety nets in place makes the entire process of running chaos experiments in high value environments a little safer.
```
ptnapoleon, line 5:
```suggestion
Chaos engineering can be scary to use in production environments, but we have learned a lot after using the chaos-controller for years. We've attempted to coalesce these lessons into safety nets that prevent common dangerous options.Having safety nets in place makes the entire process of running chaos experiments in high value environments a little safer.
```

What do you think?

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+
+Safemoe will be added true by default and will require manual disabling of safety nets in order to bypass. In the disruption schema we have `unsafeMode` which represents ways to remove the safety nets.
```
ptnapoleon, line 7:
```suggestion
Safemode is always enabled by default and will require manual disabling of safety nets in order to bypass. In the disruption schema we have `unsafeMode` which represents ways to remove the safety nets.
```

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+`disableAll` is self-explainatory while the other disables in the unsafe schema represent each safety net which you may disable one by one.
```
ptnapoleon, line 8:
```suggestion
`unsafeMode.disableAll` turns off all safety nets. The other options under `unsafeMode` represent individual safety nets which can be disabled independently.
```

docs/safemode.md
```diff
@@ -0,0 +1,55 @@
...
+Please take a look at the example below to see how to use the `unsafeMode` parameter.
```
ptnapoleon, line 9:
```suggestion
Please take a look at the example below to see how to use `unsafeMode`.
```

examples/disk_pressure_read_safemode.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: Disruption
+metadata:
+  name: disk-pressure-read
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 1
+  safeMode:
```
ptnapoleon, line 16:
```suggestion
  unsafeMode: 
```

examples/disk_pressure_read_safemode.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: disk-pressure-read
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 1
+  safeMode: 
+    ignoreAll: false
```
ptnapoleon, line 17:
```suggestion
    disableAll: false
```

Why disable All here? Dont we just need to disable the disk check?

ptnapoleon: Do we even need to disable the disk check?

docs/safemode.md
Azoam: Currently Ive added arbitrary values for the safety check:
80% at the Namespace level
66% at the Cluster level

Not sure what more appropriate values would be, have any ideas?

docs/safemode.md
Azoam: I will try and be more consistent in the language in the next commitment, please let me know if it is still not consistent.

docs/safemode.md
Azoam: *coalesce*

docs/safemode.md
ptnapoleon: The values are fine, I just feel we should tell the users the exact numbers we're checking for, here in the docs

examples/disk_pressure_read_safemode.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: disk-pressure-read
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 1
+  unsafeMode:
+    disableAll: false
```
ptnapoleon, line 17: Do we need `DisableSpecificContainDisk` here?

docs/safemode.md
```diff
@@ -0,0 +1,48 @@
+# Safemode
+
+Safemode represents a number of safety nets we have implemented into the chaos controller to help new and experienced users feel more confident deploying new disruptions to their environments.
+
+Chaos engineering can be scary to use in production environments, but we have learned a lot after using the chaos-controller for years. We've attempted to coalesce these lessons into safety nets that prevent common dangerous options.Having safety nets in place makes the entire process of running chaos experiments in high value environments a little safer.
```
ptnapoleon, line 5:
```suggestion
Chaos engineering can be scary to use in production environments, but we have learned a lot after using the chaos-controller for years. We've attempted to coalesce these lessons into safety nets that prevent common dangerous options. Having safety nets in place makes the entire process of running chaos experiments in high value environments a little safer.
```

examples/disk_pressure_read_safemode.yaml
Azoam: Yes, nice catch.

Azoam: yes, good catch

safemode/safemode_network.go
```diff
@@ -0,0 +1,55 @@
...
+	dis    v1beta1.Disruption
+	client client.Client
+}
+
+// CreationSafetyNets Refer to safemode.Safemode interface for documentation
+func (sm *Network) CreationSafetyNets() ([]string, error) {
+	safetyNetResponses := []string{}
+	// run through the list of initial safety nets
+	if caught := sm.safetyNetNeitherHostNorPort(); caught {
+		safetyNetResponses = append(safetyNetResponses, " The specified disruption either contains no Hosts or contains a Host which only has either a port or a host. The more ambiguous, the larger the blast radius. ")
```
ptnapoleon, line 23:
```suggestion
		safetyNetResponses = append(safetyNetResponses, " The specified disruption either contains no Hosts or contains a Host which has neither a port or a host. The more ambiguous, the larger the blast radius. ")
```

safemode/safemode_disk.go
```diff
@@ -0,0 +1,79 @@
...
+// safetyNetSpecificContainDisk is the safety net regarding running a disk disruption (which hits all containers) while a user targets specific containers.
+// This is a misunderstanding of the disk disruption as disk disruption disrupt all containers and a user asking to disrupt specific containers will have unforeseen consequences.
+func (sm *Disk) safetyNetSpecificContainDisk() (bool, error) {
+	if len(sm.dis.Spec.Containers) == 0 {
+		// No specified containers in the disruption, safety net is avoided in this case
+		return false, nil
+	}
+
+	if sm.dis.Spec.Level != chaostypes.DisruptionLevelPod {
+		// Node level disruption should be clear because choosing specific containers has a null affect
```
ptnapoleon, line 52:
```suggestion
		// Node level disruption should be clear because choosing specific containers has a null effect
```

controllers/disruption_controller.go
ptnapoleon: CountTooLarge and SporadicTargets are way too time consuming to move into the reconcile loop, imo

safemode/safemode.go
```diff
@@ -0,0 +1,242 @@
...
+	return safetyNetResponses, nil
+}
+
+// safetyNetSporadicTargets is the safety net regarding sporadic targets
+// in an environment which count is constantly changing, we are looking at an environment where targets are being
+// destroyed and created frequently which is a very bad sign in terms of reliability and a applying further
+// disruptions is probably not the best idea.
+// In this function we run a check against the count of the target environment 3 times. If the count is different each of those
+// 3 times, we assume sporadic behaviour of the target environment and raise a flag.
+func (sm *Generic) safetyNetSporadicTargets() (bool, error) {
```
Devatoria, line 122: Are those 3 list calls done sequentially (with no pause) and on disruption creation? Because if it is the case, the time to pull the injector image and to effectively inject the disruption, we may list targets before the injection is really done. In addition to this, if we take the example of a pod with a liveness probe, the pod may start failing multiple seconds after the disruption is injected and you will likely miss that.

I'd also not only rely on the targets count but maybe on the targets status changes? A target going from ready to not ready during a disruption (as well as a target being replaced as you check here) is a good indicator of something going wrong. Though I feel those things should be checked during the whole disruption for effects that may take a while to appear. And in this case, relying on events may be a better solution so you can follow changes over time (like we do for dynamic target injection in Nathan's PR).

safemode/safemode.go
```diff
@@ -0,0 +1,242 @@
...
+
+	return true, nil
+}
+
+// safetyNetCountNotTooLarge is the safety net regarding the count of targets
+// it will check against the number of targets being targeted and the number of targets in the k8s system
+// > 66% of the k8s system being targeted warrants a safety check if we assume each of our targets are replicated
+// at least twice. > 80% in a namespace also warrants a safety check as namespaces may be shared between services.
+// returning true indicates the safety net caught something
+func (sm *Generic) safetyNetCountNotTooLarge() (bool, error) {
```
Devatoria, line 177: Those values are great but a bit arbitrary, not everyone may want those values. Could we configure them somewhere maybe?

controllers/disruption_controller.go
Devatoria: I also feel that most if not all (maybe just the sporadic one but I'd still have it working in a goroutine that would itself trigger the reconcile loop if caught instead) of those safety nets could be done once during validate so we do not overload the reconcile loop. We must keep in mind that the reconcile loop is not always called and that when it is called, it blocks the controller which won't reconcile other disruptions, so better make it fast. One last argument for checking that in validate is that you can immediately reject the disruption instead of going through events.

safemode/safemode.go
Devatoria: The more I think about this safety net, the more I realize that it should not be a satefy net. The idea of such a mechanism is very good to be honest (I'd love to see the controller telling to someone that the disruption they created somehow has a negative effect on a target) but I also feel that the whole mechanism should not be a safety net (which should be done before injection to me) but more a watchdog system (observing the disruption during its whole lifecycle and emitting events when something is seen).

What do you think? Should we remove it from this PR and think about a watchdog mechanism for disruptions? It would prevent to affect the reconcile loop at all because it would be totally separate.

Azoam: Im ok with this. To be honest I dont think we've ever really had an issue with a team having issues with a disruption during a sporadic environment like this, so this may have been an attempt to create a safety net for something we dont really have issues for, lots of assuming.

Its a good idea to have this in some sort of watchdog mechanism, it is great information to have as running disruptions in these environments maybe a weary operation.

I will remove Sporadic Targets in the next commit.

examples/disk_pressure_read_unsafemode.yaml
```diff
@@ -0,0 +1,25 @@
...
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 1
+  unsafeMode:
+    disableAll: false
+    config:
+      countTooLarge:
```
ptnapoleon, line 19: Is the purpose of this example to showcase `config.countTooLarge`? Then I think we should either drop the selector entirely or use a very broad one, e.g., `os: linux`, `cloud: aws`, etc., as well as rename the file. Wdyt?

controllers/disruption_controller.go
```diff
@@ -171,6 +173,39 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				// it is possible for a disruption to be restarted with new parameters, therefore safety nets need to be reinitialized to catch that case
+				// so that we are not using values from older versions of a disruption for safety nets
+				safemode.Reinit(r.SafetyNets, *instance, r.Client)
+			}
+
+			responses := []string{}
+
+			for _, safetyNet := range r.SafetyNets {
+				// safety nets may occur throughout the reconciler, the safety nets used at creation are done here
+				response, err := safetyNet.CheckInitialSafetyNets()
```
ptnapoleon, line 192: Is this still happening at every reconcile loop?

api/v1beta1/disruption_webhook.go
```diff
@@ -87,6 +94,18 @@ func (r *Disruption) ValidateCreate() error {
 		return err
 	}
 
+	// handle initial safety nets
+	if responses, err := r.initialSafetynets(); err != nil {
+		return err
+	} else if len(responses) > 0 {
+		retErr := errors.New("safety net catches")
```
ptnapoleon, line 101: I don't think this line adds much; the next line is sufficient

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
 
 	return tags
 }
+
+// initialSafetynets runs the initial safety nets for any new disruption
+// returns true if any safety net were caught and returns any errors when attempting to run the safety nets
+func (r *Disruption) initialSafetynets() ([]string, error) {
```
ptnapoleon, line 195:
```suggestion
func (r *Disruption) initialSafetyNets() ([]string, error) {
```

For consistency with other camelCasing of safetyNet

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
 
 	return tags
 }
+
+// initialSafetynets runs the initial safety nets for any new disruption
+// returns true if any safety net were caught and returns any errors when attempting to run the safety nets
```
ptnapoleon, line 194: This doesn't return a boolean, this comment is out of date

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
...
+func (r *Disruption) initialSafetynets() ([]string, error) {
+	responses := []string{}
+	// handle initial safety nets if safemode is enabled
+	if r.Spec.Unsafemode == nil || !r.Spec.Unsafemode.DisableAll {
+		if caught, err := safetyNetCountNotTooLarge(*r); err != nil {
+			return nil, fmt.Errorf("error checking for countNotTooLarge safetynet: %w", err)
+		} else if caught {
+			logger.Debugw("the specified count represents a large percentage of targets in either the namespace or the kubernetes cluster", r.Name, "SafetyNet Catch", "Generic")
+
+			responses = append(responses, "the specified count represents a large percentage of targets in either the namespace or the kubernetes cluster")
```
ptnapoleon, line 204: Could `safetyNetCountNotTooLarge` return more info than just a boolean, so we could let the user know exactly how many targets were selected compared to the safe limit?

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
...
+	}
+
+	userCount := r.Spec.Count
+	totalCount := 0
+	namespaceCount := 0
+	namespaceThreshold := 0.8
+	clusterThreshold := 0.66
+
+	if r.Spec.Unsafemode.Config != nil && r.Spec.Unsafemode.Config.CountTooLarge != nil {
+		if r.Spec.Unsafemode.Config.CountTooLarge.NamespaceThreshold != 0 {
```
ptnapoleon, line 236: It seems there is no way to disable the namespace threshold without disabling the cluster threshold. Are we okay with that? I might be fine with a disruption that targets 100% of a namespace, but not if it's also targeting 80% of the cluster.

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
...
+
+		if namespaceCount != 0 {
+			userCountVal = float64(userCountInt) / 100.0 * float64(namespaceCount)
+		} else {
+			userCountVal = float64(userCountInt) / 100.0 * float64(totalCount)
+		}
+	}
+
+	// we check to see if the count represents > 80 percent of all pods in the existing namepsace
+	// and if the count represents > 66 percent of all pods in the cluster (2/3s)
```
ptnapoleon, line 291:
```suggestion
	// and if the count represents > 66 percent of all pods in the cluster
```

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
...
+			return false, fmt.Errorf("error listing target pods: %w", err)
+		}
+
+		totalCount = len(nodes.Items)
+	}
+
+	userCountVal := float64(userCount.IntVal)
+
+	if userCount.Type != intstr.Int {
+		userCountInt, err := strconv.Atoi(strings.TrimSuffix(userCount.StrVal, "%"))
```
ptnapoleon, line 278: Can we call `GetIntOrPercentValueSafely` here?

api/v1beta1/disruption_webhook.go
```diff
@@ -170,3 +189,135 @@ func (r *Disruption) getMetricsTags() []string {
...
+		if err != nil {
+			return false, fmt.Errorf("failed to convert percentage to int: %w", err)
+		}
+
+		if namespaceCount != 0 {
+			userCountVal = float64(userCountInt) / 100.0 * float64(namespaceCount)
```
ptnapoleon, line 284: It seems to me that if we are getting the count as a %, then there's no need to do the k8s list api calls at all? We can just directly compare the target % to the safe %?

ptnapoleon: It also seems we aren't applying the selectors here, so these `userCountVal`s aren't necessarily accurate?

api/v1beta1/disruption_webhook.go
Azoam: if the user sets up the namespace targeting threshold at 100% and the cluster threshold as 80% they can accomplish what you're asking. Let me know if I am misunderstanding.

ptnapoleon: Oh, great point, I missed that we were doing a `>` check instead of a `>=` check.

api/v1beta1/disruption_webhook.go
Azoam: Comparing those two percentages wouldnt accomplish the goal I believe. Target % represents the percentage of targets out of the total number of targets while the Safe % represents the percentage of targets out of either the total number of pods in a namespace or the total number of pods in a cluster. 

 So 20% of pods labeled "app:chaos" could be much different than 20% of the namespace that the targets labeled "app:chaos" live in and of course it would be much different than 20% of the total number of pods on the cluster. So we still need to use the API to see how many total pods are in the namespace. I know there are services that have only the same targets in a namespace, but others which have several services in a single namespace. Let me know if I misunderstood.

Also to note the second comment, yes it would be inaccurate, good catch. I have added another API call to check the number of targets in the cluster that match, that value of userCountVal should be accurate now to the number of targets.

api/v1beta1/disruption_webhook.go
```diff
@@ -211,3 +225,148 @@ func (r *Disruption) getMetricsTags() []string {
 
 	return tags
 }
+
+// initialSafetyNets runs the initial safety nets for any new disruption
+// returns a list of responses related to safety net catches if any safety net were caught and returns any errors when attempting to run the safety nets
+func (r *Disruption) initialSafetyNets() ([]string, error) {
+	responses := []string{}
+	// handle initial safety nets if safemode is enabled
+	if r.Spec.Unsafemode == nil || !r.Spec.Unsafemode.DisableAll {
```
ptnapoleon, line 234: I'm not exactly sure which booleans need to change, but currently, when `unsafeMode` is nil, we will run both safety checks. We will then get a panic when we check `unsafeMode.Config` at line 272

ptnapoleon (changes requested):

api/v1beta1/disruption_webhook.go
```diff
@@ -211,3 +225,148 @@ func (r *Disruption) getMetricsTags() []string {
...
+
+	if isPercent {
+		userCountVal = float64(userCountInt) / 100.0 * float64(targetCount)
+	} else {
+		userCountVal = float64(userCountInt)
+	}
+
+	// we check to see if the count represents > 80 percent of all pods in the existing namepsace
+	// or if the count represents > 66 percent of all pods in the cluster
+	if userNamespacePercent := userCountVal / float64(namespaceCount); userNamespacePercent > namespaceThreshold {
```
ptnapoleon, line 339: We need to skip this check when `level == node`. Otherwise, we end up dividing by zero, and get a validation error like:
```
	* target selection represents +Inf % of the total pods in the namespace while the threshold is 80.00 %
```

api/v1beta1/disruption_webhook.go
Azoam: Nice catch, I've added a check on line 272, if you think there should be something in Unsafemode itself let me know.

api/v1beta1/disruption_webhook.go
```diff
@@ -269,13 +269,15 @@ func safetyNetCountNotTooLarge(r Disruption) (bool, string, error) {
 	namespaceThreshold := 0.8
 	clusterThreshold := 0.66
 
-	if r.Spec.Unsafemode.Config != nil && r.Spec.Unsafemode.Config.CountTooLarge != nil {
-		if r.Spec.Unsafemode.Config.CountTooLarge.NamespaceThreshold != 0 {
-			namespaceThreshold = float64(r.Spec.Unsafemode.Config.CountTooLarge.NamespaceThreshold) / 100.0
-		}
+	if r.Spec.Unsafemode == nil {
```
ptnapoleon, line 272:
```suggestion
	if r.Spec.Unsafemode != nil {
```

api/v1beta1/disruption_webhook.go
```diff
@@ -336,9 +338,11 @@ func safetyNetCountNotTooLarge(r Disruption) (bool, string, error) {
 
 	// we check to see if the count represents > 80 percent of all pods in the existing namepsace
 	// or if the count represents > 66 percent of all pods in the cluster
-	if userNamespacePercent := userCountVal / float64(namespaceCount); userNamespacePercent > namespaceThreshold {
-		response := fmt.Sprintf("target selection represents %.2f %% of the total pods in the namespace while the threshold is %.2f %%", userNamespacePercent*100, namespaceThreshold*100)
-		return true, response, nil
+	if r.Spec.Level != chaostypes.DisruptionLevelNode {
+		if userNamespacePercent := userCountVal / float64(namespaceCount); userNamespacePercent > namespaceThreshold {
+			response := fmt.Sprintf("target selection represents %.2f %% of the total pods in the namespace while the threshold is %.2f %%", userNamespacePercent*100, namespaceThreshold*100)
```
ptnapoleon, line 343:
```suggestion
			response := fmt.Sprintf("target selection represents %.2f %% of the total %ss in the namespace while the threshold is %.2f %%", userNamespacePercent*100, r.Spec.Level, namespaceThreshold*100)
```

api/v1beta1/disruption_webhook.go
```diff
@@ -211,3 +225,152 @@ func (r *Disruption) getMetricsTags() []string {
...
+	// or if the count represents > 66 percent of all pods in the cluster
+	if r.Spec.Level != chaostypes.DisruptionLevelNode {
+		if userNamespacePercent := userCountVal / float64(namespaceCount); userNamespacePercent > namespaceThreshold {
+			response := fmt.Sprintf("target selection represents %.2f %% of the total pods in the namespace while the threshold is %.2f %%", userNamespacePercent*100, namespaceThreshold*100)
+			return true, response, nil
+		}
+	}
+
+	if userTotalPercent := userCountVal / float64(totalCount); userTotalPercent > clusterThreshold {
+		response := fmt.Sprintf("target selection represents %.2f %% of the total pods in the cluster while the threshold is %.2f %%", userTotalPercent*100, clusterThreshold*100)
```
ptnapoleon, line 349:
```suggestion
		response := fmt.Sprintf("target selection represents %.2f %% of the total %ss in the cluster while the threshold is %.2f %%", userTotalPercent*100, r.Spec.Level, clusterThreshold*100)
```

PR #483, opened 2022-01-25
Don't attempt re-injection after Disruption Duration has ended

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Previously, after reaching the `Injected` status, we would set the requeue delay for long enough that we dont reconcile again until it's time to clean up the now finished chaos pods. This naively assumes that we will ever reach the Injected status, which can't be guaranteed. If that doesn't happen, we would actually try to continue to inject new pods after the duration has expired, but before we GC the disruption. Given that we failed to reach Injected the first time, it's likely that these new pods will fail too. Thus, we now stop injecting and immediately clean the disruption after the duration is over

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -114,8 +114,8 @@ func (dd DisruptionDuration) Duration() time.Duration {
 type DisruptionStatus struct {
 	IsStuckOnRemoval bool `json:"isStuckOnRemoval,omitempty"`
 	IsInjected       bool `json:"isInjected,omitempty"`
-	// +kubebuilder:validation:Enum=NotInjected;PartiallyInjected;Injected
-	// +ddmark:validation:Enum=NotInjected;PartiallyInjected;Injected
+	// +kubebuilder:validation:Enum=NotInjected;PartiallyInjected;Injected;PreviouslyInjected
+	// +ddmark:validation:Enum=NotInjected;PartiallyInjected;Injected;PreviouslyInjected
 	InjectionStatus chaostypes.DisruptionInjectionStatus `json:"injectionStatus,omitempty"`
 	// +nullable
 	Targets []string `json:"targets,omitempty"`
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -182,6 +182,37 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			}
 
 			return ctrl.Result{Requeue: true}, err
+		} else if calculateRemainingDuration(*instance) <= 0 {
+			if _, err := r.updateInjectionStatus(instance); err != nil {
+				r.log.Errorw("error updating disruption injection status", "error", err)
+
+				return ctrl.Result{}, fmt.Errorf("error updating disruption injection status: %w", err)
+			}
+
+			isCleaned, err := r.cleanDisruption(instance)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+
+			if !isCleaned {
+				requeueAfter := time.Duration(rand.Intn(5)+5) * time.Second //nolint:gosec
+
+				r.log.Infow(fmt.Sprintf("disruption has not been fully cleaned yet, re-queuing in %v", requeueAfter))
+
+				return ctrl.Result{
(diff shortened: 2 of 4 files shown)


Azoam (approved):

chart/values.yaml
```diff
@@ -23,13 +21,12 @@ controller:
       enabled: false
       tokenFilepath: ""
   defaultDuration: 1h # default spec.duration for a disruption with none specified
-  expiredDisruptionGCDelay: 15m # time after a disruption expires before deleting it
+  expiredDisruptionGCDelay: 1m # time after a disruption expires before deleting it
```
ptnapoleon, line 24: oh no wooops, should learn not to do `git add .`

ptnapoleon: Somehow deleted the whole branch fixing this

api/v1beta1/disruption_types.go
```diff
@@ -112,8 +112,8 @@ func (dd DisruptionDuration) Duration() time.Duration {
 type DisruptionStatus struct {
 	IsStuckOnRemoval bool `json:"isStuckOnRemoval,omitempty"`
 	IsInjected       bool `json:"isInjected,omitempty"`
-	// +kubebuilder:validation:Enum=NotInjected;PartiallyInjected;Injected
-	// +ddmark:validation:Enum=NotInjected;PartiallyInjected;Injected
+	// +kubebuilder:validation:Enum=NotInjected;PartiallyInjected;Injected;PreviouslyInjected
```
Devatoria, line 115: Why do we need to change this now? Does it mean we were never updating to this status before in the update status function?
Would it be worth adding a test case [here](https://github.com/DataDog/chaos-controller/blob/4ee14389984f7b421ddf9571955ba5ce7c356060/controllers/disruption_controller_test.go#L256) to be sure that this status is now set?

controllers/disruption_controller.go
```diff
@@ -182,6 +182,37 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
 			return ctrl.Result{Requeue: true}, err
+		} else if calculateRemainingDuration(*instance) <= 0 {
+			if _, err := r.updateInjectionStatus(instance); err != nil {
+				r.log.Errorw("error updating disruption injection status", "error", err)
+
+				return ctrl.Result{}, fmt.Errorf("error updating disruption injection status: %w", err)
+			}
+
+			isCleaned, err := r.cleanDisruption(instance)
```
Devatoria, line 192: I see that we call clean disruption here, how were we handling the cleanup before that (on expiration)?

api/v1beta1/disruption_types.go
ptnapoleon: It means we were never updating to this status beforehand, yep. I will add a test case :thubs:

controllers/disruption_controller.go
ptnapoleon: We were letting the chaos pods clean themselves up (they had a TTL set). This works most of the time (well it works all of the time if the chaos pods injected correctly), but it's helpful to clean explicitly here just to be safe but also to handle chaos pods that got stuck

controllers/disruption_controller_test.go
```diff
@@ -287,6 +287,20 @@ var _ = Describe("Disruption Controller", func() {
...
 
+			By("Waiting for the disruption to reach PreviouslyInjected")
+			Eventually(func() error {
+				if err := k8sClient.Get(context.Background(), instanceKey, disruption); err != nil {
+					return err
+				}
+
+				// check disruption injection status
+				if disruption.Status.InjectionStatus != chaostypes.DisruptionInjectionStatusPreviouslyInjected {
+					return fmt.Errorf("disruptions is not injected, current status is %s", disruption.Status.InjectionStatus)
```
Devatoria, line 298:
```suggestion
					return fmt.Errorf("unexpected disruption status, current status is %s (expected PreviouslyInjected)", disruption.Status.InjectionStatus)
```

Devatoria (approved):
A small comment on the test log to make it easier to understand if we need to debug it + linting suff otherwise LGTM!

PR #485, opened 2022-01-26
Custom cache watcher example

nathantournant:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug

- Ongoing work for the live target retargeting
- Live selector-based cache system per disruption, only drops a log for now
- Lifecycle: for each disruption, the cache is created if it doesn't exist, and is deleted on disruption deletion and/or on controller shutdown
- Next step/ongoing work: enqueue a reconcile loop trigger from the cache function

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -61,7 +61,7 @@ type DisruptionReconciler struct {
 	InjectorDNSDisruptionKubeDNS          string
 	InjectorNetworkDisruptionAllowedHosts []string
 	ExpiredDisruptionGCDelay              time.Duration
-	Caches                                map[types.NamespacedName]k8scache.Cache
+	Caches                                map[types.NamespacedName]*k8scache.Cache
 }
 
 //+kubebuilder:rbac:groups=chaos.datadoghq.com,resources=disruptions,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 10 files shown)


controllers/disruption_controller.go
```diff
@@ -58,6 +60,38 @@ type DisruptionReconciler struct {
 	InjectorDNSDisruptionKubeDNS          string
 	InjectorNetworkDisruptionAllowedHosts []string
 	ExpiredDisruptionGCDelay              time.Duration
+	Caches                                map[types.NamespacedName]context.CancelFunc
+}
+
+type SelectorPodsHandler struct {
+	disruption *v1beta1.Disruption
+}
```
ptnapoleon, line 68: I know the PR is in progress still, but I don't see yet the benefit of this design, why use a struct that contains just a disruption to hold these methods?

nathantournant: This PR is long outdated but it's there for example purposes only;
The goal is to make a handler fitting the handler interface with the 3 notifying methods; since there's going to be a unique handler per selector cache, and one dedicated selector cache per ongoing disruption, this handler will be triggered iif the corresponding disruption's selector cache is triggered
Having the disruption in the struct allows to have all the needed metadata to easily identify which selector was triggered

PR #488, opened 2022-01-28
Change Disruption behavior at end of lifecycle

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Turns two log statements down to DEBUG from INFO. These are logged during normal conditions in normal behavior, and I think they really are only useful while debugging
- The activeDeadlineSeconds on a pod has precision to the nearest second. By casting the remaining disruption duration to an int, we were rounding it down. This meant that chaos pods would always terminate under a second prior to the disruption, which would trigger the informer, which would lead to the user getting notified that the disruption was either Empty or had NoTargets. This is not ideal
- We now round the chaos pod duration up. However, this was leading to a situation where the disruption would delete the chaos pod while it was cleaning. On k8s 1.19, this was sometimes leading to a very weird situation where the chaos pod would clean up but never get deleted or GC'd. For now, we wait until the disruption is being deleted before we delete the chaos pods, but add an Event so users understand the lifecycle a bit better

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -23,7 +23,7 @@ controller:
       enabled: false
       tokenFilepath: ""
   defaultDuration: 1h # default spec.duration for a disruption with none specified
-  expiredDisruptionGCDelay: 15m # time after a disruption expires before deleting it
+  expiredDisruptionGCDelay: 10m # time after a disruption expires before deleting it
   webhook: # admission webhook configuration
     generateCert: false # if you want Helm to generate certificates (e.g. in case the cert-manager is not installed in the cluster) set this to true
     certDir: "" # certificate directory (must contain tls.crt and tls.key files)
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -189,6 +189,19 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				return ctrl.Result{}, fmt.Errorf("error updating disruption injection status: %w", err)
 			}
 
+			r.Recorder.Event(instance, "Normal", "DurationOver", fmt.Sprintf("The disruption has lived longer than its specified duration, and will be garbage collected after %s.", r.ExpiredDisruptionGCDelay))
+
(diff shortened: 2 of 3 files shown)


Azoam (approved):

controllers/disruption_controller.go
```diff
@@ -189,20 +189,17 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-			if !isCleaned {
-				requeueAfter := time.Duration(rand.Intn(5)+5) * time.Second //nolint:gosec
-
-				r.log.Infow(fmt.Sprintf("disruption has not been fully cleaned yet, re-queuing in %v", requeueAfter))
-
-				return ctrl.Result{
-					Requeue:      true,
-					RequeueAfter: requeueAfter,
-				}, r.Update(context.Background(), instance)
+			for _, pod := range chaosPods {
```
Devatoria, line 201: Aren't we gonna flood someone having a disruption with a lot of chaos pods here?

ptnapoleon: We aren't notifying for every event on a chaos pod, are we? These should only show up if you describe the chaos pods

chart/values.yaml
```diff
@@ -23,7 +23,7 @@ controller:
       enabled: false
       tokenFilepath: ""
   defaultDuration: 1h # default spec.duration for a disruption with none specified
-  expiredDisruptionGCDelay: 15m # time after a disruption expires before deleting it
+  expiredDisruptionGCDelay: 10m # time after a disruption expires before deleting it
```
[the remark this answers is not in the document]
ptnapoleon: I chose to do so, thinking it would be safe to drop the default, just for users to see faster GCs

controllers/disruption_controller.go
Devatoria: Only on warning events, nvm

Devatoria (approved):

PR #492, opened 2022-02-01
Extract namespace into a Helm property.

nikos912000:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

Right now the namespace is either hardcoded or set as an [injector property for the Service Account](https://github.com/DataDog/chaos-controller/blob/1fb906c7e5cc6149a329bff4ba753f8465563702/chart/values.yaml#L36).

It would be good to consolidate and set this in a single place.
This will also allow users to install the resources in a different namespace - we have to do that for certain reasons internally.

## Code Quality Checklist

## Testing

### DIFF
--- a/chart/templates/service_account.yaml
+++ b/chart/templates/service_account.yaml
@@ -6,10 +6,10 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: chaos-controller
-  namespace: chaos-engineering
+  namespace: {{ .Values.global.namespace }}
 ---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ .Values.injector.serviceAccount }}
-  namespace: {{ .Values.injector.chaosNamespace }}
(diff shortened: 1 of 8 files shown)


chart/templates/service_account.yaml
```diff
@@ -6,10 +6,10 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: chaos-controller
-  namespace: chaos-engineering
+  namespace: {{ .Values.global.namespace }}
```
ptnapoleon, line 9: Is this still going to default to chaos-engineering for us?

ptnapoleon (approved):

chart/templates/service_account.yaml
nikos912000: oops, thanks for spotting. Updated.
The challenge of keeping internal charts in sync with the Open Source. With these changes we'll be fully in sync :)

PR #499, opened 2022-02-07
Make expiredDisruptionGCDelay optional

### LINKED ISSUE #497: User Request: Feature flag to disable deletion of Disruption
**Is your feature request related to a problem? Please describe.**
Our team is working on integrating the chaos controller with our CI/CD platform.

The setup we have is as follows:
- [Kubefed](https://github.com/kubernetes-sigs/kubefed) acts as the control plane and propagates any changes to worker clusters.
- Worker clusters are the ones where workloads get deployed. These are the clusters where we run Disruptions.
- The Spinnaker pipeline deploys Disruptions to the worker clusters through Kubefed.

Right now the deletion of Disruptions is done by the controller. This is happening once the Disruption expires and after the GC period.
This is great, but with a control plane we tend to have kubefed handling the lifecycle of any resources deployed to worker clusters. Ideally the responsibility to delete Disruptions would be on Spinnaker and Kubefed.

**Describe the solution you'd like**
Introduce a toggle in the controller's ConfigMap which disables the deletion.

**Describe alternatives you've considered**
The other option for us is to set the GC to a very large value so Spinnaker/Kubefed are always the first ones to do the deletion.

**Additional context**
Let us know if you have any other ideas. If this can be solved in a different way happy to consider it!

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Closes #497 
- When `expiredDisruptionGCDelay` is not set, we skip this behavior entirely. We still allow the valid value of `0`

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -57,7 +57,7 @@ type DisruptionReconciler struct {
 	InjectorDNSDisruptionDNSServer        string
 	InjectorDNSDisruptionKubeDNS          string
 	InjectorNetworkDisruptionAllowedHosts []string
-	ExpiredDisruptionGCDelay              time.Duration
+	ExpiredDisruptionGCDelay              *time.Duration
 }
 
 //+kubebuilder:rbac:groups=chaos.datadoghq.com,resources=disruptions,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -133,7 +133,7 @@ func main() {
 	pflag.StringVar(&cfg.Controller.ImagePullSecrets, "image-pull-secrets", "", "Secrets used for pulling the Docker image from a private registry")
 	handleFatalError(viper.BindPFlag("controller.imagePullSecrets", pflag.Lookup("image-pull-secrets")))
 
-	pflag.DurationVar(&cfg.Controller.ExpiredDisruptionGCDelay, "expired-disruption-gc-delay", time.Minute*15, "Seconds after a disruption expires before being automatically deleted")
+	pflag.DurationVar(&cfg.Controller.ExpiredDisruptionGCDelay, "expired-disruption-gc-delay", time.Minute*(-1), "Duration after a disruption expires before being automatically deleted, leave unset to disable")
```
ptnapoleon, line 136: @nikos912000 Will this work for you? Locally, I just remove this field from the configmap, and it correctly gives me the default I want

controllers/disruption_controller.go
```diff
@@ -189,27 +189,19 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-			chaosPods, err := r.getChaosPods(instance, nil)
-			if err != nil {
-				r.log.Errorw("couldn't list instance's chaos pods", "err", err)
-
-				return ctrl.Result{}, err
-			}
+				r.log.Infow("requeuing disruption to check for its expiration", "requeueDelay", requeueDelay.String())
 
-			for _, pod := range chaosPods {
-				r.Recorder.Event(&pod, "Normal", "DurationOver", fmt.Sprintf("The chaos pod has lived longer than the disruption duration, and will be garbage collected after %s.", r.ExpiredDisruptionGCDelay))
```
ptnapoleon, line 202: I think this is unnecessary now, because we see the pods move into Completed

main.go
nikos912000: I think so :)

Devatoria (approved):

PR #506, opened 2022-02-11
[CHAOS-209] Dynamic service resolution

clairecng:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

As of today, when injecting a network disruption filtering on a service, the service endpoints are resolved once on injection and never updated. It means that if some endpoints are removed or added to the targeted service while the disruption is injected, they won’t be removed from or added to tc filters. For long running disruptions, for instance during gamedays, it can change the outcomes because there’s a good chance that some endpoints being the targeted service are added or recreated.
In this PR, services are resolved directly in the inject logic.

## Code Quality Checklist

## Testing

    - Modification / Deletion of services
    - Modification of pods replicas (scaling up and down)
    - [x] locally.


injector/network_disruption.go
```diff
@@ -374,78 +416,442 @@ func (i *networkDisruptionInjector) applyOperations() error {
...
+		// To find the preferences (identifier used to remove tc filters later)
+		// we need to compare the list of tc filters before and after the addition of new filters
+		// we keep the new strings in the list of tc filters after the addition
+		// and we find the exact preference
+		computedPreferencesPerIface[iface] = []string{}
+		splitOutputBeforeAddition := strings.Split(filtersBeforeAddition[iface], "\n")
+		splitOutputAfterAddition := strings.Split(filtersAfterAddition[iface], "\n")
+
+		// We remove useless strings that a similar when we list filters before and after the addition
+		for _, befLine := range splitOutputBeforeAddition {
```
ptnapoleon, line 451: should be `that are similar`? Should it be `that are identical`??

injector/network_disruption.go
```diff
@@ -374,78 +416,442 @@ func (i *networkDisruptionInjector) applyOperations() error {
...
+				}
+			}
+
+			if toRemoveIdx >= 0 {
+				splitOutputAfterAddition = append(splitOutputAfterAddition[:toRemoveIdx], splitOutputAfterAddition[toRemoveIdx+1:]...)
+			}
+		}
+
+		// We find the preferences
+		// Exampe of string: filter parent 1: protocol all pref 49150 u32 chain 0 fh 802: ht divisor 1
```
ptnapoleon, line 468:
```suggestion
		// Example of string: filter parent 1: protocol all pref 49150 u32 chain 0 fh 802: ht divisor 1
```

injector/network_disruption.go
```diff
@@ -374,78 +416,442 @@ func (i *networkDisruptionInjector) applyOperations() error {
...
+	if err != nil {
+		return nil, fmt.Errorf("couldn't create tc filter: %v", err.Error())
+	}
+
+	filtersAfterAddition, err := i.config.TrafficController.ListFilters(interfaces)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't identify tc filter created, error on tc filter listing: %v", err.Error())
+	}
+
+	for _, iface := range interfaces {
```
ptnapoleon, line 439: Any chance you could break the logic in this for loop out into a new function that gets unit tested? It seems potentially brittle

ptnapoleon (commented):
Doing some local testing, seems to respond very well to changes in the deployment underlying the targeted service. The injector itself doesnt seem to break if you delete the targeted service, though the disruption will start warning you that your spec is invalid.

I was a bit surprised that the injector continues to watch the deployment that underlies the target service, even after the service is deleted, and all pods are recycled. Is that expected?
e.g.
Disruption targets service A.
I delete service A,
I delete all pods in deployment A (that service A used to point to)
New pods are created in deployment A
The Injector adds tc filters against ^ new pods

injector/network_disruption.go
```diff
@@ -374,78 +416,442 @@ func (i *networkDisruptionInjector) applyOperations() error {
...
+				if err := i.handleKubernetesServiceChanges(event, &watcher, interfaces, flowid); err != nil {
+					return fmt.Errorf("couldn't apply changes to tc filters: %w", err)
+				}
+			}
+		case event, ok := <-watcher.kubernetesPodEndpointsWatcher: // We have changes in the service watched
+			if !ok { // channel is closed
+				watcher.kubernetesPodEndpointsWatcher = nil
+			} else {
+				if err := i.handleKubernetesPodsChanges(event, &watcher, interfaces, flowid); err != nil {
+					return fmt.Errorf("couldn't apply changes to tc filters: %w", err)
```
ptnapoleon, line 818: Hmm, in the case of an error, should we try deleting and recreating the watcher, here?

ptnapoleon:
> I was a bit surprised that the injector continues to watch the deployment that underlies the target service, even after the service is deleted, and all pods are recycled. Is that expected?

Okay, I see why this happens now after reading the code. The pod watcher actually uses the same selector as the service, so it doesnt need to rely on the service continuing to exist. However, it seems that while we will detect changes to the pods underlying the service during the duration of the disruption, we don't actually watch for changes to the spec of the service (unless I'm missing where that happens)? I think we should document this (perfectly acceptable) limitation

clairecng:
> Doing some local testing, seems to respond very well to changes in the deployment underlying the targeted service. The injector itself doesnt seem to break if you delete the targeted service, though the disruption will start warning you that your spec is invalid.
> 
> I was a bit surprised that the injector continues to watch the deployment that underlies the target service, even after the service is deleted, and all pods are recycled. Is that expected? e.g. Disruption targets service A. I delete service A, I delete all pods in deployment A (that service A used to point to) New pods are created in deployment A The Injector adds tc filters against ^ new pods

clairecng:
> > I was a bit surprised that the injector continues to watch the deployment that underlies the target service, even after the service is deleted, and all pods are recycled. Is that expected?
> 
> Okay, I see why this happens now after reading the code. The pod watcher actually uses the same selector as the service, so it doesnt need to rely on the service continuing to exist. However, it seems that while we will detect changes to the pods underlying the service during the duration of the disruption, we don't actually watch for changes to the spec of the service (unless I'm missing where that happens)? I think we should document this (perfectly acceptable) limitation

I'm not sure what would be the best to do here.  We could also just stop watching the service and the related pods by closing the goroutine when the service is deleted or add a new watcher. Maybe this could be an improvement to do later as the PR is pretty heavy here, and I'll document it for now?

injector/network_disruption.go
```diff
@@ -53,6 +57,34 @@ type NetworkDisruptionInjectorConfig struct {
...
+}
+
+type DisruptionState struct {
+	State chan InjectorState
+}
+
+// tcServiceFilter describes a tc filter, representing the service filtered and its preferences
+type tcServiceFilter struct {
+	service     networkDisruptionService
+	preferences map[string][]string // identifiers used to delete the filter, a map of interfaces and preferences
```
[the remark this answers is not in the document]
clairecng: Switched the logic to using only one preference (also called priority) per service. It was easier to assign directly the priority at the add filter level so that the logic to discover the priority can be removed.

injector/injector.go
```diff
@@ -16,6 +16,13 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type InjectorState string
+
+const (
+	Injected InjectorState = "injected"
+	Cleaned  InjectorState = "cleaned"
+)
```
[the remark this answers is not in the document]
clairecng: Added a "created" level

injector/network_disruption.go
clairecng: I changed the logic to keep track of the priority in another way: we can directly add a filter with a priority.

injector/network_disruption_test.go
```diff
@@ -273,12 +279,56 @@ var _ = Describe("Failure", func() {
...
+					// Deleting a pod
+					time.Sleep(300 * time.Millisecond)
+					podsWatcher.Delete(fakeEndpoint)
+				}()
+
 			})
 
 			It("should add a filter to redirect targeted traffic on all interfaces on the disrupted band filter on given service cluster IP and endpoints IPs", func() {
+				// Wait for the changes to happen
+				time.Sleep(time.Second * 10)
```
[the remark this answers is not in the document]
clairecng: We still need to wait for some time in order for all the add filters used before the dynamic service resolution to be complete and then run the assertions that we want to target.

docs/changes_handling.md
```diff
@@ -0,0 +1,41 @@
...
+  name: network-drop
+  namespace: chaos-demo
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 1
+  network:
+    drop: 100 # percentage of outgoing packets to drop
+    *ervices: # optional, list of destination Kubernetes services to filter on
```
ptnapoleon, line 20:
```suggestion
    services: # optional, list of destination Kubernetes services to filter on
```

injector/network_disruption_test.go
```diff
@@ -273,22 +278,65 @@ var _ = Describe("Failure", func() {
...
+
+			})
+
+			It("should add a filter for every service and pods filtered on", func() {
+				// wait for all the addFilters at the beginning of injection to complete
+				time.Sleep(5 * time.Second)
+				priority := uint32(49153)
+
+				Eventually(func() bool {
+					return tc.AssertCalled(GinkgoT(), "AddFilter", []string{"lo", "eth0", "eth1"}, "1:0", mock.Anything, mock.Anything, "nil", "172.16.0.1/32", 0, 80, "TCP", "1:4")
```
ptnapoleon, line 309: Does this work? I've previously seen that `tc.AssertCalled` will directly fail the test, because it's being passed `GinkgoT()`, rather than just returning `false`. But, maybe you're passing `T` better than I was?

injector/network_disruption.go
```diff
@@ -70,19 +104,25 @@ func NewNetworkDisruptionInjector(spec v1beta1.NetworkDisruptionSpec, config Net
 		config.DNSClient = network.NewDNSClient()
 	}
 
-	return networkDisruptionInjector{
+	config.State = DisruptionState{}
```
ptnapoleon, line 107: Could we set `config.State.State` to created here? (or does that cause problems)

injector/network_disruption.go
```diff
@@ -135,11 +175,19 @@ func (i networkDisruptionInjector) Inject() error {
...
+		i.config.State.State <- Injected
+	}()
+
 	return nil
 }
 
 // Clean removes all the injected disruption in the given container
-func (i networkDisruptionInjector) Clean() error {
+func (i *networkDisruptionInjector) Clean() error {
+	go func() {
```
ptnapoleon, line 187: Should we defer this? We dont want to set this to Cleaned before we're sure the cleaning worked, right?

injector/network_disruption.go
```diff
@@ -198,6 +246,8 @@ func (i networkDisruptionInjector) Clean() error {
 //           |- (4:) <-- second operation
 //             ...
 func (i *networkDisruptionInjector) applyOperations() error {
+	i.tcFilterPriority = 49150
```
ptnapoleon, line 249: Magic number? Could you define this constant at the top of the file in that case? Thanks!

ptnapoleon (commented):
I left a few more questions and comments, but this looks really great!

injector/network_disruption.go
clairecng: This is in the range of the numbers assigned by tc automatically! I'll add a comment on that

injector/network_disruption_test.go
clairecng: It works, I didn't encounter this problem

ptnapoleon (approved):

chart/templates/injector_role.yaml
```diff
@@ -16,9 +16,14 @@ rules:
...
+  - watch
 - apiGroups:
   - ""
   resources:
   - services
+  - namespaces
   verbs:
+  - list
   - get
+  - update
```
nikos912000, line 28: Hi @clairecng, @ptnapoleon 👋

I was going through the 6.0.0 release notes.
Was there any reason behind these additional permissions? Being a bit careful, especially with `update` permissions on services and namespaces.

ptnapoleon: Ahh, hmm, @clairecng why did we need update for services actually?

nikos912000: I noticed these permissions have now been removed. Thanks for that!

Issue #511, opened 2022-02-16
User Issue: Status not being a subresource causes issues with control planes


nikos912000:
**Describe the bug**
Right now, the status is not defined as a [subresource](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource) in the CRD. This means that the entire object gets updated, including its generation.

This causes issues, especially when integrating with control planes, like Kubefed; the control plane sees the object's generation getting updated and reconciles, causing an infinite loop.

I was curious if there was a particular reason behind the implementation. Is it possible to change this?

**To Reproduce**
The issue is more obvious when having a control plane.

**Expected behavior**
The status is defined as a subresource in the CRD. This would allow updating just the status rather than the entire object.

**Additional information**
Subresources in Kubebuilder: https://book.kubebuilder.io/reference/generating-crd.html?highlight=subresourc#subresources

ptnapoleon:
>  there was a particular reason behind the implementation

Not especially, it was just extra work at the time, and we didn't have a need for it. I think we're willing to change it

nikos912000:
That would be very useful. We are integrating with Kubefed and Spinnaker and this unfortunately causes a lot of issues :(

I hope it makes sense from a design point of view - this is how most controllers define the status afaik.

From a quick look, the following changes are needed:
- [This](https://github.com/DataDog/chaos-controller/blob/a16e95d9f238d5eed6cee2c07430ecdb3d8e1c90/controllers/disruption_controller.go#L313) is the only place where the controller itself updates the status.
- The webhooks patch the object, so these calls need to be updated.
- And then the CRD plus the addition of the [kubebuilder annotation](https://book.kubebuilder.io/reference/generating-crd.html?highlight=subresourc#status).

PS: Thanks for being super fast with any issues!

ptnapoleon:
Closed by https://github.com/DataDog/chaos-controller/pull/520 , hopefully? We will cut a release this week

nikos912000:
Amazing, thank you @ptnapoleon!

PR #514, opened 2022-02-18
Make status a subresource.

nikos912000:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

Makes the status a subresource in the CRD. See #511.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -133,6 +133,7 @@ type DisruptionStatus struct {
 
 // Disruption is the Schema for the disruptions API
 // +kubebuilder:resource:shortName=dis
+// +kubebuilder:subresource:status
 type Disruption struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
+++ b/chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
@@ -18,6 +18,8 @@ spec:
     singular: disruption
   preserveUnknownFields: false
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       description: Disruption is the Schema for the disruptions API
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -310,7 +310,7 @@ func (r *DisruptionReconciler) updateInjectionStatus(instance *chaosv1beta1.Disr
 	// update instance status
 	instance.Status.InjectionStatus = status
 
-	if err := r.Client.Update(context.Background(), instance); err != nil {
+	if err := r.Client.Status().Update(context.Background(), instance); err != nil {
 		return false, err
 	}
 
@@ -632,7 +632,7 @@ func (r *DisruptionReconciler) handleChaosPodsTermination(instance *chaosv1beta1
 		}
 	}
 
-	return r.Update(context.Background(), instance)
+	return r.Status().Update(context.Background(), instance)
 }
 
 // ignoreTarget moves the given target from the list of targets to the list of ignored targets
@@ -654,7 +654,7 @@ func (r *DisruptionReconciler) ignoreTarget(instance *chaosv1beta1.Disruption, t
 		instance.Status.IgnoredTargets = append(instance.Status.IgnoredTargets, target)
 	}
 
-	return r.Update(context.Background(), instance)
+	return r.Status().Update(context.Background(), instance)
 }
 
 // selectTargets will select min(count, all matching targets) random targets (pods or nodes depending on the disruption level)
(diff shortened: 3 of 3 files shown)


nikos912000:
@ptnapoleon @Devatoria we were testing these changes and would like to get some feedback.
These are causing issues with the user info webhook. The user info is there in the request, but patching the status/object does not have any effect.

This then causes the controller to panic:
```
{"level":"info","ts":1645192724108.0188,"caller":"webhook/user_info.go:50","message":"storing user info in disruption","name":"container-failure-nikos","namespace":"chaos-demo","req":{"username":"minikube-user","groups":["system:masters","system:authenticated"]}}
2022/02/18 13:58:44 http: panic serving 10.244.0.1:37045: runtime error: invalid memory address or nil pointer dereference
goroutine 1002 [running]:
net/http.(*conn).serve.func1()
	/usr/local/Cellar/go/1.17.2/libexec/src/net/http/server.go:1801 +0xb9
panic({0x1729aa0, 0x289cd90})
	/usr/local/Cellar/go/1.17.2/libexec/src/runtime/panic.go:1047 +0x266
github.com/DataDog/chaos-controller/api/v1beta1.(*Disruption).getMetricsTags(0xc000ade000)
	/Users/nkatirtzis/projects/chaos-controller/api/v1beta1/disruption_webhook.go:140 +0xdd
github.com/DataDog/chaos-controller/api/v1beta1.(*Disruption).ValidateCreate(0xc000ade000)
	/Users/nkatirtzis/projects/chaos-controller/api/v1beta1/disruption_webhook.go:88 +0x30f
sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*validatingHandler).Handle(_, {_, _}, {{{0xc0002f6990, 0x24}, {{0xc0004ae9a8, 0x13}, {0xc000a22a60, 0x7}, {0xc000a22a70, ...}}, ...}})
```

A few questions:
- Would we need any changes in the [webhook](https://github.com/DataDog/chaos-controller/blob/main/webhook/user_info.go)? We tried replacing the logic with a patch operation on the status but still...
- Would it make sense to treat the UserInfo as optional [here](https://github.com/DataDog/chaos-controller/blob/01dc8e45c4a9bea821862ff3ee3c1d7c521a2b19/api/v1beta1/disruption_webhook.go#L140)? Imo the controller should not panic if this info is not there?

ptnapoleon:
>Would we need any changes in the [webhook](https://github.com/DataDog/chaos-controller/blob/main/webhook/user_info.go)? We tried replacing the logic with a patch operation on the status but still...

Okay, this has jogged my memory. This was the issue I had, and I never solved it, so I gave up on making status a subresource.

> Would it make sense to treat the UserInfo as optional [here](https://github.com/DataDog/chaos-controller/blob/01dc8e45c4a9bea821862ff3ee3c1d7c521a2b19/api/v1beta1/disruption_webhook.go#L140)? Imo the controller should not panic if this info is not there?

Yes that's valid, but we still need to solve the update within the webhook?

ptnapoleon:
> We don't make the status a subresource but we replace any status updates as this PR does. This may work; the generation does not get updated which may fix the infinite loop we're having with the control plane. Still need to test end-to-end.

If this works, that does seem simplest (if worst?)

> We make the webhook optional, behind a flag. I think that would still be fair since the user info is complementary?

We should be okay with this, afaik

nikos912000:
> If this works, that does seem simplest (if worst?)
Indeed. I will test if that works.

Regarding the user status webhook being optional, it would be nice to do this anyways? Mutating webhooks can cause many issues, especially when there is a control plane.
On our side we might have to disable them soon unfortunately...

nikos912000:
@ptnapoleon I was thinking about this more while trying different solutions.
I would argue the `UserInfo` section is metadata. Wouldn't it make more sense if this info was an annotation instead?  And since it would be challenging to model that in annotations, wouldn't just the `username` be enough?

Looking at other projects, the ones that store the entire UserInfo section in the CRD have this as part of the spec. I suspect the reason is that mutating webhooks don't play nicely with subresources.

Still testing whether it all works without making the status a subresource...

nikos912000:
```
We don't make the status a subresource but we replace any status updates as this PR does. This may work; the generation does not get updated which may fix the infinite loop we're having with the control plane. Still need to test end-to-end.
```
After running a few tests, I was wrong on that. The `/status` endpoint that gets called on `Status().Update` is only available when the status is a subresource.

Would you be open to moving the UserInfo section either to the spec or to metadata?

ptnapoleon:
> Would you be open to moving the UserInfo section either to the spec or to metadata?

Would this end up as labels? /cc @Devatoria

nikos912000:
> I can, initially, open a PR that makes the UserInfo webhook optional, (as well as anywhere we check that code in the reconcile loop)?

This would be useful for sure. If this makes any mutating webhooks optional, behind a flag, that would be brilliant.

After trying a few workarounds, one solution that would work for everyone is the following:
- The status becomes a subresource. This has been tested with Kubefed and works as expected. Otherwise unfortunately we can't do only status updates.
- We move the UserInfo outside of the status block. We were thinking of modelling this as a JSON value in the annotations? 

We are working on that solution so happy to contribute it as part of this PR.

nikos912000:
Closing this in favour of David's PR: https://github.com/DataDog/chaos-controller/pull/518
Thanks for the feedback @ptnapoleon.

PR #517, opened 2022-02-22
Use config to gate mutating hooks

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Dont initialize the handler webhook if the handler is disabled
- Require a config option to initialize the user info webhook

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -137,12 +137,15 @@ func (r *Disruption) getMetricsTags() []string {
 	tags := []string{
 		"name:" + r.Name,
 		"namespace:" + r.Namespace,
-		"username:" + r.Status.UserInfo.Username,
 	}
 
-	// add groups
-	for _, group := range r.Status.UserInfo.Groups {
-		tags = append(tags, "group:"+group)
+	if r.Status.UserInfo != nil {
+		tags = append(tags, "username:"+r.Status.UserInfo.Username)
+
+		// add groups
+		for _, group := range r.Status.UserInfo.Groups {
+			tags = append(tags, "group:"+group)
+		}
 	}
 
 	// add selectors
--- a/chart/templates/configmap.yaml
+++ b/chart/templates/configmap.yaml
@@ -27,6 +27,7 @@ data:
       imagePullSecrets: {{ .Values.images.pullSecrets }}
       defaultDuration: {{ .Values.controller.defaultDuration }}
       expiredDisruptionGCDelay: {{ .Values.controller.expiredDisruptionGCDelay }}
+      userInfoHook: {{ .Values.controller.userInfoHook }}
       webhook:
         {{- if .Values.controller.webhook.generateCert }}
         certDir: /tmp/k8s-webhook-server/serving-certs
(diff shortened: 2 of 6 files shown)


nikos912000:
Thanks for that!

Two comments:
- Can we also update the [Slack notifier](https://github.com/DataDog/chaos-controller/blob/349e69d37ae7c809a93f355b9b2b7eea5e823a4d/eventnotifier/slack/slack.go#L112) which expects the UserInfo to be there?
- The scope of this PR includes the user info and pod init mutating webhooks. There is one more for the [duration](https://github.com/DataDog/chaos-controller/blob/349e69d37ae7c809a93f355b9b2b7eea5e823a4d/api/v1beta1/disruption_webhook.go#L48). It'd be great to have that behind a flag as well.

PS: I like the idea of having separate flags; the one for the Duration is more critical, regardless of the fact that we may need to disable that internally.

Devatoria (approved):

ptnapoleon:
> Can we also update the [Slack notifier](https://github.com/DataDog/chaos-controller/blob/349e69d37ae7c809a93f355b9b2b7eea5e823a4d/eventnotifier/slack/slack.go#L112) which expects the UserInfo to be there?

Seems that the slack notifier just needs disabled if it always requires the UserInfo to be present?

nikos912000:
> Seems that the slack notifier just needs disabled if it always requires the UserInfo to be present?

I'd say the user info is complementary there? It would still be useful without that info.

nikos912000:
> Seems that the slack notifier just needs disabled if it always requires the UserInfo to be present?

I'd say the user info is complementary there? It would still be useful without that info.

ptnapoleon:
@nikos912000 Without setting a default duration (because we've disabled the mutating webhook), it seems that nothing errors, but any disruption without a duration set will immediately be considered expired. Are you planning on always having `spec.duration` set? Is there a way we can fix the webhook such that it doesnt cause problems for you?

ptnapoleon:
> I'd say the user info is complementary there? It would still be useful without that info.

Right now it only knows who to send the slack message to based on the UserInfo. We could add a config value for a slack channel to send all these updates to?

nikos912000:
> @nikos912000 Without setting a default duration (because we've disabled the mutating webhook), it seems that nothing errors, but any disruption without a duration set will immediately be considered expired. Are you planning on always having `spec.duration` set? Is there a way we can fix the webhook such that it doesnt cause problems for you?

Our intention is to always set the duration on the Spinnaker side.
Happy to keep the duration mutating webhook for now and until we have a clear ask for that :)

nikos912000:
> > I'd say the user info is complementary there? It would still be useful without that info.
> 
> Right now it only knows who to send the slack message to based on the UserInfo. We could add a config value for a slack channel to send all these updates to?

Oh ok. To be fair we haven't enabled the Slack integration (yet).
Whatever is easier for you. We can always update that later.

ptnapoleon:
In that case, I'm going to merge the current PR as-is. I'll open a ticket internally for "Let slack notifier be useful even w/o user info".

PR #518, opened 2022-02-23
Makes the status a subresource in the CRD.

expFlower:
Signed-off-by: David Fowler <dafowler@expediagroup.com>

## What does this PR do?

- [ ] Adds new functionality
- [x] Alters existing functionality
- [ ] Fixes a bug
- [ ] Improves documentation or testing

Please briefly describe your changes as well as the motivation behind them:
- Makes the status a subresource in the CRD. See https://github.com/DataDog/chaos-controller/issues/511.
- Store UserInfo in annotations

## Code Quality Checklist

- [ ] The documentation is up to date.
- [ ] My code is sufficiently commented and passes continuous integration checks.
- [ ] I have signed my commit (see [Contributing Docs](../CONTRIBUTING.md)).

## Testing

- [ ] I leveraged continuous integration testing
    - [ ] by depending on existing `unit` tests or `end-to-end` tests.
    - [ ] by adding new `unit` tests or `end-to-end` tests.
- [ ] I manually tested the following steps:
    - [x] locally.
    - [ ] as a canary deployment to a cluster.

PR #520, opened 2022-02-24
Duplicate of #518: Make status a subresource and move UserInfo to Annotations

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Replaces https://github.com/DataDog/chaos-controller/pull/518
- Had to push myself so that I could test internally
- Also merged with main

## Code Quality Checklist

## Testing

I plan to deploy this internally for testing


ptnapoleon:
> Why should we move UserInfo to annotations and how is that better than keeping it in the status resource?

From @nikos912000 's comment in https://github.com/DataDog/chaos-controller/pull/514, it seems like they need either status as a subresource or UserInfo outside of status? So I suppose we dont have to make this second change?

expFlower:
> > Why should we move UserInfo to annotations and how is that better than keeping it in the status resource?
> 
> From @nikos912000 's comment in #514, it seems like they need either status as a subresource or UserInfo outside of status? So I suppose we dont have to make this second change?

Yes, as per #514 we need status to be a subresource as described in issue #511. When you make the status a subresource the webhook can't patch the status in the admission request and add the UserInfo. The UserInfo was therefore moved to an Annotation, which we believed to be a good location for this data.

Devatoria:
> > > Why should we move UserInfo to annotations and how is that better than keeping it in the status resource?
> > 
> > 
> > From @nikos912000 's comment in #514, it seems like they need either status as a subresource or UserInfo outside of status? So I suppose we dont have to make this second change?
> 
> Yes, as per #514 we need status to be a subresource as described in issue #511. When you make the status a subresource the webhook can't patch the status in the admission request and add the UserInfo. The UserInfo was therefore moved to an Annotation, which we believed to be a good location for this data.

Ok got it, thanks! No problem for me to move it to annotations, we just need to do some extra testing on our side about using status as a subresource.

Devatoria (approved):

PR #523, opened 2022-03-07
Dynamic Targeting

nathantournant:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug

### Current state of targeting
- As of right now, the controller does not continue targeting in time; at appliance, the disruption selects its targets once and will stick to them. 
- This is a very useful behaviour for a couple of reasons:
  - it makes the monitoring of targeting simpler to manage, more stable and transparent
  - it limits the impact of a disruption as new pods (which could compensate for the disruption) will not be targeted

Non-dynamic targeting will stay the default behaviour of the controller for safety and simplicity reason, as it fits most use cases. However, it is also a limit for some situations: a notorious case being the constantly-deploying services. If a service's pods are rolled out every 5-10 minutes, it is impossible to target a service properly for an extended period of time, as new pods will be left intact. Hence the implementation of Dynamic Targeting within the `chaos-controller`

### New Feature: Dynamic Targeting
Dynamic Targeting will allow the controller to follow-up on a cluster/deployment/target/etc. state in order to update the disruption and maintain it's targeting objective. With dynamic targeting activated, a disruption will watch the cluster's activity in order to disrupt new targets (if required) and stop counting dead targets as targets (eg. a 100%-count disruption will constantly disrupt 100% of old and new fitting targets within the namespace)

To achieve this result, the controller watches over the disruption's given namespace and target label selector.

Activity diagram for Dynamic Targeting (2)
Targets Update_Reconcile Logic (2)

## Code Quality Checklist

## Testing

    - On minikube
        - Install the controller on minikube
        - Edit any pod disruption with `dynamicTargeting: true` and `count: 100%` (or `50%`)
        - `examples/demo.yaml`: Edit the `demo-curl` container CPU limit to 0.1 (currently 1)
        - Deploy the demo
        - apply the chosen disruption
        - edit the demo-curl amount of replicas to *X* (arbitrary number) -- apply
        - Observe the number of chaos-pods go to *X*
        - edit the demo-curl amount of replicas to *X/2* (lower number) -- apply
        - Observe the number of chaos-pods go to *X/2*
        - Start again with `dynamicTargeting: false` and this time the amount of pods will not change
    - [X] locally.
    - [X] as a canary deployment to a cluster.


docs/features.md
```diff
@@ -52,6 +52,16 @@ The `Disruption` resource uses [label selectors](https://kubernetes.io/docs/conc
...
+
+By default, targeting is limited to a single target selection at the disruption's creation. It allows for more controlled disruption impact and propagation, as the targets will never change and canbe compensated in case of issue. Its major limit is not being able to follow targets through deployments/rollouts.
+By setting the `DynamicTargeting` flag to `true` in the Disruption's yaml spec, you activate a constant re-targeting for this disruption. This means at any given time, any target within the selector's scope will be added to the target list and be disrupted.
+This is a feature to use with care, as it can quickly get out of control: per example, a disruption targeting 100% of an application's pod will affect all existing **and** future pods which can appear once the disruption started. As long as this 100% disruption exists, there will be no spared pod.
+Even a 50%-set disruption, if the disruption gets targeted pods to die, will constantly re-target 50% of all selector-fitting pods and end-up killing every pods.
+`DynamicTargeting` behavior design choices:
+- the controller will consider as a still-alive target any pod that exists - regardeless of its state.
```
ptnapoleon, line 62:
```suggestion
- the controller will consider as a still-alive target any pod that exists - regardless of its state.
```

docs/features.md
```diff
@@ -52,6 +52,16 @@ The `Disruption` resource uses [label selectors](https://kubernetes.io/docs/conc
...
+- the controller will reconcile/update its targets list on any chaos pod or selector mouvement (create, update, delete)
```
ptnapoleon, line 63:
```suggestion
- the controller will reconcile/update its targets list on any chaos pod or selector movement (create, update, delete)
```

docs/features.md
```diff
@@ -52,6 +52,16 @@ The `Disruption` resource uses [label selectors](https://kubernetes.io/docs/conc
 
 **NOTE:** If you are targeting pods, the disruption must be created in the same namespace as the targeted pods.
 
+### DynamicTargeting
+
+By default, targeting is limited to a single target selection at the disruption's creation. It allows for more controlled disruption impact and propagation, as the targets will never change and canbe compensated in case of issue. Its major limit is not being able to follow targets through deployments/rollouts.
+By setting the `DynamicTargeting` flag to `true` in the Disruption's yaml spec, you activate a constant re-targeting for this disruption. This means at any given time, any target within the selector's scope will be added to the target list and be disrupted.
+This is a feature to use with care, as it can quickly get out of control: per example, a disruption targeting 100% of an application's pod will affect all existing **and** future pods which can appear once the disruption started. As long as this 100% disruption exists, there will be no spared pod.
+Even a 50%-set disruption, if the disruption gets targeted pods to die, will constantly re-target 50% of all selector-fitting pods and end-up killing every pods.
```
ptnapoleon, line 60:
```suggestion
Even a 50%-set disruption, if the disruption gets targeted pods to die, will constantly re-target 50% of all selector-fitting pods and end up killing every pod.
```

Is that true? I think I know what you mean here, but as long as k8s is scheduling new pods, we'll keep at least half of them up, so we dont really kill "every" pod, though we may end up killing any given pod eventually as the disruption/rescheduling continues

docs/features.md
```diff
@@ -52,6 +52,16 @@ The `Disruption` resource uses [label selectors](https://kubernetes.io/docs/conc
 
 **NOTE:** If you are targeting pods, the disruption must be created in the same namespace as the targeted pods.
 
+### DynamicTargeting
+
+By default, targeting is limited to a single target selection at the disruption's creation. It allows for more controlled disruption impact and propagation, as the targets will never change and canbe compensated in case of issue. Its major limit is not being able to follow targets through deployments/rollouts.
+By setting the `DynamicTargeting` flag to `true` in the Disruption's yaml spec, you activate a constant re-targeting for this disruption. This means at any given time, any target within the selector's scope will be added to the target list and be disrupted.
+This is a feature to use with care, as it can quickly get out of control: per example, a disruption targeting 100% of an application's pod will affect all existing **and** future pods which can appear once the disruption started. As long as this 100% disruption exists, there will be no spared pod.
```
ptnapoleon, line 59: Let's discuss this within the team, as I personally think we may want to default to this being true in the future, as a lot of users seem to expect this behavior

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
ptnapoleon, line 1: What is your local go version? I dont think this should get removed normally

api/v1beta1/disruption_types.go
```diff
@@ -317,3 +317,43 @@ func ReadUnmarshal(path string) (*Disruption, error) {
...
+func (status *DisruptionStatus) RemoveTargets(toRemoveTargetsCount int) {
+	for i := 0; i < toRemoveTargetsCount; i++ {
+		index := rand.Intn(len(status.Targets)) //nolint:gosec
+		status.Targets[len(status.Targets)-1], status.Targets[index] = status.Targets[index], status.Targets[len(status.Targets)-1]
+		status.Targets = status.Targets[:len(status.Targets)-1]
+	}
+}
+
+// contains returns true when the given string is present in the given slice
+func contains(s []string, str string) bool {
```
ptnapoleon, line 351: This is a re-implement of the function in `controller/helpers.go` right? Could we extract from both places, and move to some sort of util or helpers package?

docs/features.md
```diff
@@ -52,6 +52,16 @@ The `Disruption` resource uses [label selectors](https://kubernetes.io/docs/conc
 
 **NOTE:** If you are targeting pods, the disruption must be created in the same namespace as the targeted pods.
 
+### DynamicTargeting
+
+By default, targeting is limited to a single target selection at the disruption's creation. It allows for more controlled disruption impact and propagation, as the targets will never change and canbe compensated in case of issue. Its major limit is not being able to follow targets through deployments/rollouts.
```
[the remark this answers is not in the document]
nathantournant: I meant if disruption neutralises the pods, new pods can be deployed to keep the service active/bring back service activity in case of an issue with the disruption.
I agree it's not a clean formulation.

docs/features.md
nathantournant: Yes that's accurate; though that will most likely create a race condition and it'll be a weird balance to maintain.
I'll test this locally with a pod-killing disruption.

docs/features.md
nathantournant: I'll start a thread !

api/v1beta1/disruption_types.go
nathantournant: Sounds like a fair package to create

api/v1beta1/zz_generated.deepcopy.go
nathantournant: I am on 1.16 which is recommended
Not sure why this disappears quite constantly on my computer

api/v1beta1/disruption_types.go
nathantournant: Created an `utils` package with only this for now
I can imagine with time we can move various things there

api/v1beta1/disruption_types.go
```diff
@@ -317,3 +318,32 @@ func ReadUnmarshal(path string) (*Disruption, error) {
 
 	return &parsedSpec, nil
 }
+
+// RemoveDeadTargets removes targets not found in matchingTargets from the targets list
+func (status *DisruptionStatus) RemoveDeadTargets(matchingTargets []string) {
+	for index := 0; index < len(status.Targets); index++ {
+		if !utils.Contains(matchingTargets, status.Targets[index]) {
+			status.Targets[len(status.Targets)-1], status.Targets[index] = status.Targets[index], status.Targets[len(status.Targets)-1]
+			status.Targets = status.Targets[:len(status.Targets)-1]
```
ptnapoleon, line 327: Does this work? We're modifying a list while we're in the middle of traversing it, won't we end up skipping over some indices?

ptnapoleon: I ran some tests and I can confirm this function won't work as is, due to the aforementioned issue

controllers/disruption_controller.go
```diff
@@ -272,6 +329,97 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+		if err != nil {
+			return fmt.Errorf("cache gen error: %w", err)
+		}
+
+		info.AddEventHandler(DisruptionSelectorHandler{disruption: instance, reconciler: r})
+
+		// start the cache with a cancelable context and attach it to the controller as a watch source
+		ch := make(chan error)
+		cacheCtx, cacheCancelFunc := context.WithCancel(context.TODO())
```
ptnapoleon, line 389: I think it may be helpful to set a timeout here, equivalent to a bit longer than the disruption duration + GC. Right now the only way we remove these caches is upon working disruption deletion, but that isn't technically _guaranteed_ to happen (though it does in the 99.9% case, I know). After the timeout, we can just go ahead and cancel the cache and delete it.

controllers/disruption_controller.go
```diff
@@ -272,6 +329,97 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+
+	disNamespacedName := types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}
+	disCompleteSelector, err := targetselector.GetLabelSelectorFromInstance(instance)
+
+	if err != nil {
+		return fmt.Errorf("error getting instance selector: %w", err)
+	}
+	// if it doesn't exist, create the cache context to re-trigger the disruption
+	if r.CachesCancel[disNamespacedName] == nil {
```
ptnapoleon, line 345: We run the teeny tiny risk here that a Disruption can be deleted without the proper teardown, then re-created with the same name but a different spec, and that could screw up the cache. I dont have a suggested fix, it's extremely unlikely anyway. Maybe just leave a comment that we are accepting that failure case?

controllers/disruption_controller.go
```diff
@@ -557,134 +727,105 @@ func (r *DisruptionReconciler) handleChaosPodsTermination(instance *chaosv1beta1
 	}
 
 	for _, chaosPod := range chaosPods {
-		removeFinalizer := false
-		ignoreStatus := false
-		target := chaosPod.Labels[chaostypes.TargetLabel]
+		r.terminateChaosPod(instance, chaosPod)
```
ptnapoleon, line 730: I think we should rename this function, otherwise this sort of reads like we always terminate every chaos pod in every reconcile loop. I think `maybeTerminateChaosPod`, `handleChaosPodTermination`, etc?

docs/features.md
nathantournant: Follow-Up: decided to have a "StaticTargeting" flag defaulted to true as a soft transition
Default behaviour will be in time become StaticTargeting set to false

api/v1beta1/disruption_types.go
nathantournant: Rookie mistake -- will edit asap

controllers/disruption_controller.go
nathantournant: This seems fair. However I need to find a way to keep the sync between the context delete and the cancel func array cleanup. I'll find something.

controllers/disruption_controller.go
```diff
@@ -343,91 +491,120 @@ func (r *DisruptionReconciler) updateInjectionStatus(instance *chaosv1beta1.Disr
...
+	if len(targetChaosPods) == 0 {
+		r.Recorder.Event(instance, corev1.EventTypeWarning, "EmptyDisruption", fmt.Sprintf("No disruption recognized for \"%s\" therefore no disruption applied.", instance.Name))
+		return nil
+	}
 
-					return fmt.Errorf("error creating chaos pod: %w", err)
-				}
+	// create injection pods
+	for _, chaosPod := range targetChaosPods {
+		// check if an injection pod already exists for the given (instance, namespace, disruption kind) tuple
```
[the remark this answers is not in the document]
nathantournant: As generated here:
https://github.com/DataDog/chaos-controller/blob/727a7b00a8543da8a6b0b2cf692bb99feb36c6fc/controllers/disruption_controller.go#L1229-L1232

controllers/disruption_controller.go
nathantournant: yes, fixed

api/v1beta1/zz_generated.deepcopy.go
ptnapoleon: Ahh, I think it's because this is added with go 1.17, which I use. We can discuss this offline then

controllers/disruption_controller.go
nathantournant: I just pushed a fix; on top of the timeout, the reconciler will now store both the cacheContext and its cancel function; if [`cacheContext.Err()`](https://pkg.go.dev/context#Context) is not nil, the map entry will be deleted. It seems bit heavy on the reconcile loop but I see no other solution than to check either periodically or with every reconcile.
Checking periodically (i.e. like the metrics) would add a new timing logic we haven't introduced before and I'm afraid it would be very unclear.
Obviously open to debate this if there's a disagreement.

controllers/disruption_controller.go
nathantournant: Map key is now a hash of the disruption's `Spec` (`Count` excluded, like the admission hook) + Namespaced Name. It  excludes the scenario you mentionned.
I imagine it is still possible to get something messy if a single disruption is manually pulsed and a race condition occurs on the cache deletion/creation, and I'll set up a monitor for such a leak.

controllers/disruption_controller.go
```diff
@@ -272,6 +334,132 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		}
+	}
+
+	for _, key := range deletionList {
+		delete(r.CacheContextStore, key)
+	}
+
+	// remove check when StaticTargeting is defaulted to false
+	if instance.Spec.StaticTargeting == nil {
+		r.log.Debugw("StaticTargeting pointer is nil")
```
ptnapoleon, line 354: Worth including info about the `instance` here, maybe at least the name?

ptnapoleon: Or are we guaranteed that `r.log` is already set to include those values correctly?

controllers/disruption_controller.go
```diff
@@ -272,6 +334,132 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+		info.AddEventHandler(DisruptionSelectorHandler{disruption: instance, reconciler: r})
+
+		// start the cache with a cancelable context and duration, and attach it to the controller as a watch source
+		ch := make(chan error)
+		cacheCtx, cacheCancelFunc := context.WithTimeout(context.Background(), instance.Spec.Duration.Duration()+*r.ExpiredDisruptionGCDelay*2)
+
+		go func() { ch <- cache.Start(cacheCtx) }()
+
+		r.CacheContextStore[disCacheHash] = CtxTuple{cacheCtx, cacheCancelFunc}
```
ptnapoleon, line 423: A context doesn't remember its own cancel func? that's disappointing, but this is a good solution

controllers/disruption_controller.go
```diff
@@ -343,91 +531,128 @@ func (r *DisruptionReconciler) updateInjectionStatus(instance *chaosv1beta1.Disr
...
+	chaosPods, err := r.getChaosPods(instance, nil)
+	if err != nil {
+		return fmt.Errorf("error getting chaos pods: %w", err)
+	}
+
+	for _, chaosPod := range chaosPods {
+		if !utils.Contains(instance.Status.Targets, chaosPod.Labels[chaostypes.TargetLabel]) {
+			r.deleteChaosPod(instance, chaosPod)
+		} else {
+			chaosPodsMap[chaosPod.Labels[chaostypes.TargetLabel]+"-"+chaosPod.Labels[chaostypes.DisruptionKindLabel]] = chaosPod
```
ptnapoleon, line 550: What do you think about changing `chaosPodsMap` to be `map[string]map[string]corev1.Pod{}` where the inner key is the kind, so that we aren't using composite keys like this?

ptnapoleon: It also seems like we don't ever check the value of this map? We can just use `bool` in that case, for "it exists", if that's all we really care about, rather than store whole Pods (we're not even storing pointers)

controllers/disruption_controller.go
```diff
@@ -343,91 +531,128 @@ func (r *DisruptionReconciler) updateInjectionStatus(instance *chaosv1beta1.Disr
...
-				return fmt.Errorf("error getting target pod container ID: %w", err)
+			if err = r.createChaosPods(instance, target); err != nil {
+				return fmt.Errorf("error creating chaos pods: %w", err)
 			}
 
-			// get IP of targeted pod
-			targetPodIP = pod.Status.PodIP
-		case chaostypes.DisruptionLevelNode:
-			targetNodeName = target
+			break
```
ptnapoleon, line 569: Why `break` here?

controllers/disruption_controller.go
```diff
@@ -557,134 +775,105 @@ func (r *DisruptionReconciler) handleChaosPodsTermination(instance *chaosv1beta1
...
+			r.log.Errorw("error removing chaos pod finalizer", "error", err, "chaosPod", chaosPod.Name)
+
+			return
 		}
-	}
+	} else {
+		// if the chaos pod finalizer must not be removed and the chaos pod must not be deleted
+		// and the cleanup status must not be ignored, we are stuck and won't be able to remove the disruption
+		r.log.Infow("instance seems stuck on removal for this target, please check manually", "target", target, "chaosPod", chaosPod.Name)
+		r.Recorder.Event(instance, corev1.EventTypeWarning, "StuckOnRemoval", "Instance is stuck on removal because of chaos pods not being able to terminate correctly, please check pods logs before manually removing their finalizer")
```
ptnapoleon, line 873: When you merge in main, make sure you get the latest change to this notice

ptnapoleon (approved):
A few comments, but looks great!

controllers/disruption_controller.go
nathantournant: The logger [includes the instance name + namespace](https://github.com/DataDog/chaos-controller/blob/1d77e8d736679c6049640692de4f841e1decb21e/controllers/disruption_controller.go#L141) in every log

controllers/disruption_controller.go
nathantournant: The scenario is reached when there are multiple kinds on the disruption, and multiple chaos pods for one target. In the event a chaos pod is missing, `r.createChaosPods` is launched, which will attempt to create all chaos pods for a target (and cancel the creation of the already existing ones with a debug-only line). It shouldn't run again for this target.
I think this logic is fine, we can discuss it internally elsewhere.
--
If deemed problematic, this should be adjusted later in another PR; chaos pods could indeed be "created" individually for each target+disruption kind and not bundled like this -- but that's not the primary focus of this PR that is already a heavy bit of change.

docs/features.md
```diff
@@ -48,16 +48,19 @@ If a `pulse` is not specified, then a disruption will not be pulsing.
 
 ## Targeting
 
+**NEW:** StaticTargeting currently defaults to true. It will default to false after some transition time. Read StaticTargeting.
+
```
ptnapoleon, line 52: You can link directly to the StaticTargeting header within the doc here, That would be nice to turn "Read StaticTargeting" into a hyperlink

nathantournant: Done, TIL

api/disruption_types_test.go
```diff
@@ -0,0 +1,222 @@
...
+			Targets: []string{"0.0.0.0", "1.1.1.1", "2.2.2.2", "3.3.3.3", "4.4.4.4"},
+		}
+	})
+
+	Context("asks to remove no targets", func() {
+		BeforeEach(func() {
+			toRemoveTargetsCount = 0
+		})
+
+		It("expects to remove part of the targets", func() {
```
ptnapoleon, line 184:
```suggestion
		It("expects to remove none of the targets", func() {
```

api/disruption_types_test.go
```diff
@@ -0,0 +1,222 @@
...
+		When("matchingTargets contains exactly all the current targets", func() {
+			It("expects status.Targets to be intact", func() {
+				status.RemoveDeadTargets(matchingTargets)
+				Expect(status.Targets).To(HaveLen(5))
+				Expect(status.Targets).To(Equal(matchingTargets))
+			})
+		})
+
+		When("matchingTargets has all current targets and more", func() {
+			BeforeEach(func() {
```
ptnapoleon, line 49:
```suggestion
			JustBeforeEach(func() {
```

to make the timing more explicit

nathantournant: done

PR #525, opened 2022-03-17
Refactor end-to-end tests

nikos912000:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

Makes the end-to-end tests more generic so they can be executed in non-minikube clusters.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -45,8 +45,8 @@ import (
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
 const (
-	timeout = time.Second * 45
-	HostTopologyKey = "kubernetes.io/hostname"
+	timeout         = time.Second * 45
+	hostTopologyKey = "kubernetes.io/hostname"
 )
 
 var (
@@ -223,7 +223,7 @@ var _ = AfterSuite(func() {
 func setupEnvironment() {
 	namespace = "chaos-engineering"
 	testPodImage = "k8s.gcr.io/pause:3.4.1"
-	nodeLabel = map[string]string{HostTopologyKey: "minikube"}
+	nodeLabel = map[string]string{hostTopologyKey: "minikube"}
 }
 
 // podsAreRunning returns true when all the given pods have all their containers running
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -46,6 +46,7 @@ import (
 
 const (
 	timeout = time.Second * 45
+	HostTopologyKey = "kubernetes.io/hostname"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


controllers/suite_test.go
```diff
@@ -99,19 +103,25 @@ var _ = BeforeSuite(func(done Done) {
 		Expect(err).ToNot(HaveOccurred())
 	}()
 
-	instanceKey = types.NamespacedName{Name: "foo", Namespace: "default"}
+	// wait for the cache to sync
+	time.Sleep(10 * time.Second)
```
Devatoria, line 107: What cache are we talking about here?

nikos912000: I was running these tests vs a remote cluster and I was getting an error message along the lines of:
```
the cache is not started, can not read objects
```

Here we create the Kubernetes client. This will watch for and cache objects. I'm assuming in remote clusters this takes some time.

I can drop this tbh as it is not an issue on minikube and may have to do with our clusters. But it took me a while to debug!

Devatoria: Do you recall on which test you were getting this error? I would prefer to have a check in a `Eventually` clause somewhere in the tests setup so we do not blindly wait for 10s but start when we can start reading objects (if that is possible of course).
We will soon run those tests both against minikube and remote clusters too so we'll likely face the same issue.

nikos912000: I can't remember, sorry... It was at the very beginning though, either as part of the setup or when deploying the first CR.
I implemented this as a temporary workaround but I imagine we could use the `waitForCacheSync` method.

nikos912000: Ok, I've reproduced this.

This is because we have more logic in our tests; we run tests for network Availability Zone failures. For this we need to check in which Node/Availability Zone the controller gets deployed at so we do not target that one. We achieve this through `Affinity` rules on the test Pods.

The error message is the following:
```
Unexpected error:
      <*fmt.wrapError | 0xc0008ec540>: {
          msg: "can't list controller pods: the cache is not started, can not read objects",
          err: <*cache.ErrCacheNotStarted | 0x323dbb0>{},
      }
      can't list controller pods: the cache is not started, can not read objects
  occurred
```

I have a feeling this was happening even before adding that logic though. Ultimately, in any calls where we are listing Resources we need to make sure the cache is ready.

I've pushed an update to the PR. I hope that makes more sense.

nikos912000:
@Devatoria do you think these changes would be valuable? I'm happy to close it if not.

ptnapoleon (approved):

PR #529, opened 2022-03-30
[CHAOS-234] Fix flaky cleanup for DNS disruption on restarted containers

nathantournant:
## What does this PR do?

- [ ] Adds new functionality
- [X] Fixes a bug

- Noticed during some tests a DNS disruption on containers that had restarted after injection cleanup was flaky
- After investigation, DNS disruption would not check for cgroup existence prior the write-to-reset attempt, making the injector crash and not consistently clean up.

## Code Quality Checklist

## Testing

    - Applying the tests descibed in the [Jira ticket]()
    - [X] locally.

### DIFF
--- a/injector/dns_disruption.go
+++ b/injector/dns_disruption.go
@@ -156,9 +156,16 @@ func (i DNSDisruptionInjector) Clean() error {
 	}
 
 	if i.config.Level == chaostypes.DisruptionLevelPod {
-		// reset net_cls classid
-		if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0x0"); err != nil {
-			return fmt.Errorf("error writing classid to pod net_cls cgroup: %w", err)
+		// write default classid to pod net_cls cgroup if it still exists
+		exists, err := i.config.Cgroup.Exists("net_cls")
+		if err != nil {
+			return fmt.Errorf("error checking if pod net_cls cgroup still exists: %w", err)
+		}
+
+		if exists {
+			if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0x0"); err != nil {
+				return fmt.Errorf("error reseting classid of pod net_cls cgroup: %w", err)
+			}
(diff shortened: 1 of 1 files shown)


injector/dns_disruption.go
```diff
@@ -156,9 +156,16 @@ func (i DNSDisruptionInjector) Clean() error {
 	}
 
 	if i.config.Level == chaostypes.DisruptionLevelPod {
-		// reset net_cls classid
-		if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0x0"); err != nil {
-			return fmt.Errorf("error writing classid to pod net_cls cgroup: %w", err)
+		// write default classid to pod net_cls cgroup if it still exists
+		exists, err := i.config.Cgroup.Exists("net_cls")
```
ptnapoleon, line 160: Should we add a similar check to the other injectors? I think the problem of "pod fails liveness probe, is restarted, container cgroup files reset" isn't specific to DNS?

nathantournant: It's specific to cgroups access, which is specific to DNS and Network disruptions
I made sure it's checked in both; I failed (8 months ago while writing a DNS bugfix) to copy/paste these exact lines from `network_disruption.go`
I hesitated to integrate/factorise this behavior into a new cgroup class function, and deemed it too specific for this class

nathantournant: Actually it's only consistently checked at cleanup, and consistently not at injection.
The risk we have is the container not existing -- and if a container doesn't exist the injection will fail much earlier

nathantournant: I have to test if the disruption is currently re-injected on container restart following the Dynamic Disruption change, but I don't think it is. Currently there's no mechanic to re-inject anything.
I'll create a low-priority feature ticket for this, this PR just aimed as fixing the cleanup issue.

ptnapoleon (approved):

PR #537, opened 2022-05-09
[CHAOS-266] Send events on target status changes and failures

clairecng:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

tldr: Send events to both target and disruption on status changes and warning / recovered events through the dynamic targeting cache

More detailed:

With the dynamic targeting now operating, we can detect every change in the targets.
With this idea in mind, on every update of any target:
- We find status changes on containers, pods, nodes... To notify of an error status
 - We check for new events to detect the warning events that could be a sign of the pod / container failing. For example, we can't only watch the changes happening in the container statuses, we need to get the events sent to notify of a liveness probe to have the information that the failure comes from a failing liveness probe.
 - We also check for new normal events to make sure we are also notifying of a recovered pod / container failing

After gathering all the new changes on the target, we:
1. Post an event to the target to keep track of what has been notified and say that this failure probably occured because of the disruption
2. Post an event to the disruption to broadcast this event to the user, as we only broadcast disruption events

Sans-titre-2022-05-03-1540

Example of notif in slack:

(CrashLoopBackOff container and recovered event sent)
Screenshot 2022-05-25 at 10 14 18

## Code Quality Checklist

## Testing

    - notified in case of failures of pods / nodes
    - [x] locally.
    - [x] as a deployment to a cluster. (Only pods here)


controllers/cache_handler.go
```diff
@@ -0,0 +1,497 @@
...
+
+		info.AddEventHandler(DisruptionTargetWatcherHandler{disruption: instance, reconciler: r})
+
+		// start the cache with a cancelable context and duration, and attach it to the controller as a watch source
+		ch := make(chan error)
+		cacheCtx, cacheCancelFunc := context.WithTimeout(context.Background(), instance.Spec.Duration.Duration()+*r.ExpiredDisruptionGCDelay*2)
+
+		go func() { ch <- cache.Start(cacheCtx) }()
+
+		r.CacheContextStore[disCacheHash] = CtxTuple{cacheCtx, cacheCancelFunc}
```
Devatoria, line 444: Should we monitor the cache size? I know that this code was already existing before so eventually something to do in another PR.

controllers/cache_handler.go
```diff
@@ -0,0 +1,497 @@
...
+
+				return
+			}
+		}
+
+		h.reconciler.Recorder.Event(h.disruption, eventType, eventReason, chaosv1beta1.Events[eventReason].OnDisruptionTemplateAggMessage)
+	}
+}
+
+func (h DisruptionTargetWatcherHandler) getEventsFromCurrentDisruption(kind string, objectMeta v1.ObjectMeta, disruptionStateTime time.Time) ([]corev1.Event, error) {
```
Devatoria, line 138: We are ok on the tradeoff that if a disruption event happens while the controller is down, this event will be skipped (which is to me better than renotifying about something that happened in the past). Still need to be confirmed by tests.

Devatoria (approved):

controllers/cache_handler.go
clairecng: @nathantournant Did add some metrics regarding the cache, notably: how many caches are still in the cache array, I think?

nathantournant: Currently there's a cache-array gauge metric to make sure the number of caches is never durably superior to the number of ongoing disruptions. It's currently reported [with the recurrent metrics](https://github.com/DataDog/chaos-controller/blob/4eca165f072f00c54ca008f8434dcad51b99ef38/controllers/disruption_controller.go#L1505)
Adding a count on cache addition could help for debugging but I believe a debugging-level log line would be more useful for this
Otherwise I wouldn't exactly know how to monitor an individual cache/memory consumption, but would be happy to work on it with @clairecng

nathantournant (approved):

controllers/cache_handler.go
Devatoria: Oops missed that one, I think we are good for now then!

chart/templates/configmap.yaml
```diff
@@ -23,6 +23,7 @@ data:
         slack:
           enabled: {{ .Values.controller.notifiers.slack.enabled }}
           tokenFilepath: {{ .Values.controller.notifiers.slack.tokenFilepath | quote }}
+          mirrorSlackChannelId: {{ .Values.controller.notifiers.slack.mirrorSlackChannelId | quote }}
```
ptnapoleon, line 26: What do you think of this taking a list of channels, rather than single? Or naming it `notification_channels`?

clairecng: I was more using this as a temporary parameter to broadcast only to our team and remove it when the feature will be deployed, however if you think this would be useful in a more permanent way, yes.

ptnapoleon (approved):

PR #556, opened 2022-07-05
Disruption reinjection on container restart

clairecng:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

### Container Restart

When a container restarts during a disruption targeting pods, the disruption is not re-injected. Because we rely most of the time on container scoped disruptions, the disruption is then not working anymore on this container. In this PR we re-inject the disruption when this occurs by:

1. watching the targeted pod, checking on every modification on the containers to see if any container has changed its ID
2. on ID change, initiating a `clean` of the disruption, an `update` of the configuration of the injector, and an `injection` of the disruption on the whole pod. We don't apply this only to the restarting container as this could potentially create cases where we clean some disruption changes that are not only applied to one container, breaking the disruption on other containers.
3. that's it

Small nits:
- a dedicated metric to re-injection
- adding bookmark on all our watchers to allow for safe error handling

### Fix on init

When using dynamic targeting the targeted pods are dropped on the next reconcile loop when the init container chaos-handler gets terminated. Added a check to not do that and a unit test.

## Code Quality Checklist

## Testing

    - Container restart in a loop
    - [x] locally.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,16 +10,13 @@ run:
     - cpuset/*
 linters:
   enable:
-    - deadcode
     - errcheck
     - gosimple
     - govet
     - ineffassign
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
     - bodyclose
     - depguard
     - dogsled
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -68,8 +68,9 @@ type DisruptionSpec struct {
 	GRPC *GRPCDisruptionSpec `json:"grpc,omitempty"`
 }
 
-//go:embed *
 // EmbeddedChaosAPI includes the library so it can be statically exported to chaosli
+//
+//go:embed *
 var EmbeddedChaosAPI embed.FS
 
 type DisruptionDuration string
--- a/api/v1beta1/groupversion_info.go
+++ b/api/v1beta1/groupversion_info.go
@@ -4,8 +4,8 @@
 // Copyright 2022 Datadog, Inc.
 
 // Package v1beta1 contains API Schema definitions for the chaos v1beta1 API group
-//+kubebuilder:object:generate=true
-//+groupName=chaos.datadoghq.com
+// +kubebuilder:object:generate=true
+// +groupName=chaos.datadoghq.com
 package v1beta1
 
 import (
--- a/chart/templates/configmap.yaml
+++ b/chart/templates/configmap.yaml
@@ -32,6 +32,8 @@ data:
             - {{ . | quote }}
           {{- end }}
           headersFilepath: {{ .Values.controller.notifiers.http.headersFilepath | quote }}
+        datadog:
+          enabled: {{ .Values.controller.notifiers.datadog.enabled }}
       deleteOnly: {{ .Values.controller.deleteOnly }}
       imagePullSecrets: {{ .Values.images.pullSecrets }}
       defaultDuration: {{ .Values.controller.defaultDuration }}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -25,6 +25,8 @@ controller:
       enabled: false
       tokenFilepath: ""
       mirrorSlackChannelId: ""
+    datadog:
+      enabled: false
     http:
       enabled: false
       url: ""
--- a/container/container_mock.go
+++ b/container/container_mock.go
@@ -8,6 +8,7 @@ package container
 import "github.com/stretchr/testify/mock"
 
 // ContainerMock is a mock implementation of the Container interface
+//
 //nolint:golint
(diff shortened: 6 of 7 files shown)


utils/utils.go
```diff
@@ -49,3 +51,40 @@ type SetupWebhookWithManagerConfig struct {
 	HandlerEnabledFlag     bool
 	DefaultDurationFlag    time.Duration
 }
+
+// GetTargetedContainersInfo gets the IDs of the targeted containers or all container IDs found in a Pod
+func GetTargetedContainersInfo(pod *corev1.Pod, targets []string) (map[string]string, error) {
```
clairecng, line 56: Moved the function from controllers/helper to utils in order to use it in the injection logic when reinjection

injector/container_failure.go
```diff
@@ -35,18 +35,18 @@ func NewContainerFailureInjector(spec v1beta1.ContainerFailureSpec, config Conta
 		config.ProcessManager = process.NewManager(config.DryRun)
 	}
 
-	return containerFailureInjector{
+	return &containerFailureInjector{
```
clairecng, line 38: On all injectors:
- added the method UpdateConfig to be able to reset it when the container changes PID and needs to reset the managers
- transformed the injector struct on all methods to a pointer to be able to keep the conf updated when resetting it

api/disruption_kind.go
```diff
@@ -23,7 +24,7 @@ type DisruptionKind interface {
 
 type DisruptionArgs struct {
 	AllowedHosts         []string
-	TargetContainerIDs   []string
+	TargetContainers     map[string]string
```
clairecng, line 27: Changed the logic to transmitting the id of the targeted containers to transmit the names instead: we need to keep the names only so that we can update the ids in case of a container restart

targetselector/running_target_selector.go
```diff
@@ -64,22 +75,25 @@ func (r runningTargetSelector) GetMatchingPodsOverTotalPods(c client.Client, ins
...
 
-		// if the disruption is applied on init, we only target pending pods with a running
+		// if the disruption is applied on init, we only target pending pods with a running (or terminated)
 		// chaos handler init container
 		// otherwise, we only target running pods
 		if instance.Spec.OnInit {
 			hasChaosHandler := false
 
 			// search for a potential running chaos handler init container
 			for _, initContainerStatus := range pod.Status.InitContainerStatuses {
```
clairecng, line 85: With dynamicTargeting, the targeted pod(s) from an on init disruption is removed because the status changes from Pending to Running and the chaos-handler init container is then Terminated.

injector/dns_disruption.go
```diff
@@ -155,8 +155,12 @@ func (i DNSDisruptionInjector) Inject() error {
 	return nil
 }
 
+func (i *DNSDisruptionInjector) UpdateConfig(config Config) {
```
ptnapoleon, line 158: Is the reason all the receivers in the injector package changed to pointers because of updateConfig?

clairecng: Yes, I'm not sure that's the right move

cli/injector/main.go
```diff
@@ -50,7 +55,8 @@ var (
 	ms                   metrics.Sink
 	sink                 string
 	level                string
-	targetContainerIDs   []string
+	tmpTargetContainers  []string
```
ptnapoleon, line 58: Would it be accurate to call this `targetContainerNames`?

ptnapoleon: Oh, no, it's a raw string holding both names and ids?

clairecng: Ah yeah, sorry I was wrong, that's name:id, not sure if it's ideal. I could definitely only transmit the names and get the ids in the injector. What do you think?

ptnapoleon: Oh no, I think the format is great, maybe we could call it `rawTargetContainers` or `initialTargetContainers`? `tmp` was confusing to me, even though they are technically temporary (until/if we need to re-inject)

cli/injector/main.go
```diff
@@ -308,9 +359,58 @@ func inject(kind string, sendToMetrics bool) bool {
 	return !errOnInject
 }
 
+// reinject reinitialize conf, clean and inject all the disruptions
+func reinject(pod *v1.Pod, cmdName string) error {
+	var err error
+
+	// Clean all injections to reinject on an empty slate
+	if ok := clean(cmdName, true, true); !ok {
```
ptnapoleon, line 367: Why do we need to clean? Do all disruption kinds need to clean? I would think we could just edit the cgroup files for the new container id, what complications am I overlooking?

ptnapoleon (approved):

ptnapoleon (approved):

PR #563, opened 2022-07-21
Limit Safemode List Digest

Azoam:
## What does this PR do?

- [ ] Adds new functionality
- [X] Fixes a bug

- We found that the safemode feature which checks cluster information to determine blast radius was taking up a lot of memory and OOMing pods. This PR is a potential fix to this, by limiting how many objects are digested at a time, hopefully not filling the memory up.

In regards the branch name consisting of etcd - This was the original idea - to use etcd for the object count look up, but this fell through and now we are using the limiting functionality of the client api.

## Code Quality Checklist

## Testing
 - From the initial sight of this bug, the injector pods were OOMing/timing out. I tested this branch in an environment with 10k+ pods and no OOMing or Timing out was occurring.


api/v1beta1/safemode.go
```diff
@@ -23,10 +23,10 @@ type Config struct {
 // CountTooLargeConfig represents the configuration for the countTooLarge safetynet
 // +ddmark:validation:AtLeastOneOf={NamespaceThreshold,ClusterThreshold}
 type CountTooLargeConfig struct {
-	// +ddmark:validation:Minimum=0
+	// +ddmark:validation:Minimum=1
 	// +ddmark:validation:Maximum=100
 	NamespaceThreshold int `json:"namespaceThreshold,omitempty"`
```
ptnapoleon, line 28: Can we combine a minimum of 1 and `omitempty` in this way?

ptnapoleon (approved):
I had hoped there was a way to query the API for just the total count, not a full list, but seems thats lacking

Azoam:
> I had hoped there was a way to query the API for just the total count, not a full list, but seems thats lacking

From my understanding and research, there is a way to query, but it requires specific configuration on a kubernetes environment which everyone may not have. Also those who do have it configured may have it locked up behind administrator permissions. So it seems like a safe bet to use the generic client that should be generally accessible.

api/v1beta1/safemode.go
Azoam: Not that I could find.

Azoam: Added a new commit which better combines the two.

PR #576, opened 2022-08-22
bugfix: cache deletion delay safe call to ExpiredDisruptionDelay

nathantournant:
## What does this PR do?

- [ ] Adds new functionality
- [X] Fixes a bug

- If ExpiredDisruptionGCDelay isn't fixed, the dynamic targeting observation cache crashes out the controller calling a non-existing pointer value. This sets a 2 minutes default deletion value and safely attempts to call the ExpiredDisruptionGCDelay value.

## Code Quality Checklist

## Testing

    - [ ] 
    - [ ] locally.


clairecng (approved):
LGTM
Should we add a small documentation line that says when using dynamic targeting, disruption  dynamic targeting expires after 2 minutes if the expiredDisruptionGCDelay is not set? (If I understood this line correctly)

controllers/cache_handler.go
```diff
@@ -511,7 +511,13 @@ func (r *DisruptionReconciler) manageInstanceSelectorCache(instance *chaosv1beta
 
 		// start the cache with a cancelable context and duration, and attach it to the controller as a watch source
 		ch := make(chan error)
-		cacheCtx, cacheCancelFunc := context.WithTimeout(context.Background(), instance.Spec.Duration.Duration()+*r.ExpiredDisruptionGCDelay*2)
+		deletionDelay := time.Minute * 2
+
+		if r.ExpiredDisruptionGCDelay != nil {
```
expFlower, line 516: If `ExpiredDisruptioonGCDelay` is not set and it [defaults to `-1`](https://github.com/DataDog/chaos-controller/blob/d405aa7935475f82b8896151eb338f12b6f1b11a/main.go#L152-L153) will this have the desired effect?

controllers/cache_handler.go
```diff
@@ -511,7 +511,13 @@ func (r *DisruptionReconciler) manageInstanceSelectorCache(instance *chaosv1beta
 
 		// start the cache with a cancelable context and duration, and attach it to the controller as a watch source
 		ch := make(chan error)
-		cacheCtx, cacheCancelFunc := context.WithTimeout(context.Background(), instance.Spec.Duration.Duration()+*r.ExpiredDisruptionGCDelay*2)
+		deletionDelay := time.Minute * 2
```
Devatoria, line 514: This default value seems risky as a disruption can have no GC expiration delay and last longer than a couple of minutes. In this case, it would clear the cache for a still existing disruption.

Devatoria (changes requested):

controllers/cache_handler.go
nathantournant: yes, the value is set to an empty pointer [there](https://github.com/DataDog/chaos-controller/blob/d405aa7935475f82b8896151eb338f12b6f1b11a/main.go#L354). it's what created the segfault

controllers/cache_handler.go
nathantournant: Agreed. It would then recreate another cache with the disruption duration, and so on until the disruption is deleted. This is not a suitable solution.

ptnapoleon (approved):
lgtm, let's test in staging to confirm the new polling doesn't have too big a perf impact?

PR #577, opened 2022-08-22
staticTargeting: add flag to switch default to true

nathantournant:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Now places the default value for staticTargeting to a flag in the controller's config map.

## Code Quality Checklist

## Testing

    - Deploy the controller with the `staticTargetingDefault` field set to false, launch a disruption with `staticTargeting` unspecified, check out the description
    - Do all of those combinations (flag false/true, staticTargeting field unspecified/false/true) and assert you have the expected results.
    - [X] locally.

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -68,8 +68,9 @@ type DisruptionSpec struct {
 	GRPC *GRPCDisruptionSpec `json:"grpc,omitempty"`
 }
 
-//go:embed *
 // EmbeddedChaosAPI includes the library so it can be statically exported to chaosli
+//
+//go:embed *
 var EmbeddedChaosAPI embed.FS
 
 type DisruptionDuration string
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -44,7 +44,7 @@ type DisruptionSpec struct {
 	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
 	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
 	Unsafemode       *UnsafemodeSpec                   `json:"unsafeMode,omitempty"`       // unsafemode spec used to turn off safemode safety nets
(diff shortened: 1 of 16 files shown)


api/v1beta1/disruption_types.go
```diff
@@ -44,7 +44,7 @@ type DisruptionSpec struct {
 	DryRun           bool                              `json:"dryRun,omitempty"`           // enable dry-run mode
 	OnInit           bool                              `json:"onInit,omitempty"`           // enable disruption on init
 	Unsafemode       *UnsafemodeSpec                   `json:"unsafeMode,omitempty"`       // unsafemode spec used to turn off safemode safety nets
-	StaticTargeting  bool                              `json:"staticTargeting,omitempty"`  // enable dynamic targeting and cluster observation
+	StaticTargeting  *bool                             `json:"staticTargeting,omitempty"`  // disable dynamic targeting and cluster observation
```
ptnapoleon, line 47: Why did this have to become a pointer again? The default for a `bool` is already `false,` no?

nathantournant: This PR aims to allow `true` to be a possible default value based on the chart
This value, if set, needs to not be overridden by an empty one

ptnapoleon: I see. There is another option, I think? At the cost of some wordier unmarshaling, we can avoid making this be a pointer everywhere.

If we make the type of `StaticTargeting` not `bool`, but a custom type: `type StaticTargetingFlag bool`, we can add a custom Unmarshal for that type that uses the chart-defined default when empty.

nathantournant: I tried a few things, hoping this would work; without a new dedicated struct type for the field -which would not really help- the UnmarshalJSON function is not called on an empty value 😞 
I'm afraid the pointer is unavoidable there

ptnapoleon: Is it..... too late to change this field to "Targeting" with the options of "dynamic,static,empty", where empty uses the default? Representing three states by using a pointer to a bool feels wrong

PR #583, opened 2022-08-29
Cloud Managed Service Disruption

clairecng:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

### Context

Large cloud services providers are using wide IP ranges. Hostnames used to identify those services are resolving with some IPs of that range, and resolved IPs can change between each DNS request. Applying a network disruption using those hostnames only doesn’t work well since retrying the resolution of such hostname would return new IPs (not disrupted) and the disruption would be ineffective. 
We should add a new option to the network disruption to target specific cloud services providers. Each provider hosts a JSON file with the list of all IP ranges. The format differs between each provider but all of them contain information about the IP ranges and the associated service(s).

In this PR, we're creating:
- the logic behind the pull and parse of those json files
- the logic behind transforming the new disruption into a hosts disruption
- a first cloud provider: AWS.

### How

1. On startup, we get the ip range file for each cloud provider. We parse it and keep in memory those ip ranges per service cloud service.
2. On disruption creation, we verify the user is creating a valid cloud managed service disruption and we transform the cloud disruption to create a hosts disruption, where we put all ip ranges as hosts.

### Spec

Similar to a network disruption, a new field has been added: `cloud` field.

```
cloud:
  aws: # cloud provider name. For now, we can only support AWS
    - service: "S3" ## the list of services we want to affect
    - service: "EC2"
    - ...
```

Full disruption example:
```
kind: Disruption
metadata:
  name: network-cloud
  namespace: chaos-demo
spec:
  level: pod
  selector:
    app: demo-cirl
  count: 1
  network:
    cloud:
      aws:
        - service: "S3"
          flow: "egress" # optional
          protocol: "tcp" # optional
    delay: 1000 # delay (in milliseconds) to add to outgoing packets, 10% of jitter will be added by default
    delayJitter: 5 # (optional) add X % (1-100) of delay as jitter to delay (+- X% ms to original delay), defaults to 10%
```

## Code Quality Checklist

## Testing

    - `aws cloud managed disruption with 1 service / with too much services `
    - [x] locally.


api/v1beta1/disruption_webhook.go
```diff
@@ -99,6 +104,22 @@ func (r *Disruption) ValidateCreate() error {
...
+
+		for cloudName, serviceList := range clouds {
+			for _, service := range serviceList {
+				if !cloudProviderManager.ServiceExists(cloudName, service) {
+					return fmt.Errorf("service %s of %s does not exist. Available are: %s", service, cloudName, strings.Join(cloudProviderManager.GetServiceList(cloudName), ", "))
+				}
+			}
+		}
+	}
+
```
clairecng, line 122: Added this in disruption_webhook.go instead of the specific Validate() method of the NetworkDisruptionSpec [here](https://github.com/DataDog/chaos-controller/blob/d934f1534a70126846d2f81493147e3777bce6c8/api/v1beta1/network_disruption.go#L104) due to a problem with the package deepcopy-gen. It's preventing us from passing the cloudProviderManager to those methods because we can't generate the DeepCopyInto method on interfaces.
In my opinion, we should put this in Validate() but I'm not sure the time wasted trying to find a workaround is worh it.

controllers/helpers.go
```diff
@@ -126,6 +129,50 @@ func validateLabelSelector(selector labels.Selector) error {
...
+				Host:     ip,
+				Protocol: "tcp",
+				Flow:     "egress",
+			})
+		}
+	}
+
+	return hosts
+}
+
```
clairecng, line 175: Same problem as the comment above:

Added this in helpers.go instead of the specific GenerateArgs() method of the NetworkDisruptionSpec [here](https://github.com/DataDog/chaos-controller/blob/d934f1534a70126846d2f81493147e3777bce6c8/api/v1beta1/network_disruption.go#L136) due to a problem with the package deepcopy-gen. It's preventing us from passing the cloudProviderManager to those methods because we can't generate the DeepCopyInto method on interfaces.
In my opinion, we should put this in GenerateArgs() but I'm not sure the time wasted trying to find a workaround is worth it.

controllers/helpers.go
```diff
@@ -126,6 +129,50 @@ func validateLabelSelector(selector labels.Selector) error {
...
+	for cloudName, serviceList := range clouds {
+		var ips []string
+
+		var err error
+
+		if len(serviceList) == 0 {
+			ips, err = cloudManager.GetServiceIPRanges(cloudName, "")
+			if err != nil {
+				log.Errorf("could not retrieve the ip range of all services of %s", cloudName)
```
ptnapoleon, line 149: We need to log the error message here and below. The only errors I can possibly see coming from this function are "cloud provider doesn't exist" and "service name doesn't exist", both of which should already have been caught via validation before we get to this point. So these might be serious errors

main.go
```diff
@@ -245,6 +249,9 @@ func main() {
 		"Threshold which safemode checks against to see if the number of targets is over safety measures within a cluster")
 	handleFatalError(viper.BindPFlag("controller.safemode.clusterThreshold", pflag.Lookup("safemode-cluster-threshold")))
 
+	pflag.StringVar(&cfg.Controller.CloudProviders.Aws.IPRangesURL, "cloud-providers-aws-ip-ranges-url", "", "URL used to pull ip ranges from AWS (defaulted to \"\")")
```
ptnapoleon, line 252: We default to an empty string, but won't we error out in our Cloud Service if that happens? How do we want to handle a situation where we don't get a config (or if the config we do have doesn't work)? Fail all disruptions that try to block to aws?

clairecng: Changed this part to not have any configuration as the URL similar for everyone.
For now, we don't start the controller if an error is thrown at the parsing of the ip ranges.

api/v1beta1/network_disruption.go
```diff
@@ -89,6 +91,15 @@ type NetworkDisruptionServiceSpec struct {
 	Namespace string `json:"namespace"`
 }
 
+type NetworkDisruptionCloudSpec struct {
+	// +nullable
+	AWS []string `json:"aws,omitempty"`
```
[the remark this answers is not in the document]
clairecng: Yes, renamed the field to be clearer

controllers/helpers.go
clairecng: Chose to log it in the createChaosPods func instead, wdyt?

cloudservice/aws/aws_test.go
```diff
@@ -0,0 +1,32 @@
...
+// Copyright 2022 Datadog, Inc.
+
+package aws
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Label Selector Validation", func() {
```
Devatoria, line 13: I think this describe text is wrong

cloudservice/aws/aws_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+		It("should parse the ip range file", func() {
+			By("Ensuring that no error was thrown")
+			Expect(err).To(BeNil())
+
+			By("Ensuring that the right version string was parsed")
+			Expect(info.Version).To(Equal("1000000000"))
+
+			By("Ensuring that we have the right info")
+			Expect(len(info.IPRanges["AMAZON"])).To(Equal(2))
```
Devatoria, line 28: I thought the `AMAZON` service would be pruned from the parsed ranges but we check it exists here. Did I understand it wrong?

injector/network_disruption.go
```diff
@@ -30,7 +30,7 @@ import (
 type linkOperation func([]string, string, uint32) error
 
 // tcPriority the lowest priority set by tc automatically when adding a tc filter
-var tcPriority = uint32(49149)
+var tcPriority = uint32(1000)
```
Devatoria, line 33: Do we know where the old value comes from? Just to be sure it has no side effect to start from 1000 now.

injector/network_disruption.go
```diff
@@ -369,7 +369,12 @@ func (i *networkDisruptionInjector) applyOperations() error {
 	if len(i.spec.Hosts) == 0 && len(i.spec.Services) == 0 {
 		_, nullIP, _ := net.ParseCIDR("0.0.0.0/0")
 
-		if err := i.config.TrafficController.AddFilter(interfaces, "1:0", i.getNewPriority(), 0, nil, nullIP, 0, 0, "", "1:4"); err != nil {
+		prio, err := i.getNewPriority()
```
Devatoria, line 372: I'm a bit annoyed by all those `getNewPriority()` calls now done outside of the `AddFilter` function because of the returned error. It duplicates those 4 LOC a bit everywhere. Should this logic be included into the `AddFilter` function itself since it is totally tc related?

Devatoria (changes requested):

injector/network_disruption.go
clairecng: the old value comes from the automatically set priority. I took it because of this, but it's totally safe to use this instead

cloudservice/aws/aws_test.go
clairecng: Yes, sorry about that. Corrected the tests.

controllers/helpers.go
```diff
@@ -126,6 +129,31 @@ func validateLabelSelector(selector labels.Selector) error {
...
+		ipRanges, err := cloudManager.GetServiceIPRanges(cloudName, *serviceList)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, ip := range ipRanges {
+			hosts = append(hosts, v1beta1.NetworkDisruptionHostSpec{
+				Host:     ip,
+				Protocol: "tcp",
+				Flow:     "egress",
```
Azoam, line 149: I may be wrong, but I think there are some load balancers and services from AWS that accept udp packets. Also is there really no case that someone may want an ingress disruption instead?

Azoam (approved):

chart/templates/configmap.yaml
```diff
@@ -34,6 +34,8 @@ data:
           headersFilepath: {{ .Values.controller.notifiers.http.headersFilepath | quote }}
         datadog:
           enabled: {{ .Values.controller.notifiers.datadog.enabled }}
+      cloudproviders:
```
Devatoria, line 37:
```suggestion
      cloudProviders:
```

chart/values.yaml
```diff
@@ -32,6 +32,8 @@ controller:
       url: ""
       headersFilepath: ""
       headers: []
+  cloudproviders:
```
Devatoria, line 35:
```suggestion
  cloudProviders:
```

Devatoria (changes requested):

controllers/helpers.go
clairecng: Thanks for catching that!

ptnapoleon (approved):

PR #585, opened 2022-09-12
feat(webhook): prevent finalizer removal

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- Chaos controller guarantee `disruption` is not deleted until associated chaos pods are 
- Chaos controller guarantee finalizer is added as long as a `disruption` is NOT deleted [here](https://github.com/DataDog/chaos-controller/blob/main/controllers/disruption_controller.go#L209)
- However, we did not prevent `disruption` finalizer removal AFTER a `disruption` was marked for deletion
- Hence, a user seeing a `disruption` blocked could be tempted to remove the finalizer to force/speed up the `disruption` removal, leading to all associated chaos pods being left un-supervised in the cluster in their statuses, forever
- The proposed implementation blocks such removal and list associated chaos pods waiting for pending action to enable/unblock `disruption` deletion

During testing I noticed the following:
* we were not using `-race` flag so I added it and fixed associated data race detected [here](https://github.com/DataDog/chaos-controller/pull/585/files#diff-71cea628d383a9b3f0e1a950238cd4543ec0994beb01e9201b3f87c2815fda30R143) and [here](https://github.com/DataDog/chaos-controller/pull/585/files#diff-d3922ac8e373158d56a2ce016a6bfc8f97d576f2f0149d971037685d1d4a0cb9R75)
* running e2e-tests locally failed for 1 test due to the `expiredDisruptionGCDelay` being `10m` and testing requiring full cleanup 
  * I suggested another implementation to be able to have all e2e-tests successful locally, however, the drawback is it COULD lead to local disruptions being deleted earlier than expected (as the GC delay is a configuration change requiring a new chaos controller deploy)
  * Let me know if you believe we should keep the old way of doing this

## Code Quality Checklist

## Testing

    - [x] locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,8 @@ colima-all:
 	$(MAKE) colima-build
 	$(MAKE) colima-install
 
+colima-deploy: colima-build colima-install colima-restart
+
 install-cert-manager:
 	$(KUBECTL) apply -f https://github.com/jetstack/cert-manager/releases/download/v1.9.1/cert-manager.yaml
 
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ all: manager injector handler
 
 # Run unit tests
 test: generate manifests
-	go test $(shell go list ./... | grep -v chaos-controller/controllers) -coverprofile cover.out
+	go test -race $(shell go list ./... | grep -v chaos-controller/controllers) -coverprofile cover.out
 
 # This target is dedicated for CI and aims to reuse the Kubernetes version defined here as the source of truth
 ci-install-minikube:
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,10 @@ KUBERNETES_VERSION ?= v1.22.13
 
 CLUSTER_NAME ?= colima
 
+# expired disruption gc delay enable to speed up chaos controller disruption removal for e2e testing
+# it's used to check if disruptions are deleted as expected as soon as the expiration delay occurs
+EXPIRED_DISRUPTION_GC_DELAY ?= 10m
+
 OS_ARCH=amd64
 ifeq (arm64,$(shell uname -m))
 OS_ARCH=arm64
@@ -42,8 +46,9 @@ ci-install-minikube:
 	minikube status
 
 # Run e2e tests (against a real cluster)
-e2e-test: generate colima-install
-	USE_EXISTING_CLUSTER=true CLUSTER_NAME=${CLUSTER_NAME} go test ./controllers/... -coverprofile cover.out
+e2e-test: generate
+	$(MAKE) colima-install EXPIRED_DISRUPTION_GC_DELAY=10s
+	USE_EXISTING_CLUSTER=true CLUSTER_NAME=${CLUSTER_NAME} go test -race ./controllers/... -coverprofile cover.out
 
 # Build manager binary
 manager: generate
@@ -105,6 +110,7 @@ install-longhorn:
 colima-install: manifests
 	helm template \
 		--set controller.enableSafeguards=false \
+		--set controller.expiredDisruptionGCDelay=${EXPIRED_DISRUPTION_GC_DELAY} \
 		./chart | $(KUBECTL) apply -f -
 
 # Uninstall CRDs and controller from a colima k3s cluster
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/DataDog/chaos-controller/ddmark"
@@ -27,6 +28,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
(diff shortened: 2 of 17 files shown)


Makefile
```diff
@@ -65,6 +68,10 @@ colima-all:
 	$(MAKE) colima-build
 	$(MAKE) colima-install
 
+colima-redo:
```
ptnapoleon, line 71: What does this reset?

controllers/disruption_controller.go
```diff
@@ -140,7 +140,7 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	// check whether the object is being deleted or not
-	if !instance.ObjectMeta.DeletionTimestamp.IsZero() {
+	if !instance.DeletionTimestamp.IsZero() {
```
ptnapoleon, line 143: Afaict, `instance.DeletionTimestamp` is the exact same as `instance.ObjectMeta.DeletionTimestamp`, yes?

api/v1beta1/disruption_webhook.go
```diff
@@ -138,15 +142,45 @@ func (r *Disruption) ValidateCreate() error {
...
+
+	var err error
+
+	oldDisruption := old.(*Disruption)
+
+	// ensure finalizer removal is only allowed if no related chaos pods exists
+	// such verification is only needed if disruption is currently in removal as controller will always add finalizer to disruptions not in removal during the reconcile loop
+	// we should NOT always prevent finalizer removal because chaos controller reconcile loop will go through this mutating webhook when perfoming updates
+	// and need to be able to remove the finalizer to enable the disruption to be garbage collected on successful removal
+	if !oldDisruption.ObjectMeta.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(oldDisruption, chaostypes.DisruptionFinalizer) && !controllerutil.ContainsFinalizer(r, chaostypes.DisruptionFinalizer) {
```
ptnapoleon, line 156: Why use `oldDisruption.ObjectMeta.DeletionTimestamp` here instead of `oldDisruption.DeletionTimestamp` given similar changes elsewhere?

api/v1beta1/disruption_webhook.go
```diff
@@ -138,15 +142,45 @@ func (r *Disruption) ValidateCreate() error {
...
-	logger.Debugw("validating updated disruption", "instance", r.Name, "namespace", r.Namespace)
+	logger := logger.With("instance", r.Name, "namespace", r.Namespace)
+	logger.Debugw("validating updated disruption")
+
+	var err error
+
+	oldDisruption := old.(*Disruption)
+
+	// ensure finalizer removal is only allowed if no related chaos pods exists
+	// such verification is only needed if disruption is currently in removal as controller will always add finalizer to disruptions not in removal during the reconcile loop
```
ptnapoleon, line 153: Is this resilient to a user first removing the finalizer, then deleting the disruption? Are we guaranteed the controller will reconcile between the two as a response to the update that removes the finalizer?

api/v1beta1/v1beta1_suite_test.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestV1Beta1(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "v1beta1 Suite")
+}
+
+var _ = BeforeSuite(func() {
+	By("initiliazing disruption_webhook global variables")
```
ptnapoleon, line 25:
```suggestion
	By("initializing disruption_webhook global variables")
```

Makefile
luphaz: I changed the name to `colima-deploy` that is doing build/install/restart.

If we believe it's not relevant I will remove it, let me know.

controllers/disruption_controller.go
ptnapoleon: We don't favor one or the other, afaik. I was more curious if you knew a reason that the shorter form was preferable. If they're identical but one is more concise, I'm definitely in support of the shorter option

api/v1beta1/disruption_webhook.go
ptnapoleon: I didn't think it through enough, I rescind my worries

controllers/disruption_controller.go
luphaz: Not really, I favoured it because it was the shorter option indeed.
The main difference is the short option due to go, is that we can have a meaning that changes:
* as long as there is no field in Disruption called `DeletionTimestamp` it's linking to `ObjectMeta.DeletionTimestamp`
* if at any time we have a second embedded struct containing this field, it will lead to a compile time error 
```text
ambiguous selector instance.DeletionTimestampcompiler[AmbiguousSelector](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal?utm_source%3Dgopls#AmbiguousSelector)
```
* if we create the same field `DeletionTimestamp` we would have a compile error due to missing method

The only risk in using the shortest would hence be if we introduce the exact same field with the exact same type. I would say it's a rather edge case we might not be interested in preventing.

Makefile
```diff
@@ -29,11 +32,16 @@ all: manager injector handler
...
 # Run unit tests
 test: generate manifests
-	go test $(shell go list ./... | grep -v chaos-controller/controllers) -coverprofile cover.out
+	go test -race $(shell go list ./... | grep -v chaos-controller/controllers) -coverprofile cover.out
 
 # Run e2e tests (against a real cluster)
-e2e-test: generate minikube-install
-	USE_EXISTING_CLUSTER=true go test ./controllers/... -coverprofile cover.out
+e2e-test: generate
+	$(MAKE) minikube-install EXPIRED_DISRUPTION_GC_DELAY=10s
```
ptnapoleon, line 39: wow, so fancy!

ptnapoleon (approved):

ptnapoleon (approved):

PR #597, opened 2022-10-05
Change the reinject loop

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Fix a potential livelock issue. The reinject loop listens for sigterms and leaves the loop. However when we return to our caller, we'll be greeted by a blocking call waiting for a sigterm. But, the reinject loop already consumed the sigterm from that channel! So we'll end up blocking until the controller or human retries sending the sigterm. That's not a great behavior, so instead slightly modify the behavior. We still return to our caller, but we also put the sigterm back into the channel.
- Handle expired resource versions during reinject. Previously we were relying on the BOOKMARK event to let us know when a new resource version was available. But sometimes we'd get explicitly told our version was expired. In these instances we'd just exit. However, there is an explicit status code for this situation. Now we check for that, and reset our resource version, and re-create the channel

## Code Quality Checklist

## Testing

    - [ ] locally.

### DIFF
--- a/cli/injector/main.go
+++ b/cli/injector/main.go
@@ -677,6 +677,7 @@ out:
 
 			if event.Type == watch.Bookmark {
 				resourceVersion = pod.ResourceVersion
+				log.Debugw("received bookmark event, new resource version found", "resourceVersion", pod.ResourceVersion)
 			}
 
 			if event.Type != watch.Modified {
--- a/cli/injector/main.go
+++ b/cli/injector/main.go
@@ -637,6 +637,7 @@ func watchTargetAndReinject(deadline time.Time, commandName string, pulseActiveD
 
 			pod, ok := event.Object.(*v1.Pod)
 			if !ok {
+				log.Debugw("received event was not a pod", "event", event, "event.Object", event.Object)
 				return fmt.Errorf("watched object received from event is not a pod")
 			}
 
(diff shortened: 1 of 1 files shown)


cli/injector/main.go
```diff
@@ -630,16 +633,49 @@ func watchTargetAndReinject(deadline time.Time, commandName string, pulseActiveD
 			if !ok {
 				channel = nil
 
-				break
+				continue
```
ptnapoleon, line 636: This is mostly a no-op, but I think it tracks what we want to do more accurately. It isn't that we want to leave the `select`, but that we want to return to the top of the `for`, to re-create the channel

cli/injector/main.go
```diff
@@ -608,6 +608,9 @@ func watchTargetAndReinject(deadline time.Time, commandName string, pulseActiveD
 		select {
 		case sig := <-signals:
 			log.Infow("an exit signal has been received", "signal", sig.String())
+			// We've already consumed the signal from the channel, so our caller won't find it when checking after we return.
+			// Thus we need to put the signal back into the channel. We do it in a gothread in case we are blocked when writing to the channel
+			go func() { signals <- sig }()
```
taihuynh167, line 613: would love to see a unit test for this.

clairecng (approved):

cli/injector/main.go
ptnapoleon: I don't have a practical way to mock things such that we can test this literal line of code. In terms of testing this functionality of channels to prevent blocking in our caller, I've proven it out in the golang playground:
https://go.dev/play/p/UzCeUb2eh3C

PR #601, opened 2022-10-10
feat(eventnotifier/slack): extend disruption info

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- This PR aims to add `DryRun/Duration/Author` info to slack message sent

Screenshot of a local triggered notification inside Colima:

image

## Code Quality Checklist

## Testing

    - [x] locally.
    - [ ] as a canary deployment to a cluster. _maybe soon, does not want to have yet another forked version running in the wild for now_


api/v1beta1/disruption_webhook.go
```diff
@@ -245,18 +243,15 @@ func (r *Disruption) getMetricsTags() []string {
...
-		var annotation v1.UserInfo
-
-		err := json.Unmarshal([]byte(r.Annotations["UserInfo"]), &annotation)
+	if userInfo, err := r.UserInfo(); !errors.Is(err, ErrNoUserInfo) {
 		if err != nil {
-			logger.Errorw("Error decoding annotation", err)
+			logger.Errorw("error retrieving user info from disruption, using empty user info", "error", err, "name", r.Name, "namespace", r.Namespace)
 		}
 
-		tags = append(tags, "username:"+annotation.Username)
```
clairecng, line 256: I wonder if we should change the tag name to disruptionUser?

webhook/user_info.go
```diff
@@ -55,15 +55,13 @@ func (m *UserInfoMutator) Handle(ctx context.Context, req admission.Request) adm
 		annotations[k] = v
 	}
 
-	marshaledUserInfo, err := json.Marshal(req.UserInfo)
+	dis.Annotations = annotations
+
+	err := dis.SetUserInfo(req.UserInfo)
 	if err != nil {
-		m.Log.Errorw("error encoding UserInfo", "error", err)
+		m.Log.Errorw("error defining UserInfo", "error", err, "name", dis.Name, "namespace", dis.Namespace)
```
clairecng, line 62: small nit: disruptionName and disruptionNamespace?

clairecng:
Could you add a screenshot of the new slack message to the description of the PR? I'm just curious! 😸

webhook/user_info.go
Devatoria: Good catch! Definitely worth changing that to keep the same everywhere + avoid collision with this kind of generic tag names.

clairecng (approved):
Looks great, thanks for the tests!! 👍

webhook/user_info.go
luphaz: It's not consistent with other log usage in the chaos controller 😬 
I'm fine aligning this TBH, just wanted to mention we changed this due to metrics tagging, not logs tagging.

If we want to be consistent everywhere I can adapt accordingly 🤷 

Not sure if the alignment would risk to break anything (saved view or log filtering by disruption name?) for us I don't think so.

WDYT?

clairecng: IMO we should use disruptionName and disruptionNamespace everywhere, and make it consistent. We could do this in a separate PR where we review all our tags instead if you think it has not its place here?

api/v1beta1/disruption_webhook.go
luphaz: This tag seems as of today, purely related to our metric, and I do not see any Datadog added tag resembling to this, not sure it worth to make it prefixed for this specific use case?

Looking at available data since 3 months and `username` was always filled by our values:
* https://ddstaging.datadoghq.com/metric/explorer?start=1657458970377&end=1665407770377&paused=false#N4IgNghgng9grgFxALlADxQRgDQil3AdwEsATBACwJAoFNiBzCpZTAX11NoDNiA7YgmIw+KUAigAHWihBCAtrQDOtAE7FlIXKtoBHOMoRKUAbVA6lkkSoD63GKvkQWc4opXrNufWo3HkZiB8EIqyPqpQmFogpM4QNkrwqgDGMsggigjqycbeBhGyEABuDMjJFBAwSgB0ySJZMGBgatVFEGBkzsJ8tTrOtKTAAFRsAAQARlCjwHAewYps1RBKNnVwfAgAFACUIGwAurj2jnCQ-oHH8qcQYfmRe4cgShLNYiCS7bQICGkxMAyrM5KYjJaIdPi0GwSaSyRIdUhg-iQkjkKjpPgOJxgPacYiWSBQKFSX7jCCqYwHNhsIA

We isolated `name` to `disruptionName` in metrics because `name` was an already defined tag by `host` I believe

webhook/user_info.go
luphaz: I would at least prefer to have the opportunity to discuss this with @ptnapoleon because he did the first round of tags alignment.

I'm fine to either decision we make, I just want to be sure we first agree on which convention to choose 😅

luphaz:
> Could you add a screenshot of the new slack message to the description of the PR? I'm just curious! 😸

Screenshot added 🤗

taihuynh167 (approved):
nice work <3

PR #602, opened 2022-10-12
Fix bug not count all distinct cores when applying cpu-pressure.

taihuynh167:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- The `injector` currently inspects each container cpuset to determine how many cores it should stress. 2 containers may have different subset of cores and we're supposed to stress all of the involved cores.

## Code Quality Checklist

## Testing

    - Check that the new CPUSet which is created by appending all the cores from each container has the correct value
    - [x] locally.

---------------
some interesting local log lines
```
{"level":"info","ts":1666113867823.0247,"caller":"injector/cpu_pressure.go:161","message":"target identified to be running on 4 cores","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima","cores":[0,1,2,3]}
{"level":"info","ts":1666113867823.0354,"caller":"injector/cpu_pressure.go:168","message":"changed GOMAXPROCS value from 4 to 4","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima"}
{"level":"info","ts":1666113867905.8123,"caller":"injector/cpu_pressure.go:113","message":"all routines have been created successfully, now stressing","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima","StresserPIDPerCore":{"0":10100,"1":10119,"2":10118,"3":10113}}

{"level":"info","ts":1666113867926.2175,"caller":"injector/cpu_pressure.go:161","message":"target identified to be running on 4 cores","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima","cores":[0,1,2,3]}
{"level":"info","ts":1666113867926.2415,"caller":"injector/cpu_pressure.go:168","message":"changed GOMAXPROCS value from 4 to 4","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima"}
{"level":"info","ts":1666113867926.2505,"caller":"injector/cpu_pressure.go:78","message":"core 0 is already stressed, skipping","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima"}
{"level":"info","ts":1666113867926.2822,"caller":"injector/cpu_pressure.go:78","message":"core 1 is already stressed, skipping","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima"}
{"level":"info","ts":1666113867926.3054,"caller":"injector/cpu_pressure.go:78","message":"core 2 is already stressed, skipping","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima"}
{"level":"info","ts":1666113867926.3184,"caller":"injector/cpu_pressure.go:78","message":"core 3 is already stressed, skipping","disruptionName":"cpu-pressure","disruptionNamespace":"chaos-demo","targetName":"demo-curl-79bf7b74c9-tchgv","targetNodeName":"colima"}
```

### DIFF
--- a/cli/injector/cpu_pressure.go
+++ b/cli/injector/cpu_pressure.go
@@ -8,6 +8,7 @@ package main
 import (
 	"github.com/DataDog/chaos-controller/api/v1beta1"
 	"github.com/DataDog/chaos-controller/injector"
+	"github.com/DataDog/chaos-controller/stress"
 	"github.com/spf13/cobra"
 )
 
@@ -18,10 +19,15 @@ var cpuPressureCmd = &cobra.Command{
 	PreRun: func(cmd *cobra.Command, args []string) {
 		// prepare spec
 		spec := v1beta1.CPUPressureSpec{}
+		stresserManager := stress.NewCPUStresserManager(log)
 
 		// create injector
 		for _, config := range configs {
-			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config}))
+			injector, _ := injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{
+				Config:          config,
+				StresserManager: stresserManager,
+			})
+			injectors = append(injectors, injector)
 		}
 	},
 }
--- a/injector/cpu_pressure.go
+++ b/injector/cpu_pressure.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 
 	"github.com/DataDog/chaos-controller/api/v1beta1"
-	"github.com/DataDog/chaos-controller/cpuset"
 	"github.com/DataDog/chaos-controller/process"
 	"github.com/DataDog/chaos-controller/stress"
 	"github.com/DataDog/chaos-controller/types"
@@ -26,13 +25,14 @@ type cpuPressureInjector struct {
 // CPUPressureInjectorConfig is the CPU pressure injector config
 type CPUPressureInjectorConfig struct {
 	Config
-	Stresser       stress.Stresser
-	StresserExit   chan struct{}
-	ProcessManager process.Manager
+	Stresser        stress.Stresser
+	StresserExit    chan struct{}
+	ProcessManager  process.Manager
+	StresserManager stress.StresserManager
 }
 
 // NewCPUPressureInjector creates a CPU pressure injector with the given config
-func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInjectorConfig) Injector {
+func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInjectorConfig) (Injector, error) {
 	// create stresser
 	if config.Stresser == nil {
 		config.Stresser = stress.NewCPU(config.DryRun)
@@ -47,10 +47,14 @@ func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInje
 		config.ProcessManager = process.NewManager(config.DryRun)
 	}
 
+	if config.StresserManager == nil {
+		return nil, fmt.Errorf("StresserManager does not exist")
+	}
+
 	return &cpuPressureInjector{
 		spec:   spec,
 		config: config,
-	}
+	}, nil
 }
 
 func (i *cpuPressureInjector) GetDisruptionKind() types.DisruptionKindName {
@@ -58,29 +62,14 @@ func (i *cpuPressureInjector) GetDisruptionKind() types.DisruptionKindName {
 }
 
 func (i *cpuPressureInjector) Inject() error {
-	// read cpuset allocated cores
-	i.config.Log.Infow("retrieving target cpuset allocated cores")
+	cores, err := i.config.StresserManager.TrackInjectorCores(i.config.Cgroup)
 
-	cpusetCores, err := i.config.Cgroup.Read("cpuset", "cpuset.cpus")
 	if err != nil {
-		return fmt.Errorf("failed to read the target allocated cpus from the cpuset cgroup: %w", err)
+		return fmt.Errorf("failed to parse CPUSet %w", err)
 	}
-
-	// parse allocated cores
-	cores, err := cpuset.Parse(cpusetCores)
-	if err != nil {
-		return fmt.Errorf("error parsing cpuset allocated cores: %w", err)
-	}
-
-	i.config.Log.Infow(fmt.Sprintf("target identified to be running on %d cores", cores.Size()), "cores", cores.ToSlice())
-
-	// set new GOMAXPROCS value
-	oldMaxProcs := runtime.GOMAXPROCS(cores.Size())
-	i.config.Log.Infof("changed GOMAXPROCS value from %d to %d", oldMaxProcs, cores.Size())
 
 	wg := sync.WaitGroup{}
 	succeeded := true
-	tids := []int{}
 	mutex := sync.Mutex{}
 
 	// create one stress goroutine per allocated core
@@ -90,19 +79,14 @@ func (i *cpuPressureInjector) Inject() error {
 	// each thread is also niced to the highest priority
 	// because of linux scheduling, each thread will occupy a different core of allocated cores when stressing the cpu
 	for _, core := range cores.ToSlice() {
+		if i.config.StresserManager.IsCoreAlreadyStressed(core) {
+			i.config.Log.Infof("core %d is already stressed, skipping", core)
+			continue
+		}
+
 		wg.Add(1)
 
 		go func(core int) {
-			var err error
-
-			defer func() {
-				if err != nil {
-					succeeded = false
-
-					wg.Done()
-				}
-			}()
-
 			// lock the routine on the current thread
 			runtime.LockOSThread()
 			defer runtime.UnlockOSThread()
@@ -110,42 +94,16 @@ func (i *cpuPressureInjector) Inject() error {
 			// retrieve current thread PID
 			pid := i.config.ProcessManager.ThreadID()
 
-			// join target CPU cgroup
-			i.config.Log.Infow("joining target CPU cgroup", "core", core, "pid", pid)
-
-			if err = i.config.Cgroup.Join("cpu", pid, false); err != nil {
-				i.config.Log.Errorw("failed join the target CPU cgroup", "error", err, "core", core, "pid", pid)
-
-				return
+			if i.joinCgroup(core, pid) && i.prioritizeStresserProcess(core, pid) {
+				mutex.Lock()
+				i.routines++
+				mutex.Unlock()
+				wg.Done()
+				i.startStresser(core, pid)
+			} else {
+				succeeded = false
+				wg.Done()
 			}
-
-			// join target cpuset cgroup in case it is used to pin the target on specific cores
-			i.config.Log.Infow("joining target cpuset cgroup", "core", core, "pid", pid)
-
-			if err = i.config.Cgroup.Join("cpuset", pid, false); err != nil {
-				i.config.Log.Errorw("failed to join the target cpuset cgroup", "error", err, "core", core, "pid", pid)
-
-				return
-			}
-
-			// prioritize the current process
-			i.config.Log.Infow("highering current process priority", "core", core, "pid", pid)
-
-			if err = i.config.ProcessManager.Prioritize(); err != nil {
-				i.config.Log.Errorw("error highering the current process priority", "error", err, "core", core, "pid", pid)
-
-				return
-			}
-
-			i.config.Log.Infow("starting the stresser", "core", core, "pid", pid)
-
-			mutex.Lock()
-			tids = append(tids, pid)
-			i.routines++
-			mutex.Unlock()
-
-			wg.Done()
-			i.config.Stresser.Stress(i.config.StresserExit)
 		}(core)
 	}
 
@@ -156,9 +114,46 @@ func (i *cpuPressureInjector) Inject() error {
 		return fmt.Errorf("at least one stresser routine failed to execute")
 	}
 
-	i.config.Log.Infow("all routines have been created successfully, now stressing", "routinesPID", tids)
+	i.config.Log.Infow("all routines have been created successfully, now stressing", "stresserPIDPerCore", i.config.StresserManager.StresserPIDs())
 
 	return nil
+}
+
+func (i *cpuPressureInjector) joinCgroup(core int, pid int) bool {
+	// join target CPU cgroup
+	i.config.Log.Infow("joining target CPU cgroup", "core", core, "pid", pid)
+
+	if err := i.config.Cgroup.Join("cpu", pid, false); err != nil {
+		i.config.Log.Errorw("failed join the target CPU cgroup", "error", err, "core", core, "pid", pid)
+		return false
+	}
+
+	// join target cpuset cgroup in case it is used to pin the target on specific cores
+	i.config.Log.Infow("joining target cpuset cgroup", "core", core, "pid", pid)
+
+	if err := i.config.Cgroup.Join("cpuset", pid, false); err != nil {
+		i.config.Log.Errorw("failed to join the target cpuset cgroup", "error", err, "core", core, "pid", pid)
+		return false
+	}
+
+	return true
+}
+
+func (i *cpuPressureInjector) startStresser(core int, pid int) {
+	i.config.Log.Infow("starting the stresser", "core", core, "pid", pid)
+	i.config.StresserManager.TrackCoreAlreadyStressed(core, pid)
+	i.config.Stresser.Stress(i.config.StresserExit)
+}
+
+func (i *cpuPressureInjector) prioritizeStresserProcess(core int, pid int) bool {
+	i.config.Log.Infow("prioritizing the current stresser process", "core", core, "pid", pid)
+
+	if err := i.config.ProcessManager.Prioritize(); err != nil {
+		i.config.Log.Errorw("failed to prioritize the current stresser process", "error", err, "core", core, "pid", pid)
+		return false
+	}
+
+	return true
 }
 
 func (i *cpuPressureInjector) UpdateConfig(config Config) {
@@ -166,7 +161,7 @@ func (i *cpuPressureInjector) UpdateConfig(config Config) {
 }
 
 func (i *cpuPressureInjector) Clean() error {
-	i.config.Log.Info("killing routines")
+	i.config.Log.Info("killing %d routines", i.routines)
 
 	// exit the stress routines
 	for r := 0; r < i.routines; r++ {
--- a/injector/cpu_pressure_test.go
+++ b/injector/cpu_pressure_test.go
@@ -5,35 +5,35 @@
 package injector_test
 
 import (
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-	"github.com/stretchr/testify/mock"
-
 	"github.com/DataDog/chaos-controller/api/v1beta1"
 	"github.com/DataDog/chaos-controller/cgroup"
 	"github.com/DataDog/chaos-controller/container"
+	"github.com/DataDog/chaos-controller/cpuset"
 	. "github.com/DataDog/chaos-controller/injector"
 	"github.com/DataDog/chaos-controller/process"
 	"github.com/DataDog/chaos-controller/stress"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/mock"
 )
 
 var _ = Describe("Failure", func() {
 	var (
-		config        CPUPressureInjectorConfig
-		cgroupManager *cgroup.ManagerMock
-		ctn           *container.ContainerMock
-		stresser      *stress.StresserMock
-		stresserExit  chan struct{}
-		manager       *process.ManagerMock
-		inj           Injector
-		spec          v1beta1.CPUPressureSpec
+		config          CPUPressureInjectorConfig
+		cgroupManager   *cgroup.ManagerMock
+		ctn             *container.ContainerMock
+		stresser        *stress.StresserMock
+		stresserExit    chan struct{}
+		manager         *process.ManagerMock
+		inj             Injector
+		spec            v1beta1.CPUPressureSpec
+		stresserManager *stress.StresserManagerMock
 	)
 
 	BeforeEach(func() {
 		// cgroup
 		cgroupManager = &cgroup.ManagerMock{}
 		cgroupManager.On("Join", mock.Anything, mock.Anything, mock.Anything).Return(nil)
-		cgroupManager.On("Read", "cpuset", "cpuset.cpus").Return("0-1", nil)
 
 		// container
 		ctn = &container.ContainerMock{}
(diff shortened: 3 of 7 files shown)


cli/injector/cpu_pressure.go
```diff
@@ -18,10 +21,40 @@ var cpuPressureCmd = &cobra.Command{
...
-			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config}))
+			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config, DistinctPodCores: cores}))
 		}
 	},
 }
+
+func getPodCores() (string, error) {
+	var sb strings.Builder
+	for _, config := range configs {
+		log.Infow("retrieving target cpuset allocated cores")
```
clairecng, line 47: nit: I think we could also add some more logging information about what container it tries to retrieve the cpuset from, how much cpuset cores is there and use Debug to not generate too much noise? Or move this log line before the for loop to have a general log about the retrieving. Or both 🤷‍♀️ ?

cli/injector/cpu_pressure.go
```diff
@@ -18,10 +21,40 @@ var cpuPressureCmd = &cobra.Command{
 	PreRun: func(cmd *cobra.Command, args []string) {
 		// prepare spec
 		spec := v1beta1.CPUPressureSpec{}
+		podCores, err := getPodCores()
+		if err != nil {
+			log.Fatalw("error creating the node injector", "error", err)
+			return
```
luphaz, line 27:
```suggestion
			return
```

Fatal imply `os.Exit(1)`, return is not necessary

cli/injector/cpu_pressure.go
```diff
@@ -18,10 +21,40 @@ var cpuPressureCmd = &cobra.Command{
...
+		}
 
+		log.Infow(fmt.Sprintf("target identified to be running on %d cores", cores.Size()), "cores", cores.ToSlice())
 		// create injector
 		for _, config := range configs {
-			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config}))
+			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config, DistinctPodCores: cores}))
```
luphaz, line 39: I would no longer do a loop here, as you are already aggregating all configuration in your previous code, it seems to me that we'll do several time a pressure 🤔 

For every container, we'll launch as many goroutines as the total number of core used by all containers.

Not sure it's what we want 😬

Devatoria (changes requested):

cli/injector/cpu_pressure.go
```diff
@@ -18,10 +21,40 @@ var cpuPressureCmd = &cobra.Command{
...
+		if err != nil {
+			return "", err
+		}
+
+		sb.WriteString(cpusetCores)
+		sb.WriteString(",")
+	}
+
+	// remove trailing ','
+	podCores := sb.String()[:sb.Len()-1]
```
luphaz, line 58: If you are sure the trailing is supposed to be `,` you may prefer `strings.TrimSuffix`

cli/injector/cpu_pressure.go
```diff
@@ -18,10 +21,40 @@ var cpuPressureCmd = &cobra.Command{
...
+		log.Infow(fmt.Sprintf("target identified to be running on %d cores", cores.Size()), "cores", cores.ToSlice())
 		// create injector
 		for _, config := range configs {
-			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config}))
+			injectors = append(injectors, injector.NewCPUPressureInjector(spec, injector.CPUPressureInjectorConfig{Config: config, DistinctPodCores: cores}))
 		}
 	},
 }
+
+func getPodCores() (string, error) {
```
luphaz, line 44: I'm not sure it should be part of the `command` file 🤔 (that is handling inputs/outputs, but should not necessarily contains business/disruption logic IMO)

It may be more the `injector/cpu_pressure.go` responsibility to detect what are the amount of cores.

It relates to above loop, migrating it to a single call having all config could be an option, there are probably several others.

cli/injector/cpu_pressure.go
Devatoria: I would also add that we should not fatal in the `PreRun`. The injector is made so it does not exit (especially something else than exit 0) on error during the injection phase because otherwise it would prevent the disruption to be deleted properly while there's no need to block the removal if injection couldn't be done.

It is why most of this logic is supposed to be done in the injector (and not in the cli itself) so it's handled later as a "normal error" and not a fatal thing.

cli/injector/cpu_pressure.go
Devatoria: I agree. It may require some extra changes but this logic should probably be done in the injector package since it is where the injection logic is expected to be (especially to handle errors properly later as I said in another comment).

taihuynh167: there's some tradeoff here. 

Essentially I need to know the `[]cores` from all the `configs`. Now, an injector ideally shouldn't have to care about another injector (hence I didn't pass `configs` to `NewCPUPressureInjector`). 

On the separation of concern aspect, the `command` in this PR is only extracting the number of cores. It hasn't involved further injection logic. It prepares the number of cores and pass to a concrete injector to actually inject.

Let's discuss more

cli/injector/cpu_pressure.go
taihuynh167: good one, thanks. I will update

cli/injector/cpu_pressure.go
taihuynh167: I will think a bit more about error handling.

cli/injector/cpu_pressure.go
taihuynh167: as per current logic, we are indeed doing several time a pressure, each pressure for a container.

> For every container, we'll launch as many goroutines as the total number of core used by all containers.
this is actually a valid point. In Jira, the ticket descriptions say we want to stress more efficiently by targeting all cores. We probably need to confirm if this is the expected behavior

luphaz: Did you had the opportunity to look at this:
* https://github.com/DataDog/chaos-controller/blob/main/docs/cpu_pressure.md

The expected behaviour, is, for each cpuset encountered among all containers of the Pod, we want a single stresser per cpuset.

Each stresser is taking a whole CPU down (100% cpu consumption) hence there is no need to have several stresser per CPU.

cli/injector/cpu_pressure.go
Devatoria: One pattern would be to delegate that to the stresser itself. Right now, we're creating one stresser per config == per container. Stressers are independent, they have no idea about other stressers. 
We could redefine that by having a stresser manager to which you can pass additional cpuset over time. The stresser manager would spin one stresser for each given cpuset and update GOMAXPROCS accordingly. It would delegate the logic of keeping track of cpusets in another package which can be called by multiple injectors and have a glocal array of cpusets.

Let me know if it is not clear, I can give an example.

cli/injector/cpu_pressure.go
taihuynh167: I think you're correct @luphaz , thank you!

injector/cpuset_manager.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (manager *cpuStressserManager) setCoresToBeStressed(cores cpuset.CPUSet) {
+	manager.mutex.Lock()
+	defer manager.mutex.Unlock()
+	manager.coresToBeStressed = cores
+}
+
+func (manager *cpuStressserManager) GetCoresToBeStressed() cpuset.CPUSet {
+	return manager.coresToBeStressed
```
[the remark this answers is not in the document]
taihuynh167: i prefer being optimistic here. The *read* is outside of go routines and is strictly before the multithread *writes*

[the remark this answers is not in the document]
aymericDD: Alright 👍

injector/cpu_pressure.go
```diff
@@ -26,13 +26,14 @@ type cpuPressureInjector struct {
...
-	ProcessManager process.Manager
+	Stresser        stress.Stresser
+	StresserExit    chan struct{}
+	ProcessManager  process.Manager
+	StresserManager StresserManager
 }
 
 // NewCPUPressureInjector creates a CPU pressure injector with the given config
-func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInjectorConfig) Injector {
+func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInjectorConfig, stresserManager StresserManager) Injector {
```
luphaz, line 36: As it's part of `CPUPressureInjectorConfig` already provided as an argument, why adding a new argument?

You may want to change the `NewCPUPressureInjector` signature to return an error if it's empty, as it's required for most of this struct methods.

```suggestion
func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInjectorConfig) (Injector, error) {
```

aymericDD (approved):
LGTM 👍

injector/cpu_pressure.go
```diff
@@ -90,83 +75,106 @@ func (i *cpuPressureInjector) Inject() error {
...
 
-	return nil
+	i.config.Log.Infow(fmt.Sprintf("target identified to be running on %d cores", cores.Size()), "cores", cores.ToSlice())
+
+	return cores
+}
+
+func (i *cpuPressureInjector) setMaxCPU(cores int) {
+	oldMaxProcs := runtime.GOMAXPROCS(cores)
+	i.config.Log.Infof("changed GOMAXPROCS value from %d to %d", oldMaxProcs, cores)
```
luphaz, line 169: I believe this is more the responsibility of the manager to now how to define the MaxCPU.
I would move this method to it.
It would no longer need argument and would use the all cores union to define maxProc (probably protected by a mutex read)

injector/cpu_pressure.go
```diff
@@ -90,83 +75,106 @@ func (i *cpuPressureInjector) Inject() error {
...
 
-	if !succeeded {
-		return fmt.Errorf("at least one stresser routine failed to execute")
+func (i *cpuPressureInjector) detectTargetCores() cpuset.CPUSet {
+	if err := i.config.StresserManager.TrackInjectorCores(i.config); err != nil {
+		i.config.Log.Errorw("failed to parse CPUSet", "error", err)
 	}
 
-	i.config.Log.Infow("all routines have been created successfully, now stressing", "routinesPID", tids)
+	cores := i.config.StresserManager.GetCoresToBeStressed()
```
luphaz, line 160: Here we are returning all cores up to our containers.
This seems counter intuitive to me.
We still want for an injector config, the associated cores. Not the whole list of cores IMO.

I understand this works, however it feels misleading to me and error prone.

injector/cpu_pressure.go
```diff
@@ -90,83 +75,106 @@ func (i *cpuPressureInjector) Inject() error {
...
 
-	// wait for stress routines to finish initializing
-	wg.Wait()
+	return true
+}
 
-	if !succeeded {
-		return fmt.Errorf("at least one stresser routine failed to execute")
+func (i *cpuPressureInjector) detectTargetCores() cpuset.CPUSet {
+	if err := i.config.StresserManager.TrackInjectorCores(i.config); err != nil {
```
luphaz, line 156: When you track, you want indeed to perform the union, however, you still want to return the list of cores for this configuration.
You may want to change this method signature to return the list of core stressed for this injector config (that is a subset of the union, not the union, see below comments)

injector/cpuset_manager.go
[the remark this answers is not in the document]
luphaz: I'm not sure it's true, we have write/read that follow each others here:
* https://github.com/DataDog/chaos-controller/pull/602/files#diff-71cea628d383a9b3f0e1a950238cd4543ec0994beb01e9201b3f87c2815fda30R155 `detectTargetCores`

I would say we hence need to protect the read the same as the write 🤷 (with `mutex.RLock()` as @aymericDD suggested, changing your mutex definition to `sync.RWMutex`

injector/cpuset_manager.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (manager *cpuStressserManager) IsCoreAlreadyStressed(core int) bool {
+	_, isCoreStressed := manager.stresserPIDPerCore[core]
```
luphaz, line 46: Maps are not thread safe in go, hence you need also to protect the read here.
Map being accessed by several goroutines, one could change the value when you are reading it.

injector/cpuset_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	return isCoreStressed
+}
+
+func (manager *cpuStressserManager) TrackStresser(core int, stresserPID int) {
```
luphaz, line 50: As most of other methods are named another way, I would change the name to something like:
```suggestion
func (manager *cpuStressserManager) TrackCoreAlreadyStressed(core int, stresserPID int) {
```

injector/cpuset_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+}
+
+func (manager *cpuStressserManager) setCoresToBeStressed(cores cpuset.CPUSet) {
+	manager.mutex.Lock()
+	defer manager.mutex.Unlock()
+	manager.coresToBeStressed = cores
+}
+
+func (manager *cpuStressserManager) GetCoresToBeStressed() cpuset.CPUSet {
```
luphaz, line 41: GO expect getter like methods to be named without `Get`:
```suggestion
func (manager *cpuStressserManager) CoresToBeStressed() cpuset.CPUSet {
```

https://go.dev/doc/effective_go#Getters

injector/cpuset_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	return isCoreStressed
+}
+
+func (manager *cpuStressserManager) TrackStresser(core int, stresserPID int) {
+	manager.mutex.Lock()
+	defer manager.mutex.Unlock()
+	manager.stresserPIDPerCore[core] = stresserPID
+}
+
+func (manager *cpuStressserManager) GetStresserPIDs() map[int]int {
```
luphaz, line 56:
```suggestion
func (manager *cpuStressserManager) StresserPIDs() map[int]int {
```

injector/cpuset_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	// parse allocated cores
+	cores, err := cpuset.Parse(cpusetCores)
+	if err != nil {
+		return fmt.Errorf("error parsing cpuset allocated cores: %w", err)
+	}
+
+	manager.setCoresToBeStressed(manager.coresToBeStressed.Union(cores))
+
+	return nil
+}
```
luphaz, line 78: As suggested above:
```suggestion
func (manager *cpuStressserManager) TrackInjectorCores(config CPUPressureInjectorConfig) (cpuset.CPUSet , error) {
	// read cpuset allocated cores
	config.Log.Infow("retrieving target cpuset allocated cores")

	cpusetCores, err := config.Cgroup.Read("cpuset", "cpuset.cpus")
	if err != nil {
		return cpuset.CPUSet{}, fmt.Errorf("failed to read the target allocated cpus from the cpuset cgroup: %w", err)
	}

	// parse allocated cores
	cores, err := cpuset.Parse(cpusetCores)
	if err != nil {
		return cpuset.CPUSet{}, fmt.Errorf("error parsing cpuset allocated cores: %w", err)
	}

	manager.setCoresToBeStressed(manager.coresToBeStressed.Union(cores))

	return cores, nil
}
```

luphaz (commented):
Great progress here 👏 

Some suggestions to improve readability and ease further maintenance 🤗 

Feel free to ask for a chat/pairing if something is unclear to help making it to 💯

injector/cpuset_manager.go
taihuynh167: i'll add read lock as Aymeric has rightly suggested. Thanks both

injector/cpu_pressure.go
taihuynh167: yeah, good call out. Thanks

injector/cpuset_manager.go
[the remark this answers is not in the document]
luphaz: I was wondering indeed
For me it seems it's part of what is executed concurrently between two `CPUPressureInjector`.
My understanding is we do a for loop to call Inject (for each container), this is sequential, however, inside `Inject` we are creating go routines, so it feels at least the second call to `Inject`, if there is two containers, is in a concurrent context, hence requires the locking mechanism as there could be a track+detect simultaneously.

injector/cpu_pressure.go
taihuynh167: I get it. Let me rework to do 2 things
1) Get `injectorCores` (to be stressed) from the config
2) TrackInjectors should union and set GOMAXPROCS with the whole list of cores

luphaz: Yes this seems definitely where the GOMAXPROCS should be changed, once the amount of cores to be stressed changes 👍

injector/cpuset_manager.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (manager *cpuStressserManager) setCoresToBeStressed(cores cpuset.CPUSet) {
+	manager.mutex.Lock()
+	defer manager.mutex.Unlock()
+
+	manager.coresToBeStressed = cores
+	oldMaxProcs := runtime.GOMAXPROCS(cores.Size())
+
+	manager.log.Infof("changed GOMAXPROCS value from %d to %d", oldMaxProcs, cores)
```
luphaz, line 46:
```suggestion
	manager.log.Infof("changed GOMAXPROCS value from %d to %d", oldMaxProcs, cores.Size())
```

injector/cpuset_manager.go
```diff
@@ -0,0 +1,92 @@
...
+	manager.mutex.Lock()
+	defer manager.mutex.Unlock()
+	manager.stresserPIDPerCore[core] = stresserPID
+}
+
+func (manager *cpuStressserManager) StresserPIDs() map[int]int {
+	manager.mutex.RLock()
+	defer manager.mutex.RUnlock()
+	return manager.stresserPIDPerCore
+}
```
luphaz, line 72: Here as you are returning the whole map and not reading it, the `mutex` is not necessary.

However, you provide the map as a whole value object, hence, if you read it later the underlying object could be modified unexpectedly 🙈 

It's used only for logging so it may be ok 😬 

One way to ensure it's not modified would be to copy the dictionary to another one and return the copy.

injector/cpuset_manager.go
```diff
@@ -0,0 +1,92 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
Devatoria, line 1: Why is this file not part of the `stress` package? It sounds like it is related to the stresser itself.

injector/cpu_pressure_test.go
```diff
@@ -58,15 +67,16 @@ var _ = Describe("Failure", func() {
...
+			StresserExit:    stresserExit,
+			ProcessManager:  manager,
+			StresserManager: stresserManager,
 		}
 
 		// spec
 		spec = v1beta1.CPUPressureSpec{}
 
-		inj = NewCPUPressureInjector(spec, config)
+		inj, _ = NewCPUPressureInjector(spec, config)
```
luphaz, line 79:
```suggestion
		var err error
		inj, err = NewCPUPressureInjector(spec, config)
		Expect(err).To(BeNil())
```

injector/cpu_pressure_test.go
```diff
@@ -80,18 +90,26 @@ var _ = Describe("Failure", func() {
 	})
 	Describe("injection", func() {
```
luphaz, line 91: It seems we were not asserting our mock `On/Return` setup where effectively called 😬 
That is usually not a good thing because you expect such to be validated.
I discovered this because:
```go
stresserManager.On("CoresToBeStressed").Return(cpuset.NewCPUSet(0, 1))
```

Is part of our setup but never called 😬 Hence I would have expected a test failure that wasn't there.

Here is a proposal to add all of our mock assertions for all tests (tested locally):
```suggestion

	AfterEach(func() {
		manager.AssertExpectations(GinkgoT())
		stresser.AssertExpectations(GinkgoT())
		stresserManager.AssertExpectations(GinkgoT())
		cgroupManager.AssertExpectations(GinkgoT())
		ctn.AssertExpectations(GinkgoT())
	})

	Describe("injection", func() {
```

You will also need to remove the following line from the setup:
```go
cgroupManager.On("Read", "cpuset", "cpuset.cpus").Return("0-1", nil)
```

This is no longer expected as we are now mocking this

injector/cpu_pressure.go
```diff
@@ -47,40 +47,29 @@ func NewCPUPressureInjector(spec v1beta1.CPUPressureSpec, config CPUPressureInje
 		config.ProcessManager = process.NewManager(config.DryRun)
 	}
 
+	if config.StresserManager == nil {
+		return nil, fmt.Errorf("StresserManager does not exist")
```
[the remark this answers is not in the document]
taihuynh167: we share the same `StresserManager` instance across `configs`. This is required to keep track of alreadyStressedCores

luphaz (approved):
All good 💪 

Small changes in test will be appreciated to guarantee we are properly asserting our mock expectations on all runs.

I made the changes locally to confirm it's working as expected, I may have missed to properly share everything so feel free to ping if the suggested changes are not enough to have tests working.

To convince yourself it was necessary you can first add the `AssertExpectations` witness the failure, then comment out the two mentioned lines, witness the success.

injector/cpuset_manager.go
taihuynh167: it can't be part of `stress` package due to cyclic dependency.

you'll see that this `StressManager` needs to operate on `injector` (parse cores and update GOMAXPROCS). `injector` needs to share a `StressManager` instance. Hence the cycle

luphaz: What if instead of providing `config CPUPressureInjectorConfig` to `TrackInjectorCores` you provide `config.CGroup` to it? It seems it's enough to properly resolves this issue and having a clean package responsible.

It would remove the cyclic dependency.

And in fact, `StressManager` feels to need to leave into our `stress` package.

taihuynh167: good point, i think it's correct, let me rework the parameter.

Devatoria (approved):
LGTM

aymericDD (approved):
LGTM 👍

PR #603, opened 2022-10-13
Network Cloud GCP Disruption

clairecng:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- Adds gcp as a cloud provider that could be blocked in a network cloud disruption.

Google has 2 different ip ranges files: https://support.google.com/a/answer/10026322?hl=en. We chose to take the first one as it seemed the second one didn't hold all API calls to some products in Google Cloud (google cloud storage api for example).

Google unfortunately has no information about the different services it blocks in its ip ranges files. As a result, the only service available for google cloud is `Google`.

Example of disruption:
```
kind: Disruption
metadata:
  name: network-cloud
  namespace: chaos-demo
spec:
  level: pod
  selector:
    app: demo-curl
  count: 100%
  network:
    cloud:
      gcp:
        - service: "Google"
    drop: 100
```

## Code Quality Checklist

## Testing

    - GCP disruption
    - [x] locally.


controllers/disruption_controller.go
```diff
@@ -1202,7 +1202,7 @@ func (r *DisruptionReconciler) generateChaosPods(instance *chaosv1beta1.Disrupti
 		}
 
 		// get the ip ranges of cloud provider services
-		if instance.Spec.Network.Cloud != nil {
+		if instance.Spec.Network != nil && instance.Spec.Network.Cloud != nil {
```
ptnapoleon, line 1205: I thought this change was already on `main`? If it isn't, then `main` is broken right now?

clairecng: https://github.com/DataDog/chaos-controller/blob/e4a67cc099308b8f57d0f13941d801ed180de4b0/controllers/disruption_controller.go#L1205
Not sure why it shows this. But I struggled with merging 😅

clairecng: I think the GitHub view is misleading. It's the commit where I pulled from AWS onto my branch, the changes are from AWS here

ptnapoleon: Huh, that's weird. Github /shrug

api/v1beta1/disruption_webhook.go
```diff
@@ -62,6 +62,7 @@ func (r *Disruption) SetupWebhookWithManager(setupWebhookConfig utils.SetupWebho
 	handlerEnabled = setupWebhookConfig.HandlerEnabledFlag
 	defaultDuration = setupWebhookConfig.DefaultDurationFlag
 	cloudServicesProvidersManager = setupWebhookConfig.CloudServicesProvidersManager
+	cloudServicesProvidersManager = setupWebhookConfig.CloudServicesProvidersManager
```
luphaz, line 65: Not sure it's needed to have this assignation twice, the line above already does the same thing?

cloudservice/aws/aws.go
```diff
@@ -49,25 +49,28 @@ func (s *CloudProviderIPRangeManager) ConvertToGenericIPRanges(unparsedIPRanges
...
+		if ipRange.Service == "AMAZON" || ipRange.IPPrefix == "" {
 			continue
 		}
 
-		if len(genericIPRanges.IPRanges[ipRange.Service]) == 0 {
-			genericIPRanges.IPRanges[ipRange.Service] = []string{}
+		if len(result.IPRanges[ipRange.Service]) == 0 {
+			result.ServiceList = append(result.ServiceList, ipRange.Service)
+
+			result.IPRanges[ipRange.Service] = []string{}
```
luphaz, line 69: This may not be needed as append will create an array if `nil`
Here the goal is to add the service to the service list once only correct?

docs/network_disruption/cloud-managed-services.md
```diff
@@ -51,4 +56,12 @@ Available services are:
...
\ No newline at end of file
+We are using the URL **https://ip-ranges.amazonaws.com/ip-ranges.json** to pull all the IP Ranges of AWS.
+
+## GCP
+
+Available service is `Google Cloud`.
+
+Google does not indicates which ip ranges correspond to which service in its ip ranges files.
+
+We are using the URL **https://www.gstatic.com/ipranges/goog.json**. This file is the generic Google ip ranges file. We could not use the Google Cloud specific file due to some ip ranges from the apis being in the first file (goog.json). ([More info here](https://support.google.com/a/answer/10026322?hl=en))
```
luphaz, line 67: If I understand their doc and some discussions I found:
* goog.json is the whole IP list of the whole Google IPs from Google Cloud
* cloud.json is the IPs customers can use 

Hence, to have their services only we need to remove from `goog.json` what is in `cloud.json` 🙈 

Here we are blocking more things, so it seems ok if the limitation we have on `tc` allow us that.

cloudservice/gcp/gcp.go
```diff
@@ -0,0 +1,79 @@
...
+
+		// Remove empty IPPrefixes (can happen if we only have IpV6) and remove the dns servers of Google in the list of ip ranges available to disrupt
+		if ipRange.IPPrefix == "" || strings.HasPrefix(ipRange.IPPrefix, "8.8") {
+			continue
+		}
+
+		result.IPRanges[ipRange.Service] = append(result.IPRanges[ipRange.Service], ipRange.IPPrefix)
+	}
+
+	return result, nil
```
luphaz, line 78: Not sure it worth repeating the same loop than on AWS as we do not have per services:
```suggestion
// define a const to GoogleCloudService = "Google Cloud"
	result := &types.CloudProviderIPRangeInfo{
		ServiceList: []string{GoogleCloudService},
		IPRanges:    map[string][]string{
		  GoogleCloudService: []string{},
		},
		Version:     ipRanges.SyncToken,
	}

	for _, ipRange := range ipRanges.Prefixes {
		// Remove empty IPPrefixes (can happen if we only have IpV6) and remove the dns servers of Google in the list of ip ranges available to disrupt
		if ipRange.IPPrefix == "" || strings.HasPrefix(ipRange.IPPrefix, "8.8") {
			continue
		}

		result.IPRanges[GoogleCloudService] = append(result.IPRanges[GoogleCloudService], ipRange.IPPrefix)
	}

	return result, nil
```

WDYT?

cloudservice/gcp/gcp_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+var _ = BeforeSuite(func() {
+})
+
+var _ = AfterSuite(func() {
+})
+
+var _ = Describe("GCP Parsing", func() {
+	Context("Parse GCP IP Range file", func() {
+		ipRangeFile := "{\"syncToken\":\"1000000000\",\"createDate\":\"2022-09-01-22-03-06\",\"prefixes\":[{\"ipv4Prefix\": \"34.80.0.0/15\",\"service\": \"Google Cloud\",\"scope\": \"asia-east1\"},{\"ipv4Prefix\": \"34.80.0.0/15\",\"service\": \"Google Cloud\",\"scope\": \"asia-east1\"},{\"ipv4Prefix\": \"34.80.0.0/15\",\"service\": \"Google Cloud\",\"scope\": \"asia-east1\"},{\"ipv4Prefix\": \"34.80.0.0/15\",\"service\": \"Google Cloud\",\"scope\": \"asia-east1\"}]}"
```
luphaz, line 28: This value does not seems representative of what is in:
* https://www.gstatic.com/ipranges/goog.json

Is this intentional?


It seems more representative of this file:
* https://www.gstatic.com/ipranges/cloud.json

That we do not plan to use, am I correct?

cloudservice/gcp/gcp.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func New() *CloudProviderIPRangeManager {
+	return &CloudProviderIPRangeManager{}
+}
+
+// IsNewVersion Check if the ip ranges pulled are newer than the one we already have
+func (s *CloudProviderIPRangeManager) IsNewVersion(newIPRanges []byte, oldVersion string) bool {
+	ipRanges := GCPIPRanges{}
+	if err := json.Unmarshal(newIPRanges, &ipRanges); err != nil {
```
taihuynh167, line 38: does this take care of missing `newIPRanges`?

cloudservice/gcp/gcp.go
```diff
@@ -0,0 +1,79 @@
...
+		ServiceList: []string{},
+		IPRanges:    make(map[string][]string),
+		Version:     ipRanges.SyncToken,
+	}
+
+	for _, ipRange := range ipRanges.Prefixes {
+		// in the IP Ranges provided by google, no service is explicitly set
+		if ipRange.Service == "" {
+			ipRange.Service = "Google Cloud"
+		}
```
taihuynh167, line 62: how do we detect future `ipRange.Service` values 🙈 ? Our logic seems to rely on the `Google Cloud` value assignment, which is volatile.

cloudservice/aws/aws.go
```diff
@@ -49,25 +49,28 @@ func (s *CloudProviderIPRangeManager) ConvertToGenericIPRanges(unparsedIPRanges
 		return nil, err
 	}
 
-	genericIPRanges := types.CloudProviderIPRangeInfo{
-		CloudProviderServiceName: types.CloudProviderAWS,
-		Version:                  ipRanges.SyncToken,
-		IPRanges:                 make(map[string][]string),
+	result := &types.CloudProviderIPRangeInfo{
+		ServiceList: []string{},
+		IPRanges:    make(map[string][]string),
```
taihuynh167, line 54: we may need to strong-type the key of this map. Leaving it as string may cause surprise.

ptnapoleon (approved):

api/v1beta1/disruption_webhook.go
clairecng: sorry, this is a merging mistake 😓

cloudservice/aws/aws.go
clairecng: We can't really have a static list of services in case gcp decide to add a new one. What would you suggest?

cloudservice/aws/aws.go
clairecng: Thanks! I did not know an append would create the array itself

docs/network_disruption/cloud-managed-services.md
clairecng: That's what I also gathered... Even though it was not official documentation on my part, so I'm not even sure of this... Do you have a link of those discussions?

However, I think that it would add more complexity for a small improvement. I believe that we could first test out if our users are ok with blocking all google services including gcp (excluding the dns servers). If they are not, and we need to be more precise, I think we can iterate in a next PR.

The tc limitation is 2048 so we're safe (for now)! The goog.json contains only 76 ip ranges as of today.

cloudservice/gcp/gcp.go
clairecng: I'm not sure I'm understanding right what you want to say here. Could you elaborate?

cloudservice/gcp/gcp.go
clairecng: 👍 If you think it's best!

I wanted to be as generic as possible if they decide to actually add the service in their ip ranges file, or if we decide to iterate and have a more complex logic (as stated in another comment about which ip ranges we should get if we decide to get both files)

cloudservice/gcp/gcp_test.go
clairecng: You're right, thanks for catching that!

I began the implementation with the cloud.json file as it seemed logical, and forgot to replace the data.

cloudservice/gcp/gcp.go
clairecng: @taihuynh167 If we decide on having a static list of services, I'll also update the type of the key of the map that you mentioned in another comment

docs/network_disruption/cloud-managed-services.md
luphaz: This sentence made me wondered:
> The default domains' IP address ranges for Google APIs and services fit within the list of ranges between these 2 sources. (Subtract the usable ranges from the complete list.) 

I also find this:
* https://gist.github.com/n0531m/f3714f6ad6ef738a3b0a?permalink_comment_id=3880657#gistcomment-3880657

However it's a little more complex because it's CIDR substraction... So a big range could be amputed by smaller ranges, leading to a new range...

Indeed a lot of complexity that seems unnecessary for now, in particular because in our case we want to block, and not allow this traffic.
Usually people want to allow, hence aims for the more restrictive list.

We can aim for a broader one.

cloudservice/gcp/gcp.go
clairecng: Got it, you're right it would be easier to understand too. I'll change that 👍

cloudservice/gcp/gcp.go
taihuynh167: 😃 sure, i meant to check if `json.Unmarshal([0]byte{}, &ipRanges)` or `json.Unmarshal([0]byte{}, &ipRanges)` could cause problem

cloudservice/gcp/gcp.go
taihuynh167: yes Claire, thanks. It will be clearer to default `scope` and `service name` in  GCPIPRanges{}

cloudservice/aws/aws.go
taihuynh167: on second read, isn't this map always just `Google Cloud`: [IPPrefixes]` ?

cloudservice/gcp/gcp.go
clairecng: Added new unit testing cases verifying this

cloudservice/gcp/gcp.go
clairecng: 👍 Removed those as it's empty, per the suggestion of @luphaz to be easier to understand.

cloudservice/aws/aws.go
clairecng: Yes, whole explanation behind this in @luphaz comment!

luphaz (approved):

taihuynh167 (approved):

PR #607, opened 2022-10-19
Add grpc conn timeout and allow disruption clean up when there is con…

taihuynh167:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- As of now, when we created a gRPC disruption on a wrong port, we cannot connect to the gRPC server. The problem is this connection failure goes on to prevent us from deleting the disruption.
- I'm introducing timeout when connecting to server, as well as a "global" `State` so that injector `Clean()` knows if the disruption was injected earlier or not.

## Code Quality Checklist

## Testing

    - [x] locally.


injector/grpc_disruption.go
```diff
@@ -98,10 +102,11 @@ func (i *GRPCDisruptionInjector) Clean() error {
 	return conn.Close()
 }
 
-func connectToServer(serverAddr string) (*grpc.ClientConn, error) {
+func connectToServer(serverAddr string, timeout time.Duration) (*grpc.ClientConn, error) {
 	var opts []grpc.DialOption
 	opts = append(opts, grpc.WithInsecure()) // Future Work: make secure
 	opts = append(opts, grpc.WithBlock())
+	opts = append(opts, grpc.WithTimeout(timeout))
```
luphaz, line 109: I believe this is deprecated in favour of `context` usage

injector/grpc_disruption.go
```diff
@@ -98,10 +102,11 @@ func (i *GRPCDisruptionInjector) Clean() error {
...
 
 	conn, err := grpc.Dial(serverAddr, opts...)
```
luphaz, line 111: You might want to use `DialContext` here and create a context with deadline to manage the timeout.

injector/grpc_disruption.go
```diff
@@ -98,10 +102,11 @@ func (i *GRPCDisruptionInjector) Clean() error {
 	return conn.Close()
 }
 
-func connectToServer(serverAddr string) (*grpc.ClientConn, error) {
+func connectToServer(serverAddr string, timeout time.Duration) (*grpc.ClientConn, error) {
```
luphaz, line 105: As this seems part of our struct responsibility, I may move this method to a struct related, hence no more need of providing arguments at all.

cli/injector/grpc_disruption.go
```diff
@@ -72,7 +72,10 @@ var grpcDisruptionCmd = &cobra.Command{
 		// create injectors
 		for i, config := range configs {
 			if i == 0 {
-				injectors = append(injectors, injector.NewGRPCDisruptionInjector(spec, injector.GRPCDisruptionInjectorConfig{Config: config}))
+				injectors = append(injectors, injector.NewGRPCDisruptionInjector(spec, injector.GRPCDisruptionInjectorConfig{
+					Config:      config,
+					TimeoutInMs: 5000,
```
luphaz, line 77: Usually in go you use duration, they are easily parseable even when provided as arguments from the command line.
It would feel more natural to have here `Timeout` as a `time.Duration`.

You might want to create a const to make it more readable and explain briefly the arbitrary chosen value.

injector/grpc_disruption.go
```diff
@@ -81,10 +89,12 @@ func (i *GRPCDisruptionInjector) Clean() error {
 		return nil
 	}
 
-	conn, err := connectToServer(i.serverAddr)
+	conn, err := i.connectToServer()
 	if err != nil {
 		i.config.Log.Errorf(err.Error())
-		return err
+		// swallow connection err because there is nothing else we can do with the error
```
ptnapoleon, line 95: But if we are failing to connect, then we can't have cleaned, so why do we want to allow `clean` to exit successfully?

taihuynh167: my reason is that there is nothing this command can do when server is unreachable

luphaz (approved):
👍 let's see if it's what @ptnapoleon had in mind 🤗 

Looks good from my POV

PR #611, opened 2022-10-25
Removing warning event on pod deletion + better doc

clairecng:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

With the target observer, we previously would create a Warning event when a pod would be deleted even if the deletion is normal.

- Prevent the creation of this event in a normal pod deletion case
- Adds more documentation around events
- Unify the message of Events to always have an upper case
- As the documentation around events has been improved, also improving the documentation around metrics (same file)

## Code Quality Checklist

## Testing

    - no warning event thrown on rollout!
    - [x] locally.


metrics/datadog/datadog.go
```diff
@@ -73,7 +73,7 @@ func (d *Sink) MetricCleanedForReinjection(succeed bool, kind string, tags []str
 	t := []string{"status:" + status, "kind:" + kind}
 	t = append(t, tags...)
 
-	return d.metricWithStatus(metricPrefixInjector+"cleanedForReinjection", t)
+	return d.metricWithStatus(metricPrefixInjector+"cleaned_for_reinjection", t)
```
clairecng, line 76: we were using both snake case and camel case, removing camel case here.

api/v1beta1/events.go
```diff
@@ -73,95 +74,95 @@ var Events = map[string]DisruptionEvent{
...
+		OnDisruptionTemplateAggMessage: "Targeted node(s) are not ready",
 		OnTargetTemplateMessage:        EventOnTargetTemplate + "node is not ready",
 		Category:                       TargetEvent,
 	},
 	EventNodeRecoveredState: {
 		Type:                           corev1.EventTypeNormal,
 		Reason:                         EventNodeRecoveredState,
-		OnDisruptionTemplateMessage:    "targeted node %s seems to have recovered",
-		OnDisruptionTemplateAggMessage: "targeted node(s) seem to have recovered",
-		OnTargetTemplateMessage:        "node seems to have recovered from the disruption %s failure",
```
clairecng, line 164: Unifying the messages for all event to have an upper case at the beginning. This is the format of Kubernetes

controllers/cache_handler.go
```diff
@@ -243,7 +243,11 @@ func (h DisruptionTargetWatcherHandler) findNotifiableEvents(eventsToSend map[st
 					"message", event.Message,
 					"timestamp", event.LastTimestamp.Time.Unix(),
 				)
-			} else if event.Reason == "Started" {
+
+				continue
+			}
+
+			if event.Reason == "Started" {
```
taihuynh167, line 250: `else if event.Reason == "Started"` should work fine here, no?

controllers/cache_handler.go
```diff
@@ -243,7 +243,11 @@ func (h DisruptionTargetWatcherHandler) findNotifiableEvents(eventsToSend map[st
 					"message", event.Message,
 					"timestamp", event.LastTimestamp.Time.Unix(),
 				)
-			} else if event.Reason == "Started" {
+
+				continue
```
ptnapoleon, line 247: what's the difference between continuing here vs. using `else if` as before?

controllers/cache_handler.go
```diff
@@ -256,6 +260,10 @@ func (h DisruptionTargetWatcherHandler) findNotifiableEvents(eventsToSend map[st
 					"message", event.Message,
 					"timestamp", event.LastTimestamp.Time.Unix(),
 				)
+			} else if event.Reason == "Killing" && strings.Contains(event.Message, "Stopping container") && eventsToSend[chaosv1beta1.EventContainerWarningState] {
```
taihuynh167, line 263: a bit of a rant here: we have a leaky abstraction from K8s here. What better should be to define our own `Reason` enum and have another Go class to map from *K8s Reason to our own Reason*. Beside, we should less rely on string matching as these values are not in our control.

I suppose fixing this is out of scope of this PR, but hope to raise the point and hear your opinion.

controllers/cache_handler.go
clairecng: the linter was complaining as it was making a forest of ifs 😞

controllers/cache_handler.go
```diff
@@ -256,6 +260,10 @@ func (h DisruptionTargetWatcherHandler) findNotifiableEvents(eventsToSend map[st
...
+				// this event indicates a safe killing of a container (can occur when we rollout or manually delete a pod for example)
+				// we remove the warning state event if it has been created when we compared the state of the containers
+				eventsToSend[chaosv1beta1.EventContainerWarningState] = false
```
taihuynh167, line 266: what is the difference between marking it as `false` versus completely remove it from `eventsToSend` map?

controllers/cache_handler.go
ptnapoleon: I think `switch` might be more idiomatic in Go then? I'd rather suppress the linter than use the continue here, at least personally

taihuynh167 (approved):
👍 I have some minor comments. Looks good overall

controllers/cache_handler.go
clairecng: I agree with you, we can build a non-exhaustive list of all k8s events

controllers/cache_handler.go
clairecng: We can just remove it, it does not change anything. Thanks!

controllers/cache_handler.go
clairecng: I responded to the comments from @ptnapoleon

ptnapoleon (approved):

PR #612, opened 2022-10-26
feat(disruption): custom reporting

luphaz:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- This change aims to provide the ability to configure per disruption slack channel notification.

## Code Quality Checklist

## Testing

    - [x] locally.


taihuynh167 (approved):
👍 nice work!

eventnotifier/slack/slack.go
```diff
@@ -130,32 +131,21 @@ func (n *Notifier) Notify(dis v1beta1.Disruption, event corev1.Event, notifType
 		n.logger.Errorw("unable to retrieve disruption user info", "error", err, "disruption", dis.Name)
 	}
 
-	// To remove when we stop testing this feature
+	// Whenever a purpose is defined, we expect it to be available into all notifications sent messages
+	if nil != dis.Spec.Reporting && dis.Spec.Reporting.Purpose != "" {
+		disruptionBlocks = append(disruptionBlocks, slack.NewTextBlockObject("mrkdwn", "*Purpose:*\n"+dis.Spec.Reporting.Purpose, false, false))
+	}
+
 	if n.config.MirrorSlackChannelID != "" {
```
clairecng, line 139: Note: I think we can remove the config.MirrorSlackChannelID safely now, we can do it in a separate PR if you think it's irrelevant here

clairecng (approved):
LGTM! 👍

eventnotifier/slack/slack.go
luphaz: I indeed saw `// To remove when we stop testing this feature` and wondered if we still wanted to remove this in the end 😉 

Let me know if we should, I will adapt this PR accordingly.

We may not be interested of being spammed of notifications into our channel so if we believe it's more annoying than useful we should remove this.

Why was it introduced in the first place?

luphaz: Small bump here @clairecng 🤗 have you any idea on the question above?

clairecng: @luphaz It was temporarily introduced for us to see the messages users would get when we first created the event observer, so that we could detect potentially wrong events sent. I believe we can remove it now.

taihuynh167 (approved):

eventnotifier/slack/slack.go
luphaz: I will do the removal in another PR just to confirm everything is OK on the schedule side first.

PR #618, opened 2022-11-15
Fix Scope/Typo in Safemode

Azoam:
## What does this PR do?

Two big issues are fixed in this PR:
1. Typo of `networkThreshold` instead of `namespaceThreshold`
2. Using the namespace/cluster threshold GLOBAL values in a function as if they were local

The first one is pretty clear as to why it was necessary, but the second may be a bit unclear so I will explain: The way the thresholds were being used before, they were changing based on individual disruptions. If Disruption A comes in with a namespaceThreshold of 30, that was permanently change the global namespaceThreshold in memory and so the next Disruption B that comes along, we assume the default value of 30 for it's namespaceThreshold. This is not what we want, so intead we create local copies of the default values and change them ONLY if specified in their individual disruptions. If 0 is set or if the thresholds are unset, they simply keep their copied default values.

- [ ] Adds new functionality
- [X] Fixes a bug

## Code Quality Checklist

## Testing

    - `x`
    - [X] locally.


ptnapoleon (approved):

docs/safemode.md
```diff
@@ -88,4 +88,10 @@ spec:
...
     throttling:
       readBytesPerSec: 1024 # read throttling in bytes per sec
-```
\ No newline at end of file
+```
+
+## FAQ
+
+### Why is the namespace/cluster threshold not equal to what I specified in my Disruption?
+It may be the case that 0 was used as namespace/cluster threshold. Because 0 has no value as a threshold, it is simply
```
Devatoria, line 96: Do we care about the 0 value? Like, would it be useful for someone to want to specify 0 in the disruption?
If yes, why don't we turn this into a pointer to distinguish 0 from not specified?

Azoam: 0 is not useful. 0 threshold means "Trigger Safemode if the target count is > 0%" which will almost always be the case and really means nothing. We are assuming 0 is a mistake if specified or unset, in bot situations defaults to the cluster specified thresholds.

PR #621, opened 2022-11-24
feat: add status for target injections CHAOS-384

aymericDD:
## What does this PR do?

The idea is to help debugging targets by adding more context of their states.

- [x] Adds new functionality
- [ ] Fixes a bug

- Update the **DisruptionStatus**, the **TargetInjection** and the **TargetInjections** structs to add more context of targets state to know when the target changes and its current status. This will help the team and users to debug.

## View

> Disruption status: **Before** 

``` yaml
Status:
  Desired Targets Count:   1
  Ignored Targets Count:   0
  Injected Targets Count:  1
  Injection Status:        Injected
  Selected Targets Count:  1
  Targets:
    colima

```

> Disruption status: **After**

``` yaml
Status:
  Desired Targets Count:   1
  Ignored Targets Count:   0
  Injected Targets Count:  1
  Injection Status:        Injected
  Selected Targets Count:  1
  Target Injections:
    Colima:
      Injection Status:   Injected
      Injector Pod Name:  chaos-foo-7pkxl
      Since:              2022-12-08T09:57:35Z
```

As you can see, we have more context about the latest state of the target injection.

## Code Quality Checklist

## Testing

    - I have forced an injection to return a fatal error to set the status of disruption in **StuckOnRemoval**. I don't find a way to test this behavior with unit test or e2e-test for the moment. I think we need to refactor a bit the code to be able to test unitary this kind of behavior.
    - [x] locally.


api/disruption_types_test.go
```diff
@@ -22,26 +23,29 @@ var _ = Describe("DisruptionStatus.RemoveDeadTargets Test", func() {
...
-		})
-	})
-
 	When("matchingTargets contains some current targets", func() {
+		var saveTargets v1beta1.TargetInjections
+
 		BeforeEach(func() {
-			matchingTargets = status.Targets
+			matchingTargets = []string{
+				"0.0.0.0",
```
luphaz, line 35: I'm a little confused, are those supposed to be names or IPs? I believe it's the associated target name (either pod or node), maybe making those tests values more representative would be a plus (if my supposition is correct)

luphaz (commented):
👏 this is definitely a great improvement to provide visibility from the disruption

Couple of remarks focus on performance to keep the feedback loop fast in case of massive pod count.

And some doubts around maps changes in test to be confirmed/invalidated.

It seems you have created this PR from a fork 😬 so we (I) can't checkout your branch apparently (maybe I'm missing some refs on GitHub)

You may want to do your changes on the `datadog/chaos-controller` repository as you are part of Datadog 😉

controllers/disruption_controller.go
```diff
@@ -746,6 +752,12 @@ func (r *DisruptionReconciler) handleChaosPodTermination(instance *chaosv1beta1.
 		r.recordEventOnDisruption(instance, chaosv1beta1.EventDisruptionStuckOnRemoval, "")
 
 		instance.Status.IsStuckOnRemoval = true
+
+		targetInjection := instance.Status.TargetInjections[chaosPod.Labels[chaostypes.TargetLabel]]
+		targetInjection.InjectionStatus = chaostypes.DisruptionInjectionStatusIsStuckOnRemoval
+		targetInjection.Since = time.Now().Unix()
+		targetInjection.InjectorPodName = chaosPod.Name
+		instance.Status.TargetInjections[chaosPod.Labels[chaostypes.TargetLabel]] = targetInjection
```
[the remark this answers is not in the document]
Devatoria: I agree, this code seems duplicated.

controllers/disruption_controller.go
```diff
@@ -915,7 +927,10 @@ func (r *DisruptionReconciler) getEligibleTargets(instance *chaosv1beta1.Disrupt
 		}
 
 		// add target if eligible
-		eligibleTargets = append(eligibleTargets, target)
+		eligibleTargets[target] = chaosv1beta1.TargetInjection{
+			InjectionStatus: chaostypes.DisruptionInjectionStatusNotInjected,
+			Since:           time.Now().Unix(),
```
Devatoria, line 932: Once again, we have a `time.Now()` call here but keep in mind that a disruption can be reconciled very often so it does not necessarily correlate with a change on the injection status.

Devatoria (changes requested):
Firstly, could you please provide an example of the new status output in the PR description for a better understanding of the final result.

Secondly, you are using `time.Now()` at multiple occurrences in the reconcile loop (or in sub functions). Be careful, the reconcile loop can be called even outside of disruption changes and using this kind of references usually leads to wrong informations.

I would suggest to try to rely on established informations instead. As an example, a disruption is truly injected into a target only when its readiness probe is marked as ready. This information is stored in the chaos pod status [as a pod condition](https://pkg.go.dev/k8s.io/api/core/v1#PodCondition) with the last transition time (time when the pod went into this state). It is an established information.

api/v1beta1/disruption_types.go
```diff
@@ -147,6 +147,29 @@ func (dd DisruptionDuration) Duration() time.Duration {
...
+func (in TargetInjections) GetTargetNames() []string {
+	names := make([]string, len(in))
+	i := 0
+
+	for tName := range in {
+		names[i] = tName
+		i++
+	}
+
+	return names
```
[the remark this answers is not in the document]
aymericDD: Very good idea! 👍

api/disruption_types_test.go
```diff
@@ -22,26 +23,29 @@ var _ = Describe("DisruptionStatus.RemoveDeadTargets Test", func() {
...
+				"1.1.1.1",
+				"2.2.2.2",
+				"3.3.3.3",
+				"4.4.4.4",
+			}
+			saveTargets = status.TargetInjections
```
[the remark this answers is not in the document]
aymericDD: You right! I have to create a copy 👍

api/v1beta1/disruption_types.go
```diff
@@ -411,50 +434,47 @@ func (s *DisruptionSpec) GetDisruptionCount() int {
...
+	desiredTargets := TargetInjections{}
 
-	for index := 0; index < len(status.Targets); index++ {
-		if utils.Contains(matchingTargets, status.Targets[index]) {
-			desiredTargets = append(desiredTargets, status.Targets[index])
+	for name, ti := range status.TargetInjections {
+		if utils.Contains(matchingTargets, name) {
+			desiredTargets[name] = ti
 		}
 	}
```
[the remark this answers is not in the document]
aymericDD: Nice spotted!

api/disruption_types_test.go
```diff
@@ -66,49 +70,68 @@ var _ = Describe("DisruptionStatus.RemoveDeadTargets Test", func() {
...
+	var eligibleTargets v1beta1.TargetInjections
 
 	BeforeEach(func() {
 		rand.Seed(time.Now().UnixNano())
 		status = &v1beta1.DisruptionStatus{
-			Targets: []string{"0.0.0.0", "1.1.1.1", "2.2.2.2", "3.3.3.3", "4.4.4.4"},
+			TargetInjections: *makeValidTargetInjections(),
 		}
-		oldStatus = status.DeepCopy()
+		oldStatus = status
```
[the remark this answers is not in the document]
aymericDD: Oops... It's a mistake...

controllers/disruption_controller.go
```diff
@@ -342,6 +342,12 @@ func (r *DisruptionReconciler) updateInjectionStatus(instance *chaosv1beta1.Disr
 						podReady = true
 						readyPodsCount++
 
+						targetInjection := instance.Status.TargetInjections[chaosPod.Labels[chaostypes.TargetLabel]]
```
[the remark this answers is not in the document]
aymericDD: Yes, I will try this! 👍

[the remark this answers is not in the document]
aymericDD: So what I have done:
**Step 1:** Deploy the **main** branch on `colima`.
**Step 2:** Deploy the demo stack `kubectl apply -f examples/demo.yml`
**Step 3:** Apply a network disruption `kubectl apply -f examples/network_delay.yml`
**Step 4:** Check the network delay pod is deployed and running.
**Step 5:** Describe the disruption to see the targets list is up to date.
**Step 6:** Deploy the feature branch to `colima`.
**Step 7:** Describe the disruption and check if the targets list is renamed by TargetDisruptions and contains the list of target disruptions object.

Here you will find the record of the scenario: https://asciinema.org/a/M9DmC6pNh7YyFT59fBBCuVcAc

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
```diff
@@ -525,11 +525,23 @@ spec:
               selectedTargetsCount:
                 description: Actual targets selected by the disruption
                 type: integer
-              targets:
-                items:
-                  type: string
+              targetInjections:
```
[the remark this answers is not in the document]
aymericDD: Following this test I don't think wee need to handle this case https://github.com/DataDog/chaos-controller/pull/621#discussion_r1032602195

controllers/disruption_controller.go
```diff
@@ -746,16 +748,26 @@ func (r *DisruptionReconciler) handleChaosPodTermination(instance *chaosv1beta1.
...
 		instance.Status.IsStuckOnRemoval = true
+
+		r.updateTargetInjectionStatus(instance, chaosPod, chaostypes.DisruptionInjectionStatusIsStuckOnRemoval)
 	}
 }
 
+func (r *DisruptionReconciler) updateTargetInjectionStatus(instance *chaosv1beta1.Disruption, chaosPod corev1.Pod, status chaostypes.DisruptionInjectionStatus) {
+	targetInjection := instance.Status.TargetInjections[chaosPod.Labels[chaostypes.TargetLabel]]
+	targetInjection.InjectionStatus = status
+	targetInjection.Since = instance.Spec.Duration.Duration().Milliseconds()
```
Devatoria, line 759: I think I don't understand this change. The `Status.Since` is supposed to contain the timestamp of when the disruption was injected into that target. You're getting the `Spec.Duration` field value which is the maximum duration allowed for the disruption after which it'll clean itself. This value does not change over time does not represent anything related to the injection status.

What I was mentioning in my PR comment was to use pod conditions instead. On [an injector pod](https://pkg.go.dev/k8s.io/api/core/v1#Pod), and more specifically [in its status](https://pkg.go.dev/k8s.io/api/core/v1#PodStatus), you can get [its pod conditions](https://pkg.go.dev/k8s.io/api/core/v1#PodCondition). By looping over those conditions, you can search for the [PodReady](https://pkg.go.dev/k8s.io/api/core/v1#PodConditionType) one and get its `LastTransitionTime` value which is when the injector pod went to `Ready` (meaning that the readiness probe was successful, meaning that the injection was successfully done).

By the way, based on the example you pasted in the PR description, it is probably better to declare this field as a [metav1.Time](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Time) (like done in the pod condition) so on describe it displays a nice datetime format and not a timestamp number which is impossible to interpret as a human.

Devatoria (changes requested):

controllers/disruption_controller.go
aymericDD: Sorry, I don't understand well... 🙈 It's more clear for me 👍

aymericDD: Is it important to check if the pod status went ready because the pod status is defined by the function parameter? 
Is it correct if I just take the last element of the `Status.Conditions` slice and just use the value of the **LastTransitionTime** attribute?

luphaz (approved):
You might want to update your PR description with the latest type change (date-time for Since)

I'm only wondering if we want a date in the initial Since, @Devatoria may have a better idea here 🤔

api/disruption_types_test.go
```diff
@@ -66,47 +70,66 @@ var _ = Describe("DisruptionStatus.RemoveDeadTargets Test", func() {
...
 	})
 })
 
+func makeValidTargetInjections() v1beta1.TargetInjections {
+	return v1beta1.TargetInjections{
+		"target-1": {},
+		"target-2": {},
+		"target-3": {},
+		"target-4": {},
+		"target-5": {},
```
taihuynh167, line 120: consider adding an entry with valid value. It helps code-reader understand the logic more thoroughly 🤗 .

api/v1beta1/disruption_types.go
```diff
@@ -411,50 +432,57 @@ func (s *DisruptionSpec) GetDisruptionCount() int {
...
+
+	for i := 0; i < newTargetsCount && len(targetNames) > 0; i++ {
+		index := rand.Intn(len(targetNames)) //nolint:gosec
+		targetName := targetNames[index]
+		targetNames[len(targetNames)-1], targetNames[index] = targetNames[index], targetNames[len(targetNames)-1]
+		targetNames = targetNames[:len(targetNames)-1]
+
+		status.TargetInjections[targetName] = eligibleTargets[targetName]
+		delete(eligibleTargets, targetName)
 	}
```
taihuynh167, line 465: if we desire random, we can just shuffle `targetNames` before entering the for loop.

api/v1beta1/disruption_types.go
```diff
@@ -411,50 +432,57 @@ func (s *DisruptionSpec) GetDisruptionCount() int {
...
 
 // RemoveTargets removes toRemoveTargetsCount random targets from the Target List
 func (status *DisruptionStatus) RemoveTargets(toRemoveTargetsCount int) {
-	for i := 0; i < toRemoveTargetsCount && len(status.Targets) > 0; i++ {
-		index := rand.Intn(len(status.Targets)) //nolint:gosec
-		status.Targets[len(status.Targets)-1], status.Targets[index] = status.Targets[index], status.Targets[len(status.Targets)-1]
-		status.Targets = status.Targets[:len(status.Targets)-1]
+	targetNames := status.TargetInjections.GetTargetNames()
+
+	for i := 0; i < toRemoveTargetsCount && len(status.TargetInjections) > 0; i++ {
```
taihuynh167, line 472: duplicate logic, consider to extract a private method

api/v1beta1/disruption_types.go
Devatoria: I agree with that though the random logic was already here before this PR so I would not touch it within this code change which is already starting to get big.

Devatoria:
> You might want to update your PR description with the latest type change (date-time for Since)
> 
> I'm only wondering if we want a date in the initial Since, @Devatoria may have a better idea here 🤔

It would be good to see what the status looks like when injection is not done yet to see if it makes sense from a user point of view. If we really need a date in since, we can use the injector pod creation timestamp as a default value (since the state will be "not injected" and it is by default in this state since its creation).

api/disruption_types_test.go
aymericDD: I am agree with you! I will add a better context 👍

api/v1beta1/disruption_types.go
aymericDD: Good point! 👍

Devatoria (approved):
LGTM

taihuynh167 (approved):

aymericDD:
Thanks for your reviews 🙏

PR #629, opened 2022-12-27
CHAOS-408: Allow for filtering the disruption targets based on off-level resources

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- I have yet to write up docs, but I'd like to get feedback on the approach
- We have some users that want to run disruptions but need to filter their targets based on both pod and node level labels. We have no way to do that within the disruption yaml at this time, even though it is feasible. It's also quite inconvenient to manually do ahead of time. 
  - The idea is, whenever we get a list of filtered pods|nodes, we then use a secondary set of label selectors to filter for the "off-level" resource (other terms welcome). We then union the lists.
  -  For a pod level disruption, this means we will disrupt pods who match the primary selector, iff those pods are running on nodes that match our secondary selector
  - For a node level disruption, this means we will disrupt nodes who match the secondary selector, iff those nodes have pods running on them that match our secondary selector
- I chose to use the generic term "secondarySelector", to avoid having multiple new yaml fields

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -43,11 +43,11 @@ type DisruptionSpec struct {
 	// +nullable
 	AdvancedSelector []metav1.LabelSelectorRequirement `json:"advancedSelector,omitempty"` // advanced label selector
 	// +nullable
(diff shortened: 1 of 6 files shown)


ptnapoleon:
I have a better idea

PR #630, opened 2022-12-29
CHAOS-408: Allow for filtering targets with annotations

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Replacement of #629
- Adds a new top-level spec field, `filter`, which currently only has the subfield `annotation`. This field works similar to `selector`, but filters resources based on their annotations

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


targetselector/running_target_selector.go
```diff
@@ -75,6 +75,15 @@ func (r runningTargetSelector) GetMatchingPodsOverTotalPods(c client.Client, ins
...
 		}
 
+		if instance.Spec.Filters != nil && len(instance.Spec.Filters.Annotations) > 0 {
+			for k, v := range instance.Spec.Filters.Annotations {
+				podAnno, ok := pod.Annotations[k]
+				if !ok || podAnno != v {
+					continue
+				}
+			}
+		}
```
Devatoria, line 85:
```suggestion
                 // apply filters on matching resources
		if instance.Spec.Filters {
		        // apply annotation filters
			for k, v := range instance.Spec.Filters.Annotations {
				podAnno, ok := pod.Annotations[k]
				if !ok || podAnno != v {
					continue
				}
			}
		}
```

The `for range` clause will be skipped if annotations is empty so it's not needed to check its length + it prepares the ground if we want to add other kind of filters later on.

targetselector/running_target_selector.go
```diff
@@ -144,6 +153,15 @@ func (r runningTargetSelector) GetMatchingNodesOverTotalNodes(c client.Client, i
...
 
+		if instance.Spec.Filters != nil && len(instance.Spec.Filters.Annotations) > 0 {
+			for k, v := range instance.Spec.Filters.Annotations {
+				nodeAnno, ok := node.Annotations[k]
+				if !ok || nodeAnno != v {
+					continue
+				}
+			}
+		}
+
```
Devatoria, line 164: I don't like this duplicated code but I can't figure out a nice way to handle this for both kind of resources (pods and nodes) in a single function.

Devatoria (changes requested):
We should probably also adapt the chaosli to craft disruptions with this new field, right?

targetselector/running_target_selector.go
ptnapoleon: I think it's okay duplication! It's probably more extra LoC to write a block of code that works for both

cli/chaosli/cmd/create.go
```diff
@@ -674,6 +678,36 @@ For example, if you set both "app=hello-node" and "pod-name=ubuntu-uuid", then n
...
+			"Add an annotation for filtering.",
+			`Please specify this in the form of "key=value", e.g., "app=hello-node". One annotation per new-line. If you specify multiple, we will only target the union of all.
+For example, if you set both "app=hello-node" and "pod-name=ubuntu-uuid", then no matter how many pods with the annotation "app=hello-node" there were, the disruption would only target pods who also had the annotation "pod-name=ubuntu-uuid".`,
+			survey.WithValidator(survey.Required),
+			survey.WithValidator(getLabelsSetValidator()),
+		)
+
+		var filterLabels labels.Set
+
+		for _, f := range filters {
```
taihuynh167, line 694: can we merge (union) all the filters first, then call `labels`.Merge with the unified filters? Because `filterLabels` are already a `Set`, where `labels` is a customized DS

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
```diff
@@ -141,6 +141,15 @@ spec:
                 type: boolean
               duration:
                 type: string
+              filter:
```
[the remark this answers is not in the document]
ptnapoleon: As in `spec.selector`? Not without changing the format and invalidating all existing disruptions

cli/chaosli/cmd/create.go
ptnapoleon: Asked in slack, but I'm not sure how that would change the code, could you write a snippet?

Azoam (approved):
Will be really useful 🎉

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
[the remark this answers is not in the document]
taihuynh167: yes. I was thinking `key` : `operator=AnnotatedWith` : `value` properties within the `Selector`

cli/chaosli/cmd/create.go
taihuynh167: yup, I see what you meant here 🙇

api/v1beta1/disruption_types.go
```diff
@@ -187,6 +189,10 @@ type DisruptionStatus struct {
 	DesiredTargetsCount int `json:"desiredTargetsCount"`
 }
 
+type DisruptionFilter struct {
+	Annotations labels.Set `json:"annotation,omitempty"`
```
Devatoria, line 193:
```suggestion
	Annotations labels.Set `json:"annotations,omitempty"`
```

Sorry I didn't catch this one before :/
We should use `annotations` (plural) which is the same writing done in Kubernetes resources. You may want to adapt examples too (both complete and the specific one).

Devatoria (changes requested):
One typo I missed in the first review, sorry.

Also would it be possible to add this new filtering method in the tests of the reconcile loop please? We should add the annotation to some of the targeted pods ([here](https://github.com/DataDog/chaos-controller/blob/main/controllers/suite_test.go#L107)) and check for the filtering [in here](https://github.com/DataDog/chaos-controller/blob/main/controllers/disruption_controller_test.go).

Devatoria (approved):
LGTM, great work!

PR #634, opened 2023-01-09
Use Eventually when waiting for event

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- I've seen this test fail intermittently in CI these past few days. It seems very safe to wait on the event being emitted with an Eventually, rather than synchronously checking once

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/cache_handler_test.go
+++ b/controllers/cache_handler_test.go
@@ -139,7 +139,7 @@ var _ = Describe("Cache Handler verifications", func() {
 
 			By("ensuring created event was fired")
 			event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
-			Expect(event).ToNot(BeNil())
+			Eventually(event).ShouldNot(BeNil())
 
 			By("ensuring container target in warning state event was not fired")
 			event = findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventContainerWarningState], eventList.Items, targetPodA.Name)
--- a/controllers/cache_handler_test.go
+++ b/controllers/cache_handler_test.go
@@ -140,7 +140,7 @@ var _ = Describe("Cache Handler verifications", func() {
 			By("ensuring created event was fired")
 			Eventually(func(g Gomega) {
 				event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
-				Expect(event).ShouldNot(BeNil())
+				g.Expect(event).ShouldNot(BeNil())
 			}).Should(Succeed())
 
 			By("ensuring container target in warning state event was not fired")
--- a/controllers/cache_handler_test.go
+++ b/controllers/cache_handler_test.go
@@ -138,11 +138,13 @@ var _ = Describe("Cache Handler verifications", func() {
 			Expect(err).To(BeNil())
 
 			By("ensuring created event was fired")
-			event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
-			Eventually(event).ShouldNot(BeNil())
+			Eventually(func(g Gomega) {
+				event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
+				Expect(event).ShouldNot(BeNil())
+			}).Should(Succeed())
 
 			By("ensuring container target in warning state event was not fired")
-			event = findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventContainerWarningState], eventList.Items, targetPodA.Name)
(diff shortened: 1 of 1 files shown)


controllers/cache_handler_test.go
```diff
@@ -139,7 +139,7 @@ var _ = Describe("Cache Handler verifications", func() {
 
 			By("ensuring created event was fired")
 			event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
-			Expect(event).ToNot(BeNil())
+			Eventually(event).ShouldNot(BeNil())
```
ptnapoleon, line 142: dyk eventually can take values, and not just functions? I didn't until today

luphaz: Didn't knew either 💪 
I'm not sure I get how this change could help reducing the flakiness however 🤔 
We will still call `findEvent` once, and either the returned value will be valid or not, leading to the same as before?

Looking at the doc:
* https://onsi.github.io/gomega/#category-1-making-codeeventuallycode-assertions-on-values

Using Eventually with a func, by calling findEvent several time could fix the flakiness if it was caused by a too early call however.

Not sure I get everything, if I missed something please let me know 🤗

taihuynh167: TIL 💪 
Shouldn't we pass `findEvent` as a func to `Eventually`?

Haha, just realize Matthieu said the same thing.

controllers/cache_handler_test.go
```diff
@@ -138,11 +138,13 @@ var _ = Describe("Cache Handler verifications", func() {
 			Expect(err).To(BeNil())
 
 			By("ensuring created event was fired")
-			event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
-			Expect(event).ToNot(BeNil())
+			Eventually(func(g Gomega) {
+				event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
+				Expect(event).ShouldNot(BeNil())
+			}).Should(Succeed())
```
luphaz, line 144: Looking at the documentation, it seems we only have to create a func with the return type expected:

```suggestion
			Eventually(func() *v1.Event {
				return findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
			}).ShouldNot(BeNil())
```

If you still prefer the syntax with Gomega as parameter, [it seems you need](https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode) to use `g` to perform `Expect`?

```suggestion
			Eventually(func(g Gomega) {
				event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
				g.Expect(event).ShouldNot(BeNil())
			}).Should(Succeed())
```
> Note that the assertions in the body of the polled function must be performed using the passed-in g Gomega object. If you use the global DSL expectations, Eventually will not intercept any failures and the test will fail.

controllers/cache_handler_test.go
```diff
@@ -138,11 +138,13 @@ var _ = Describe("Cache Handler verifications", func() {
 			Expect(err).To(BeNil())
 
 			By("ensuring created event was fired")
-			event := findEvent(disruption.CreationTimestamp.Time, v1beta1.Events[v1beta1.EventDisrupted], eventList.Items, targetPodA.Name)
-			Expect(event).ToNot(BeNil())
+			Eventually(func(g Gomega) {
```
Devatoria, line 141: Why is `g` passed here? Naive question, I don't know what it is (we are not using it in other `Eventually` calls).

Devatoria (approved):
LGTM with a question

controllers/cache_handler_test.go
ptnapoleon: Ahh, I missed the `g.Expect`, thank you

controllers/cache_handler_test.go
ptnapoleon: I took it from the `Eventually` docs:
```

    Eventually(func(g Gomega) {
    	model, err := client.Find(1138)
    	g.Expect(err).NotTo(HaveOccurred())
    	g.Expect(model.Reticulate()).To(Succeed())
    	g.Expect(model.IsReticulated()).To(BeTrue())
    	g.Expect(model.Save()).To(Succeed())
    }).Should(Succeed())
```

luphaz: It's another signature of Gomega:
* https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

Eventually can take a value, a func returning a type to perform assertion, or taking gomega as a parameter to perform complex assertion inside the Eventually func body (and eventually returning some values also)

luphaz (approved):

taihuynh167 (approved):

PR #635, opened 2023-01-12
fix(dns_disruption): forward mitm logs

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- This PR aims to forward the Fake DNS logs to the standard outputs
  - It adapt previous `print` usage to structured logging
- This PR also fixes the following:
  - A single DNS server should be launched among all injectors (we were seeing errors "Could not start server -- is another program on udp")
  - Generate mock with Expectations to benefits from typed mock (where mock were automatically generated)
  - Fix header generation to ensure generated mocks header is properly positioned

Python log examples:
```json
{
   "name":"dns.disruption.resolver",
   "pathname":"/usr/local/bin/dns_disruption_resolver.py",
   "filename":"dns_disruption_resolver.py",
   "module":"dns_disruption_resolver",
   "exc_info":null,
   "exc_text":null,
   "stack_info":null,
   "lineno":774,
   "funcName":"<module>",
   "created":1673514234.5925462,
   "msecs":592.5462245941162,
   "relativeCreated":6.503582000732422,
   "thread":281473157210128,
   "threadName":"MainThread",
   "processName":"MainProcess",
   "process":3887,
   "disruptionName":"dns",
   "disruptionNamespace":"chaos-demo",
   "targetName":"demo-curl-598d854f7d-xgsz2",
   "targetNodeName":"colima",
   "level":"error",
   "message":"Please create a \"dns.conf\" file or specify a config path: ./fakedns.py [configfile]"
}
```

## Code Quality Checklist

## Testing

    - [x] locally.


tasks/header.py
```diff
@@ -96,7 +97,7 @@ def update_header(file: str, header: t.List, header_position: t.Dict):
     header = "".join(header)
 
     # reinsert the header on the given position
-    content[header_position["starting_pos"]:header_position["ending_pos"]] = header
+    content.insert(header_position["starting_pos"], header)
```
luphaz, line 100: @ptnapoleon this is a rollback of a change introduced [here](https://github.com/DataDog/chaos-controller/commit/b395eb7b9066423c8b34066e32d7dc35eb614cc3)

Could you confirm it's OK to rollback this?

I faced an issue using `make generate-mock` that leads to the go `package` being removed 🙈 

Why did you had to change this in the first place?

aymericDD (approved):
Great work! 👏

tasks/header.py
ptnapoleon: I made this change because I was having an issue where the new header was being inserted in front of the old header, rather than an overwrite. I tried changing some headers to be wrong and running `make header-check` with your branch here, and i'm seeing the right behavior. So.... seems fine?

ptnapoleon: Alright, i looked back at the 2021 and 2022 overwrites, and it seems I should have just used a "find all + replace" in GoLand, rather than trying to change header.py to overwrite headers, when it only seems capable of inserting headers into files that lack them. I think your changes here are correct and sufficient

injector/python_test.go
```diff
@@ -0,0 +1,202 @@
...
+		},
+		{
+			name: "when starts returns an error, error is returned to the caller",
+			expect: func(t *testing.T, mockCommand *mockCommand, logs *observer.ObservedLogs) {
+				mockCommand.EXPECT().Start().Return(errors.New("start some error"))
+			},
+			wantErr: "unable to start command, encountered error (start some error) using args ([/usr/local/bin/some_python.py -v -i]): ",
+		},
+		{
+			name: "when wait error immediatly, error is returned to the caller",
```
ptnapoleon, line 42:
```suggestion
			name: "when wait error immediately, error is returned to the caller",
```

ptnapoleon (approved):

taihuynh167 (approved):

PR #654, opened 2023-02-07
Adapt network and dns disruptions to work with cgroups v2

Devatoria:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- Network disruptions don't rely on the cgroup tc filter anymore
  - Instead, packets are marked by iptables depending on their source cgroup and filtered later by tc with a `fw` filter classifying packets based on their mark
- The solution is cgroup version agnostic and works both with v1 and v2

## Code Quality Checklist

## Testing

    - deploying the controller on lima cgroups v1 and applying network and dns disruptions and validating the result on the demo apps
    - deploying the controller on lima cgroups v2 and applying network and dns disruptions and validating the result on the demo apps
    - [x] locally.

### DIFF
--- a/injector/dns_disruption.go
+++ b/injector/dns_disruption.go
@@ -134,23 +134,23 @@ func (i *DNSDisruptionInjector) Inject() error {
 	}
 
 	// Set up iptables rules
-	if err := i.config.Iptables.CreateChain("CHAOS-DNS"); err != nil {
+	if err := i.config.Iptables.CreateChain(chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
 		return fmt.Errorf("unable to create new iptables chain: %w", err)
 	}
 
-	if err := i.config.Iptables.AddRuleWithIP("CHAOS-DNS", "udp", "53", "DNAT", podIP); err != nil {
+	if err := i.config.Iptables.AddRuleWithIP(chaostypes.InjectorIptablesChaosDNSChainName, "udp", "53", "DNAT", podIP); err != nil {
 		return fmt.Errorf("unable to create new iptables rule: %w", err)
 	}
 
 	if i.config.Level == chaostypes.DisruptionLevelPod {
 		if !i.config.OnInit {
 			// Redirect traffic marked by targeted InjectorDNSCgroupClassID to CHAOS-DNS
-			if err := i.config.Iptables.AddCgroupFilterRule("nat", "OUTPUT", i.config.Cgroup.RelativePath(""), "-p", "udp", "--dport", "53", "-j", "CHAOS-DNS"); err != nil {
+			if err := i.config.Iptables.AddCgroupFilterRule("nat", "OUTPUT", i.config.Cgroup.RelativePath(""), "-p", "udp", "--dport", "53", "-j", chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
 				return fmt.Errorf("unable to create new iptables rule: %w", err)
 			}
 		} else {
 			// Redirect all dns related traffic in the pod to CHAOS-DNS
-			if err := i.config.Iptables.AddWideFilterRule("OUTPUT", "udp", "53", "CHAOS-DNS"); err != nil {
+			if err := i.config.Iptables.AddWideFilterRule("OUTPUT", "udp", "53", chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
 				return fmt.Errorf("unable to create new iptables rule: %w", err)
 			}
 		}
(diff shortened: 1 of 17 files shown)


aymericDD (commented):
LGTM 👍

injector/dns_disruption.go
```diff
@@ -229,8 +215,29 @@ func (i *DNSDisruptionInjector) Clean() error {
 		}
 	}
 
-	if err := i.config.Iptables.ClearAndDeleteChain("CHAOS-DNS"); err != nil {
-		return fmt.Errorf("unable to remove injected iptables chain: %w", err)
+	// check if there's still a jump rule to CHAOS-DNS chain existing
+	// that would block its deletion
```
ptnapoleon, line 219: What is the leftover jump rule? Doesnt this risk leaving this chain around?

injector/dns_disruption.go
```diff
@@ -144,13 +144,8 @@ func (i *DNSDisruptionInjector) Inject() error {
...
 	if i.config.Level == chaostypes.DisruptionLevelPod {
 		if !i.config.OnInit {
-			// write classid to container net_cls cgroup - for iptable filtering
-			if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", chaostypes.InjectorCgroupClassID); err != nil {
-				return fmt.Errorf("error writing classid to pod net_cls cgroup: %w", err)
-			}
-
 			// Redirect traffic marked by targeted InjectorDNSCgroupClassID to CHAOS-DNS
-			if err := i.config.Iptables.AddCgroupFilterRule("OUTPUT", chaostypes.InjectorCgroupClassID, "udp", "53", "CHAOS-DNS"); err != nil {
+			if err := i.config.Iptables.AddCgroupFilterRule("nat", "OUTPUT", i.config.Cgroup.RelativePath(""), "-p", "udp", "--dport", "53", "-j", "CHAOS-DNS"); err != nil {
```
taihuynh167, line 148: Could i check why we no longer need to write `InjectorCgroupClassID` please? 
As per a designated comment, it seems important for the disruption to work.

```
	// InjectorCgroupClassID is linked to the TC tree in the injector network disruption.
	// Also used in the DNS Disruption to allow combined Network + DNS Disruption
	// This value should NEVER be changed without changing the Network Disruption TC tree.
	InjectorCgroupClassID = "0x00020002"
```

taihuynh167: Another question, what is the effect of adding this cgroup (relative path) into the IpTable Filter?

Consider to extract a method with a method name that explains the effect of adding these rules to help readers 🤗.

injector/dns_disruption.go
```diff
@@ -229,8 +215,29 @@ func (i *DNSDisruptionInjector) Clean() error {
...
+	// that would block its deletion
+	rules, err := i.config.Iptables.ListRules("nat", "OUTPUT")
+	if err != nil {
+		return fmt.Errorf("error listing nat output rules: %w", err)
+	}
+
+	jumpRuleExists := false
+
+	for _, rule := range rules {
+		if strings.HasSuffix(rule, "-j CHAOS-DNS") {
```
taihuynh167, line 228: this assumes that `-j CHAOS-DNS` is the last parameter. Is it always the case?

we are using string `CHAOS-DNS` quite a bit, consider to extract a constant.

taihuynh167 (changes requested):

injector/dns_disruption.go
ptnapoleon: Joris was a little terse in the PR description:

> Network disruptions don't rely on the cgroup tc filter anymore
Instead, packets are marked by iptables depending on their source cgroup and filtered later by tc with a fw filter classifying packets based on their mark

This approach won't work at all in cgroups v2, so we're completely changing how we mark the packets

injector/dns_disruption.go
taihuynh167: nit - maybe change to strings.Contains to avoid relying on order of parameter.

injector/dns_disruption.go
Devatoria: There's one rule per injector and injectors are cleaning their own rule sequentially so there's no risk of a remaining rule skipping the chain deletion unless one injector fails to clean its own rule which would lead to a stuck disruption anyway.

injector/dns_disruption.go
Devatoria: To complete what Philip said here, it's almost the same mechanism as what we used to do but instead of marking packets through the net_cls cgroup, we're marking packets through iptables depending on the cgroup path of the process creating the packet. And the filter user in tc to use that mark is different but the result is the same.

taihuynh167 (commented):
Changes LGTM!

PR #655, opened 2023-02-07
CHAOS-448: Initial working cgroups v2 library, functioning CPU pressure disruption

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Include the v1 manager in the v2 manager, so we can delegate any identical calls of the libcontainer/cgroups library
- Make some changes in the tests so that we don't run disruptions that would fail on cgroups v2 
- Change one test to only use cpu pressure so I can confirm that works
- Pass the cgroup mount to the manager so that it prefixes the paths correctly

## Code Quality Checklist

## Testing

- I've tested this locally on v2 and confirmed we can now read, join, and write to the cpuset controller.
- I can confirm net_cls disruptions explode, as expected
- W/o a larger testing environment, I can't speak to much more
- We still only have v1 CI. I would like to add v2 CI in a separate PR, if that's ok

### DIFF
--- a/controllers/disruption_controller_test.go
+++ b/controllers/disruption_controller_test.go
@@ -270,7 +270,7 @@ var _ = Describe("Disruption Controller", func() {
 		})
 	})
 
-	Context("node level", func() {
+	FContext("node level", func() {
 		BeforeEach(func() {
 			disruption = &chaosv1beta1.Disruption{
 				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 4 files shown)


aymericDD (approved):
Great job! 👍

controllers/disruption_controller_test.go
```diff
@@ -315,22 +321,10 @@ var _ = Describe("Disruption Controller", func() {
 				Unsafemode: &chaosv1beta1.UnsafemodeSpec{
 					DisableAll: true,
 				},
-				Selector:   map[string]string{"foo": "bar"},
```
Devatoria, line 318: If we're removing that because network disruption can't handle cgroups v2 right now, can we comment it out rather than removing it so we can add it back later please?

Devatoria (commented):
So in the end we're using the same logic for both cgroups v1 and v2 here, right? If yes, let's merge it like this and see how we can just be version agnostic later (pretty sure we can easily).

taihuynh167 (approved):

controllers/disruption_controller_test.go
ptnapoleon: It's more like an in-place substitution that doesn't affect what this test was originally testing (cleanup and gc), but uses a kind that we can test with v2 right now. Do we still need to change it back? I can add a comment

Devatoria: Got it :D no need to change it back

PR #659, opened 2023-02-10
Fix the filename used to read cpuset in cgroups

Devatoria:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- The file used by cpu pressure stressers to get cpuset identifiers was not working for cpu pressure, leading to no errors but no stressers being created.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/stress/cpuset_manager.go
+++ b/stress/cpuset_manager.go
@@ -7,12 +7,13 @@ package stress
 
 import (
 	"fmt"
+	"runtime"
+	"sync"
+
 	"github.com/DataDog/chaos-controller/cgroup"
 	"github.com/DataDog/chaos-controller/cpuset"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"runtime"
-	"sync"
 )
 
 type StresserManager interface {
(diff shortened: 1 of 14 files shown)


stress/cpuset_manager.go
```diff
@@ -68,7 +69,12 @@ func (manager *cpuStressserManager) TrackInjectorCores(cgroup cgroup.Manager, us
 	// read cpuset allocated cores
 	manager.log.Infow("retrieving target cpuset allocated cores")
 
-	cpusetCores, err := cgroup.Read("cpuset", "cpuset.cpus")
+	cpusetFile := "cpuset.effective_cpus"
+	if cgroup.IsCgroupV2() {
+		cpusetFile = "cpuset.cpus.effective"
```
ptnapoleon, line 74: I wonder if we can make this happen magically within `cgroup.Read`, I'll look into it

ptnapoleon (approved):

taihuynh167 (approved):

stress/cpuset_manager.go
Devatoria: That's very specific to the cpuset controller logic so I'd avoid to put that in the cgroup package. However something which could be exploited is the stats returned by the cgroups package: https://pkg.go.dev/github.com/opencontainers/runc@v1.1.4/libcontainer/cgroups#Stats

It contains stats about cpusets so maybe we can get cores from there. However it needs to make the injector process join the targeted process cgroup first to get those stats.

taihuynh167, line 74:
```suggestion
  
		// If “cpuset.cpus” is empty, the “cpuset.cpus.effective” file shows all the CPUs from the parent cgroup 
		// that can be available to be used by this cgroup. Otherwise, it should be a subset of “cpuset.cpus” 
		// unless none of the CPUs listed in “cpuset.cpus” can be granted. In this case, it will be treated just like an empty “cpuset.cpus”.
		cpusetFile = "cpuset.cpus.effective"
```

aymericDD (approved):

ptnapoleon (approved):

aymericDD (approved):
👏

PR #663, opened 2023-02-15
CHAOS-471: Dont allow disruptions to be created unless they specify where to run

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Let's bikeshed on the exact format and name of this config file.
- I've decided to put it into the spec as user's have more explicit control over that than labels.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


main.go
```diff
@@ -95,9 +95,10 @@ type controllerWebhookConfig struct {
 }
 
 type safeModeConfig struct {
-	Enable             bool `json:"enable"`
-	NamespaceThreshold int  `json:"namespaceThreshold"`
-	ClusterThreshold   int  `json:"clusterThreshold"`
+	NameMeLater        string `json:"nameMeLater"`
```
taihuynh167, line 98: some suggested names: `BlastRadius`, `ContainedTarget`, `ContainedScope`

ptnapoleon:
Not doing this as implemented

api/v1beta1/disruption_types.go
```diff
@@ -43,11 +43,12 @@ type DisruptionSpec struct {
...
-	Filter          *DisruptionFilter `json:"filter,omitempty"`
-	DryRun          bool              `json:"dryRun,omitempty"`          // enable dry-run mode
-	OnInit          bool              `json:"onInit,omitempty"`          // enable disruption on init
-	Unsafemode      *UnsafemodeSpec   `json:"unsafeMode,omitempty"`      // unsafemode spec used to turn off safemode safety nets
-	StaticTargeting bool              `json:"staticTargeting,omitempty"` // enable dynamic targeting and cluster observation
+	Filter               *DisruptionFilter `json:"filter,omitempty"`
+	DryRun               bool              `json:"dryRun,omitempty"`          // enable dry-run mode
+	OnInit               bool              `json:"onInit,omitempty"`          // enable disruption on init
+	Unsafemode           *UnsafemodeSpec   `json:"unsafeMode,omitempty"`      // unsafemode spec used to turn off safemode safety nets
+	SpecifiedEnvironment string            `json:"specifiedEnv,omitempty"`    // optional field limiting where disruptions can ru
```
taihuynh167, line 50: typo - can run (at the end of the sentence)

taihuynh167: does `Environment`  clash with some thing else?

api/v1beta1/disruption_webhook.go
```diff
@@ -100,6 +102,17 @@ func (r *Disruption) ValidateCreate() error {
 		return fmt.Errorf("invalid disruption name: %w", err)
 	}
 
+	if specifiedEnvironment != "" {
+		specifiedEnv := r.Spec.SpecifiedEnvironment
+		if len(specifiedEnv) == 0 {
```
taihuynh167, line 107: can a string != "" be empty in Go ?

examples/network_delay.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   name: network-delay
   namespace: chaos-demo
 spec:
+  specifiedEnv: lima
```
taihuynh167, line 12: oh, we have a different key in spec. Maybe just use the same for consistency?

api/v1beta1/disruption_types.go
clairecng: Maybe we should be more specific in the comment and say something like: `limiting in which kubernetes cluster disruptions can run`

api/v1beta1/disruption_webhook.go
```diff
@@ -100,6 +102,17 @@ func (r *Disruption) ValidateCreate() error {
...
+			return fmt.Errorf("disruption's does not specify an environment to run, but this controller requires it. Set your spec.specifiedEnv field")
```
clairecng, line 108:
```suggestion
			return fmt.Errorf("disruption does not specify an environment to run, but this controller requires it. Set your spec.specifiedEnv field")
```

api/v1beta1/disruption_webhook.go
clairecng: the string is by default == "" if it's not set. It can be nil if the string is a pointer (specifically the pointer will be nil), but only in this case.

api/v1beta1/disruption_types.go
aymericDD: What do you think of `AllowedEnvironment` || `AllowedEnv` or `LimitedEnvironment` ||`LimitedEnv`?

Devatoria: I'd also just go with `Environment` to keep it as simple as possible.

api/v1beta1/disruption_webhook.go
```diff
@@ -100,6 +102,17 @@ func (r *Disruption) ValidateCreate() error {
 		return fmt.Errorf("invalid disruption name: %w", err)
 	}
 
+	if specifiedEnvironment != "" {
+		specifiedEnv := r.Spec.SpecifiedEnvironment
```
Devatoria, line 106:
```suggestion
		disruptionEnv := r.Spec.SpecifiedEnvironment
```

Just to make it less confusing between the 2 variables.

main.go
```diff
@@ -95,9 +95,10 @@ type controllerWebhookConfig struct {
 }
 
 type safeModeConfig struct {
-	Enable             bool `json:"enable"`
-	NamespaceThreshold int  `json:"namespaceThreshold"`
-	ClusterThreshold   int  `json:"clusterThreshold"`
+	SpecifiedEnvironment string `json:"specifiedEnvironment"`
```
Devatoria, line 98:
```suggestion
	Environment string `json:"specifiedEnvironment"`
```

Same comment as for the disruption spec, why couldn't we shorten that to `Environment`?

api/v1beta1/disruption_types.go
Devatoria, line 50: Any reason why we're not using an annotation instead of modifying the spec for this? To me it's what it should be used for.

> You can use Kubernetes annotations to attach arbitrary non-identifying metadata to objects.

Devatoria (commented):
I added a bunch of comments but the biggest question I have right now is why we're not using an annotation instead of adding a field in the disruption spec? Annotations are perfect for metadata that are context-related and they do not change the CRD structure.

examples/network_delay.yaml
ptnapoleon: We have a different variable internally, but the json tag matches. But I'll be changing both to `environment`

api/v1beta1/disruption_webhook.go
ptnapoleon: oops, thanks!

api/v1beta1/disruption_webhook.go
```diff
@@ -44,6 +44,9 @@ var handlerEnabled bool
 var defaultDuration time.Duration
 var cloudServicesProvidersManager *cloudservice.CloudServicesProvidersManager
 var chaosNamespace string
+var safemodeEnvironment string
+
+const safeModeEnvironmentAnnotation = "chaos.datadoghq.com/environment"
```
Devatoria, line 49: I'm sorry to be triggered by that but one variable is `safemode` and the other is `safeMode` 😭 

```suggestion
const safemodeEnvironmentAnnotation = "chaos.datadoghq.com/environment"
```

api/v1beta1/disruption_webhook.go
```diff
@@ -100,6 +104,17 @@ func (r *Disruption) ValidateCreate() error {
...
 	}
 
+	if safemodeEnvironment != "" {
+		disruptionEnv, ok := r.Annotations[safeModeEnvironmentAnnotation]
+		if !ok {
+			return fmt.Errorf("disruption does not specify an environment to run, but this controller requires it. Set the annotation `%s:\"%s\"` to run on this controller", safeModeEnvironmentAnnotation, safemodeEnvironment)
+		}
+
+		if disruptionEnv != safemodeEnvironment {
+			return fmt.Errorf("disruption is configured to run in %s but has been applied in %s", disruptionEnv, safemodeEnvironment)
```
Devatoria, line 114: Can we specify where the value comes from in that error log too please (like the one above) just in case it's a mystery to the one applying the disruption?

main.go
```diff
@@ -237,6 +238,9 @@ func main() {
 	pflag.BoolVar(&cfg.Controller.UserInfoHook, "user-info-webhook", true, "Enable the mutating webhook to inject user info into disruption status")
 	handleFatalError(viper.BindPFlag("controller.userInfoHook", pflag.Lookup("user-info-webhook")))
 
+	pflag.StringVar(&cfg.Controller.SafeMode.Environment, "environment", "", "Specify the 'location' this controller is run in. All disruptions must have an annotation of chaos.datadoghq.com/environment configured with this location to be allowed to create")
```
Devatoria, line 241: Other config flags for safemode are prefixed by `safemode-`:

```suggestion
	pflag.StringVar(&cfg.Controller.SafeMode.Environment, "safemode-environment", "", "Specify the 'location' this controller is run in. All disruptions must have an annotation of chaos.datadoghq.com/environment configured with this location to be allowed to create")
```

Devatoria (changes requested):

clairecng (approved):

aymericDD (approved):
LGTM 👍 
Does the namespace safe mode should be disabled if the unsafe mode is enabled?

taihuynh167 (approved):

Devatoria (approved):

ptnapoleon:
@aymericDD 
>Does the namespace safe mode should be disabled if the unsafe mode is enabled?

I'm not sure exactly what you mean. But, despite this being under the `safemode` config section, it is enabled and disabled separately, yeah. That is a bit unintuitive, but I think for now we want this to be fully in the control of the operator, and not the user

aymericDD:
Thanks for the explanation @ptnapoleon. 👍 _I have asked this question because this is under the safemode configuration._

PR #665, opened 2023-02-17
Rework the iptables module

Devatoria:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- The iptables module from the network package has been totally revamped
- Functions now contains more business logic adapted to their use case instead of being extensions of existing iptables commands
- The module now tracks the injected rules to clean them up all at once in the `Clear` function
- It does not change any of the existing injection logic, only the code is reshaped to be easier to maintain

## Code Quality Checklist

## Testing

    - I applied and verified dns and network disruptions
    - [x] locally.

### DIFF
--- a/injector/dns_disruption.go
+++ b/injector/dns_disruption.go
@@ -133,41 +133,29 @@ func (i *DNSDisruptionInjector) Inject() error {
 		return fmt.Errorf("unable to enter the given container network namespace: %w", err)
 	}
 
-	// Set up iptables rules
-	if err := i.config.Iptables.CreateChain(chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
-		return fmt.Errorf("unable to create new iptables chain: %w", err)
-	}
-
-	if err := i.config.Iptables.AddRuleWithIP(chaostypes.InjectorIptablesChaosDNSChainName, "udp", "53", "DNAT", podIP); err != nil {
+	// Set up iptables rules to redirect dns requests to the injector pod
+	// which holds the dns proxy process
+	if err := i.config.Iptables.RedirectTo("udp", "53", podIP); err != nil {
 		return fmt.Errorf("unable to create new iptables rule: %w", err)
 	}
 
 	if i.config.Level == chaostypes.DisruptionLevelPod {
 		if !i.config.OnInit {
-			// Redirect traffic marked by targeted InjectorDNSCgroupClassID to CHAOS-DNS
-			if err := i.config.Iptables.AddCgroupFilterRule("nat", "OUTPUT", i.config.Cgroup.RelativePath(""), "-p", "udp", "--dport", "53", "-j", chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
+			// Redirect traffic coming from the targeted container to CHAOS-DNS
+			if err := i.config.Iptables.Intercept("udp", "53", i.config.Cgroup.RelativePath(""), podIP); err != nil {
 				return fmt.Errorf("unable to create new iptables rule: %w", err)
 			}
 		} else {
 			// Redirect all dns related traffic in the pod to CHAOS-DNS
-			if err := i.config.Iptables.AddWideFilterRule("OUTPUT", "udp", "53", chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
+			if err := i.config.Iptables.Intercept("udp", "53", "", ""); err != nil {
 				return fmt.Errorf("unable to create new iptables rule: %w", err)
 			}
 		}
 	}
 
 	if i.config.Level == chaostypes.DisruptionLevelNode {
-		// Exempt chaos pod from iptables re-routing
-		if err := i.config.Iptables.PrependRuleSpec(chaostypes.InjectorIptablesChaosDNSChainName, "-s", podIP, "-j", "RETURN"); err != nil {
-			return fmt.Errorf("unable to create new iptables rule: %w", err)
-		}
-
-		// Re-route all pods under node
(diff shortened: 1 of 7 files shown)


network/iptables.go
```diff
@@ -14,152 +14,189 @@ import (
...
-	return i.DeleteRuleSpec("nat", chain, "-p", protocol, "--dport", port, "-j", jump)
-}
-
-func (i iptables) DeleteRuleSpec(chain string, rulespec ...string) error {
-	if i.dryRun {
-		return nil
+	// check if the rule already exists before trying to insert it
+	exists, err := i.ip.Exists(table, chain, rulespec...)
+	if err != nil {
+		return err
```
taihuynh167, line 178: shouldn't this be the second thing we do (after `dryrun` check?)

injector/dns_disruption.go
```diff
@@ -134,40 +134,27 @@ func (i *DNSDisruptionInjector) Inject() error {
...
-		}
-
-		// Re-route all pods under node
-		if err := i.config.Iptables.PrependRuleSpec("OUTPUT", "-p", "udp", "--dport", "53", "-j", chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
-			return fmt.Errorf("unable to create new iptables rule: %w", err)
-		}
-
-		if err := i.config.Iptables.PrependRuleSpec("PREROUTING", "-p", "udp", "--dport", "53", "-j", chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
+		// Re-route all pods under node except for injector pod itself
+		if err := i.config.Iptables.Intercept("udp", "53", "", podIP); err != nil {
```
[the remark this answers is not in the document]
Devatoria: 1. We still need it, the pod IP is passed to the `Intercept` function for that reason. In fact, the `Intercept` function embeds the 3 rules (excluding chaos pod IP, prerouting redirect, output redirect).
2. It redirects any DNS packets going through prerouting to the disrupted chain, I don't have the exact details of why we need one in prerouting in addition to the one in output to be honest, I kept the same logic

network/iptables.go
Devatoria: If the given chain does not exist yet (which is the case for CHAOS-DNS when first called), the `Exists` function will return an error, this is why we first create the chain if not done yet (note that it's only for the CHAOS-DNS chain, otherwise we'll still error but it's not expected to have an unexisting chain outside of the this one).

injector/dns_disruption.go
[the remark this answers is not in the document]
taihuynh167: 1. Got it, thanks for the explanation.
2. I'm still a bit confused here. By _keeping the same logic_, do you mean the following code in `Intercept` is equivalent to the removed line? 
```
	// inject prerouting rule only if there's no cgroup path filtering
	// packets going through prerouting chain are not yet associated
	// to a process so there's no possibility to filter on a cgroup at this stage
	if cgroupPath == "" {
		if err := i.insert("nat", "PREROUTING", rulespec...); err != nil {
			return err
		}
	}
```

taihuynh167 (approved):

injector/dns_disruption.go
```diff
@@ -134,40 +134,27 @@ func (i *DNSDisruptionInjector) Inject() error {
 	}
 
 	// Set up iptables rules
-	if err := i.config.Iptables.CreateChain(chaostypes.InjectorIptablesChaosDNSChainName); err != nil {
-		return fmt.Errorf("unable to create new iptables chain: %w", err)
-	}
-
-	if err := i.config.Iptables.AddRuleWithIP(chaostypes.InjectorIptablesChaosDNSChainName, "udp", "53", "DNAT", podIP); err != nil {
+	if err := i.config.Iptables.RedirectTo("udp", "53", podIP); err != nil {
```
clairecng, line 137: Would it be possible to add a little more context to this comment? Not sure to understand what kind of rules are set up here

injector/dns_disruption_test.go
```diff
@@ -50,15 +50,9 @@ var _ = Describe("Failure", func() {
...
-		iptables.On("AddRuleWithIP", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
-		iptables.On("AddWideFilterRule", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
-		iptables.On("AddCgroupFilterRule", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
-		iptables.On("PrependRuleSpec", mock.Anything, mock.Anything).Return(nil)
-		iptables.On("DeleteRule", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
-		iptables.On("DeleteRuleSpec", mock.Anything, mock.Anything).Return(nil)
-		iptables.On("DeleteCgroupFilterRule", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
+		iptables.On("Clear").Return(nil)
+		iptables.On("RedirectTo", mock.Anything, mock.Anything, mock.Anything).Return(nil)
+		iptables.On("Intercept", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
```
clairecng, line 55: Nice 💯

injector/dns_disruption.go
Devatoria: Sure, will do that

injector/dns_disruption.go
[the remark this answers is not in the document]
Devatoria: Yes. It adds rerouting rules: https://github.com/DataDog/chaos-controller/blob/rework_iptables/network/iptables.go#L129-L141

And it ignores if it's the pod IP: https://github.com/DataDog/chaos-controller/blob/rework_iptables/network/iptables.go#L117-L120

The only difference is that instead of returning for the injector pod IP, it just ignores it in the rule.

aymericDD (approved):
Great job! 👏

PR #668, opened 2023-02-23
Fix cgroup v1 network disruptions injection

Devatoria:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- Network disruptions injection for cgroups v1 now relies on a combination of net_cls classid marking and iptables marking as suggested in the official man.
- The reason for that is that cgroup path matching in iptables was made, as a first intent, for cgroups v2. While it is working fine on our local environment, we discovered some differences in how the cgroup root is handled in linux when using cgroups v1 leading to iptables being unable to match the correct path, breaking the disruptions in some environments.
- The solution remains unchanged for cgroups v2 where there's a common root for all controllers.

## Code Quality Checklist

## Testing

    - Applying dns and network disruptions on both cgroups v1 and cgroups v2 environments.
    - [x] locally.

### DIFF
--- a/injector/dns_disruption_test.go
+++ b/injector/dns_disruption_test.go
@@ -23,18 +23,21 @@ import (
 
 var _ = Describe("Failure", func() {
 	var (
-		inj           Injector
-		config        DNSDisruptionInjectorConfig
-		spec          v1beta1.DNSDisruptionSpec
-		cgroupManager *cgroup.ManagerMock
-		netnsManager  *netns.ManagerMock
-		iptables      *network.IptablesMock
+		inj            Injector
+		config         DNSDisruptionInjectorConfig
+		spec           v1beta1.DNSDisruptionSpec
+		cgroupManager  *cgroup.ManagerMock
+		isCgroupV2Call *mock.Call
+		netnsManager   *netns.ManagerMock
+		iptables       *network.IptablesMock
 	)
 
 	BeforeEach(func() {
 		// cgroup
 		cgroupManager = &cgroup.ManagerMock{}
 		cgroupManager.On("RelativePath", mock.Anything).Return("/kubepod.slice/foo")
+		cgroupManager.On("Write", mock.Anything, mock.Anything, mock.Anything).Return(nil)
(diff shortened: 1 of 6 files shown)


injector/dns_disruption_test.go
```diff
@@ -97,16 +100,33 @@ var _ = Describe("Failure", func() {
 
 		Context("disruption is node-level", func() {
 			It("creates node-level iptable filter rules", func() {
-				iptables.AssertCalled(GinkgoT(), "Intercept", "udp", "53", "", "10.0.0.2")
+				iptables.AssertCalled(GinkgoT(), "Intercept", "udp", "53", "", "", "10.0.0.2")
```
aymericDD, line 103: What do you think if we assert the number of calls to be sure it is called once and not multiple time?
_This is valid for all Assert bellow._

network/iptables.go
```diff
@@ -134,7 +145,7 @@ func (i *iptables) Intercept(protocol string, port string, cgroupPath string, in
 	// inject prerouting rule only if there's no cgroup path filtering
 	// packets going through prerouting chain are not yet associated
 	// to a process so there's no possibility to filter on a cgroup at this stage
-	if cgroupPath == "" {
+	if cgroupPath == "" && cgroupClassID == "" {
```
aymericDD, line 148: Do you think a debug log here is pertinent?

aymericDD (approved):
LGTM! 👍  I just let some suggestions, but they are not required at all.

network/iptables.go
Devatoria: The insert call will already log at the info level the injected iptables command here so probably not useful to add one more thing.

injector/dns_disruption_test.go
Devatoria: We should do it for all calls then. It's not a bad idea but I'd suggest to do that in a separate PR for injector tests.

ptnapoleon (approved):

taihuynh167 (approved):

PR #672, opened 2023-03-08
CHAOS-471: Update README

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- I realize most people only read the readme, so we should make the annotation there clear.
- What do we think about, for DD employees, putting something in the README that urges them to check internal documentation as well? We don't need to link anything

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ kind: Disruption
 metadata:
   name: node-failure
   namespace: chaos-demo # it must be in the same namespace as targeted resources
+  annotations:
+    chaos.datadoghq.com/environment: "dev" # must match the environment you are running your disruption in. The options here are controlled by the operators of the chaos-controller
 spec:
   selector: # a label selector used to target resources
     app: demo-curl


luphaz (approved):
> What do we think about, for DD employees, putting something in the README that urges them to check internal documentation as well? We don't need to link anything

As long as the error returned by the controller provide a clear course of action, we can try first without adding more informations IMO.

README.md
```diff
@@ -38,6 +38,8 @@ kind: Disruption
 metadata:
   name: node-failure
   namespace: chaos-demo # it must be in the same namespace as targeted resources
+  annotations:
+    chaos.datadoghq.com/environment: "dev" # must match the environment you are running your disruption in. The options here are controlled by the operators of the chaos-controller
```
Devatoria, line 42: I agree with Matthieu's comment to try without this if the error message is specific enough to guide the user, OR to be more specific in the comment because it sounds to me that:
1. it is mandatory everywhere, every time
2. users will reach out to know what they are supposed to put in there (env being generic and they'll likely not check the controller settings in all envs)

ptnapoleon:
Based on Matthieu and Joris's feedback, I'll close this until we see users actually having issues

PR #678, opened 2023-03-22
CHAOS-493: Dont error on cleanup if we can't find target cgroup file

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- @luphaz this is not done, but what do you think of this approach? In `cli/injector/main.go`'s clean, we don't have enough information about the injectors or targets to skip cleaning. I also think its smart to skip as little of `clean` as possible, just in case. I am thinking we should change the `if !exists {` at L250 in `network_disruption.go` to confirm that our `Write` error was also a not found error?

I'll need to read through all the other injectors to see if we should put similar checks in there, but users are typically only hitting this on network for now.

- Has two unrelated changes. One is to block on `make lima-install-cert-manager`, because `make lima-all` kept failing for me because my laptop was a little slow in pulling the cert-manager image. Second is that we removed the `dummy` container from the `demo-curl` pod six months ago but didn't update the container failure examples.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/injector/dns_disruption.go
+++ b/injector/dns_disruption.go
@@ -206,7 +206,12 @@ func (i *DNSDisruptionInjector) Clean() error {
 	// Remove the net_cls classid for cgroup v1
 	if !i.config.Cgroup.IsCgroupV2() {
 		if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0"); err != nil {
-			return fmt.Errorf("unable to write net_cls classid: %w", err)
+			if os.IsNotExist(err) {
+				i.config.Log.Warnw("unable to find target container's net_cls.classid file, we will assume we cannot find the cgroup path because it is gone", "targetContainerID", i.config.TargetContainer.ID(), "err, err")
+				return nil
+			}
+
+			return fmt.Errorf("error cleaning net_cls classid: %w", err)
 		}
 	}
 
--- a/injector/dns_disruption.go
+++ b/injector/dns_disruption.go
@@ -207,7 +207,7 @@ func (i *DNSDisruptionInjector) Clean() error {
 	if !i.config.Cgroup.IsCgroupV2() {
 		if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0"); err != nil {
 			if os.IsNotExist(err) {
-				i.config.Log.Warnw("unable to find target container's net_cls.classid file, we will assume we cannot find the cgroup path because it is gone", "targetContainerID", i.config.TargetContainer.ID(), "err, err")
+				i.config.Log.Warnw("unable to find target container's net_cls.classid file, we will assume we cannot find the cgroup path because it is gone", "targetContainerID", i.config.TargetContainer.ID(), "err", err)
 				return nil
 			}
 
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,7 @@ lima-redeploy: lima-build-all lima-install lima-restart
 ## Install cert-manager chart
 lima-install-cert-manager:
 	$(KUBECTL) apply -f https://github.com/jetstack/cert-manager/releases/download/v1.9.1/cert-manager.yaml
+	$(KUBECTL) -n cert-manager rollout status deployment/cert-manager-webhook --timeout=180s
 
 ## Install CRDs and controller into a lima k3s cluster
 ## In order to use already built images inside the containerd runtime
--- a/cli/injector/main.go
+++ b/cli/injector/main.go
@@ -419,7 +419,7 @@ func reinject(cmdName string) error {
 	return nil
 }
 
-// clean clean all the disruptions using the list of injectors
+// clean will remove or undo all the disruptions using the list of injectors
 // returns true if cleanup succeeded, false otherwise
 func clean(kind string, sendToMetrics bool, reinjectionClean bool) bool {
 	errOnClean := false
--- a/examples/container_failure_forced.yaml
+++ b/examples/container_failure_forced.yaml
@@ -13,8 +13,8 @@ metadata:
 spec:
   selector:
     app: demo-curl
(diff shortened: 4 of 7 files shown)


container/docker.go
```diff
@@ -45,6 +46,20 @@ func (d dockerRuntime) PID(id string) (uint32, error) {
 	return uint32(ci.State.Pid), nil
 }
 
+// MaybeExists returns false if we are sure we cannot find the container, true otherwise
+func (d dockerRuntime) MaybeExists(id string) (bool, error) {
```
luphaz, line 50: Why not just `Exists`? It's always a maybe if it can not exists 😅 .

You wanted to emphasis if it exists, it's still not necessarily true? Or the opposite? 🤔

luphaz (commented):
> @luphaz this is not done, but what do you think of this approach? In cli/injector/main.go's clean, we don't have enough information about the injectors or targets to skip cleaning. I also think its smart to skip as little of clean as possible, just in case. I am thinking we should change the if !exists { at L250 in network_disruption.go to confirm that our Write error was also a not found error?

So... Let me try to rephrase what I understand:
* your actual fix check if a container exists, to avoid erroring if it doesn't
* always log informations

However, we are not sure if the file does not exists because the container does not exists...
Hence, you would also add this check (is write failing due to a `NotFound` error), and properly log it. Did I understood?
Would you also check for container existence in such case?

Or is it more if it's a file `NotFound` do not bother checking for container existence, just silently skip because there is nothing to clean as the file is missing (hence no more effect seen, hence no more clean needed)

---

Thanks for the unrelated changes 🤗

ptnapoleon:
edit: ran into a bit of a wall, seems `LoadContainer` is going to work even if the container is gone, haven't found a place in the containerd lib to tell me if the container is running or not

injector/dns_disruption.go
```diff
@@ -206,7 +207,17 @@ func (i *DNSDisruptionInjector) Clean() error {
 	// Remove the net_cls classid for cgroup v1
 	if !i.config.Cgroup.IsCgroupV2() {
 		if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0"); err != nil {
-			return fmt.Errorf("unable to write net_cls classid: %w", err)
+			targetExists, existsErr := i.config.TargetContainer.Runtime().MaybeExists(context.Background(), i.config.TargetContainer.ID())
```
luphaz, line 210: Code is 100% duplicated between disruptions (network/dns), could we find a way to have it at a single place? (the whole removal for net_clf classid for cgroup v1)?

injector/dns_disruption.go
```diff
@@ -206,7 +207,17 @@ func (i *DNSDisruptionInjector) Clean() error {
...
+			if existsErr != nil {
+				i.config.Log.Errorw("we couldn't determine if the target container exists", "cgroupErr", err, "existsErr", existsErr)
+			}
+
+			if !targetExists && os.IsNotExist(err) {
```
luphaz, line 215: > edit: ran into a bit of a wall, seems LoadContainer is going to work even if the container is gone, haven't found a place in the containerd lib to tell me if the container is running or not
Should we just assume that if the cgroup file is gone, its ok that we haven't erased our changes to it?

So I assume this would imply to remove all container related changes (`runtime.MaybeExists`) and only keep `os.IsNotExists` and do not return an error in such case, in any other error, return the error wrapped.

This sounds appealing and simpler to me 🤗

luphaz (approved):

injector/dns_disruption.go
ptnapoleon: What's left is so small that that we don't end up with less code, I no longer see the advantage

injector/dns_disruption.go
```diff
@@ -206,7 +206,12 @@ func (i *DNSDisruptionInjector) Clean() error {
 	// Remove the net_cls classid for cgroup v1
 	if !i.config.Cgroup.IsCgroupV2() {
 		if err := i.config.Cgroup.Write("net_cls", "net_cls.classid", "0"); err != nil {
-			return fmt.Errorf("unable to write net_cls classid: %w", err)
+			if os.IsNotExist(err) {
+				i.config.Log.Warnw("unable to find target container's net_cls.classid file, we will assume we cannot find the cgroup path because it is gone", "targetContainerID", i.config.TargetContainer.ID(), "err, err")
```
ptnapoleon, line 210: I am thinking we prefer Warn here, because it is slightly dangerous, and I'd definitely like to know when we did this while reading logs, rather than finding it in Debug?

luphaz: Yep 👍

injector/dns_disruption.go
luphaz: From my POV it's still 10 LOC, and a risk to miss a change if at any point we need to iterate on this.

What I am thinking here, would consist of creating a func in the `injector` package to centralise the IsCgroupV2 logic + Write + error handling. 

It's not entirely Cgroup logic, more our logic, hence `injector`.

luphaz: From my POV it's still 10 LOC, and a risk to miss a change if at any point we need to iterate on this.

What I am thinking here, would consist of creating a func in the `injector` package to centralise the IsCgroupV2 logic + Write + error handling. 

It's not entirely Cgroup logic, more our logic, hence `injector`.

luphaz (approved):

injector/dns_disruption.go
Devatoria, line 210:
```suggestion
				i.config.Log.Warnw("unable to find target container's net_cls.classid file, we will assume we cannot find the cgroup path because it is gone", "targetContainerID", i.config.TargetContainer.ID(), "err", err)
```

Devatoria (approved):

PR #679, opened 2023-03-28
Network disruption on services with a list of affected ports

clairecng:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- Network disruption on services can now optionally take a list of affected ports instead of directly targeting all of the service's ports.

## Code Quality Checklist

## Testing

    -  Bad port
    - [x] locally.
    -  Good port
    - [ ] locally.
    -  List of ports
    - [x] locally.
    -  No port
    - [x] locally.
    -  Update service ports and verify allowed ports are still there
    - [x] locally.


api/v1beta1/network_disruption.go
```diff
@@ -98,6 +98,8 @@ type NetworkDisruptionHostSpec struct {
 type NetworkDisruptionServiceSpec struct {
 	Name      string `json:"name"`
 	Namespace string `json:"namespace"`
+	// +optional
+	Ports []int `json:"ports,omitempty"`
```
clairecng, line 102: Can't really add kubebuilder validation in there. what do you think? Because we check at the validation that the port exists in the service I personally think it's fine, but I could be wrong.

clairecng: If we add protocol as well let's do this then, thanks!

luphaz: So if I take your example with `protocol`:
```yaml
services:
  - name: demo # service name
    namespace: chaos-demo # service namespace
    port: 8080 # optional. affected port. No port means all ports are affected
    protocol: udp
  - name: demo # service name
    namespace: chaos-demo # service namespace
    port: 8081 # optional. affected port. No port means all ports are affected
```

Instead of this:
```yaml
services:
  - name: demo # service name
    namespace: chaos-demo # service namespace
    ports:
       - port: 8080 # optional. affected port. No port means all ports are affected
         protocol: udp
       - port: 8081 # optional. affected port. No port means all ports are affected
```

You could also create a dedicated type being `ports: [1,2,3,4]` that would define the protocol to TCP as a default value 😄 

Both declaration seems fine to me, I may prefer the one closer to Kubernetes service definition (ease copy paste)

clairecng: I prefer to do this as well, i'll settle on your suggestion. Thanks Matthieu!

injector/network_disruption_test.go
```diff
@@ -358,6 +366,50 @@ var _ = Describe("Failure", func() {
...
+					time.Sleep(300 * time.Millisecond)
+					servicesWatcher.Add(fakeService)
+					time.Sleep(300 * time.Millisecond)
+					podsWatcher.Add(fakeEndpoint)
+				}()
+			})
+
+			It("should add a filter for every service and pods filtered on", func() {
+				// wait for all the addFilters at the beginning of injection to complete
+				time.Sleep(5 * time.Second)
```
aymericDD, line 397: Do we need this sleep if we used **Eventually** with a 5 second timeout?

injector/network_disruption_test.go
```diff
@@ -358,6 +366,50 @@ var _ = Describe("Failure", func() {
...
+				podsWatcher := watch.NewFake()
+				servicesWatcher := watch.NewFake()
+
+				k8sClient.PrependWatchReactor("pods", testing.DefaultWatchReactor(podsWatcher, nil))
+				k8sClient.PrependWatchReactor("services", testing.DefaultWatchReactor(servicesWatcher, nil))
+
+				// fake watchers for service handling
+				go func() {
+					// Set up
+					time.Sleep(300 * time.Millisecond)
```
aymericDD, line 388: Do we really need to add this two sleeps?
_IMO: Sleep statements in unit tests can significantly slow down the execution time, which can be inconvenient and difficult to work with._

aymericDD (approved):
Looks good to me 👍  Just some comments about unit tests.

api/v1beta1/network_disruption.go
```diff
@@ -389,16 +408,41 @@ func NetworkDisruptionServiceSpecFromString(services []string) ([]NetworkDisrupt
...
 			return nil, fmt.Errorf("unexpected service format: %s", service)
 		}
 
+		ports := []NetworkDisruptionServicePortSpec{}
+
+		for _, unparsedPort := range parsedService[2:] {
+			// <port-value>-<port-protocol>-<port-name>
+			parsedPort := strings.Split(unparsedPort, "-")
+			if len(parsedPort) != 3 {
+				return nil, fmt.Errorf("unexpected service port format: %s", unparsedPort)
```
luphaz, line 423:
```suggestion
				return nil, fmt.Errorf("service port format is expected to follow '<port>-<procotol>-<name>', invalid format detected: %s", unparsedPort)
```

api/v1beta1/network_disruption.go
```diff
@@ -389,16 +408,41 @@ func NetworkDisruptionServiceSpecFromString(services []string) ([]NetworkDisrupt
 
 	// parse given services
 	for _, service := range services {
-		// parse service with format <name>;<namespace>
+		// parse service with format <name>;<namespace>;<port-value>-<port-protocol>-<port-name>;<port-value>-<port-protocol>-<port-name>...
 		parsedService := strings.Split(service, ";")
-		if len(parsedService) != 2 {
+		if len(parsedService) < 2 {
 			return nil, fmt.Errorf("unexpected service format: %s", service)
```
luphaz, line 414:
```suggestion
			return nil, fmt.Errorf("service format is expected to follow '<name>;<namespace>;<port-value>-<port-protocol>-<port-name>;<port-value>-<port-protocol>-<port-name>', unexpected format detected: %s", service)
```

api/v1beta1/network_disruption.go
```diff
@@ -389,16 +408,41 @@ func NetworkDisruptionServiceSpecFromString(services []string) ([]NetworkDisrupt
...
+		for _, unparsedPort := range parsedService[2:] {
+			// <port-value>-<port-protocol>-<port-name>
+			parsedPort := strings.Split(unparsedPort, "-")
+			if len(parsedPort) != 3 {
+				return nil, fmt.Errorf("unexpected service port format: %s", unparsedPort)
+			}
+
+			port, err := strconv.Atoi(parsedPort[0])
+			if err != nil {
+				return nil, fmt.Errorf("unexpected port format in service port: %s", unparsedPort)
```
luphaz, line 428:
```suggestion
				return nil, fmt.Errorf("port format is expected to be a valid integer, unexpected format detected in service port: %s", unparsedPort)
```

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
```diff
@@ -407,6 +407,26 @@ spec:
...
+                              port:
+                                maximum: 65535
+                                minimum: 0
+                                type: integer
+                              protocol:
+                                enum:
+                                - tcp
+                                - udp
+                                - sctp
+                                - ""
```
luphaz, line 424: Could we set a default instead?
To make it explicit to our users we will use tcp when value is not specified?

cli/chaosli/cmd/create.go
```diff
@@ -525,6 +525,36 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
 		service.Name = getInput("What is the name of this service?", "", survey.WithValidator(survey.Required))
 		service.Namespace = getInput("What namespace is this service in?", "", survey.WithValidator(survey.Required))
 
+		hasPorts := confirmOption("Do you want to provide the ports affected by the disruption? (In case of no port defined, all ports of the service will be affected)", "")
+
+		ports := []v1beta1.NetworkDisruptionServicePortSpec{}
+
+		if hasPorts {
```
luphaz, line 532:
```suggestion
		if hasPorts {
		  ports := []v1beta1.NetworkDisruptionServicePortSpec{}

```

cli/chaosli/cmd/create.go
```diff
@@ -525,6 +525,36 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
+
+				if confirmOption("Would you like to specify the name for this port?", "This field is optional and is used to find the right port to be affected in case the service has multiple ports") {
+					port.Name = getInput("Please enter the name of the port for this service (or ctrl+c to go back)", "")
+				}
+
+				if confirmOption("Would you like to specify the protocol for this port?", "This field is optional and is used to find the right port to be affected in case the service has multiple ports") {
+					port.Protocol, _ = selectInput("Please choose then (or ctrl+c to go back)", []string{"tcp", "udp", "sctp"}, "")
+				}
+
+				port.Port, _ = strconv.Atoi(getInput("What port would you like to target?", "", survey.WithValidator(integerValidator), survey.WithValidator(survey.Required)))
```
luphaz, line 544: I would reorder the question to first have the mandatory field, after the optional fields

cli/chaosli/cmd/create.go
```diff
@@ -525,6 +525,36 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
+				return port
+			}
+
+			ports = append(ports, getServicePort())
+
+			for confirmOption("Would you like to add another k8s service port affected by the disruption?", "") {
+				ports = append(ports, getServicePort())
+			}
+
+			service.Ports = ports
```
luphaz, line 555: Should we instead of creating a new array directly use the `service.Ports`?

docs/network_disruption/hosts.md
```diff
@@ -22,14 +22,19 @@ as NAT rules are applied before `tc` rules, and thus the port that a pod uses to
...
+
 ```
 network:
   services:
     - name: service_name
       namespace: example_namespace
+      ports:
+        - port: 8080 
+          name: service_port_name # optional. This is used to find the right port to be affected in case the service has multiple ports
+          protocol: tcp #optional. This is used to find the right port to be affected in case the service has multiple ports
```
luphaz, line 37: You may want to clarify what value will be used if this is not defined, I believe it's tcp

examples/network_filters.yaml
```diff
@@ -25,3 +25,8 @@ spec:
     services: # filter on Kubernetes services; this will correctly handle the port differences in node vs. pod-level disruptions
       - name: demo # service name
         namespace: chaos-demo # service namespace
+        ports: # optional. List of affected ports. No list means all ports are affected
+          - port: 8080
+            name: regular # optional. Name of the port, used to identify the port affected
```
luphaz, line 30: Per my understanding, user can't specify only the name correct?
In a Kubernetes definition, there is a nice feature for `targetPort`:
* https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports

You can provide either a string or a number. If you provide a string Kubernetes will target the named port and you don't have to provide an int value.

Is this something we could /should do?
Hence probably removing "name" and renaming "port" to "servicePort" (I nearly suggested targetPort but it could be confusing in the end)

WDYT?

injector/network_disruption.go
```diff
@@ -459,13 +459,13 @@ func (i *networkDisruptionInjector) addServiceFilters(serviceName string, filter
...
 
 	for _, filter := range filters {
-		i.config.Log.Infow("found service endpoint", "resolvedEndpoint", filter.service.String(), "resolvedService", serviceName)
-
 		filter.priority, err = i.config.TrafficController.AddFilter(interfaces, "1:0", "", nil, filter.service.ip, 0, filter.service.port, filter.service.protocol, network.ConnStateUndefined, flowid)
 		if err != nil {
 			return nil, err
 		}
 
+		i.config.Log.Infow(fmt.Sprintf("added a tc filter for service %s-%s with priority %d", serviceName, filter.service.String(), filter.priority), "interfaces", strings.Join(interfaces, ", "))
```
luphaz, line 467:
```suggestion
		i.config.Log.Infow(fmt.Sprintf("added a tc filter for service %s-%s with priority %d", serviceName, filter.service, filter.priority), "interfaces", interfaces)
```
How does it look? It should be pretty similar IMO

injector/network_disruption.go
```diff
@@ -480,7 +480,7 @@ func (i *networkDisruptionInjector) removeServiceFilter(interfaces []string, tcF
 		}
 	}
 
-	i.config.Log.Infow(fmt.Sprintf("deleted a tc filter on %s", tcFilter.service.String()), "interfaces", strings.Join(interfaces, ", "))
+	i.config.Log.Infow(fmt.Sprintf("deleted a tc filter for service %s with priority %d", tcFilter.service.String(), tcFilter.priority), "interfaces", strings.Join(interfaces, ", "))
```
luphaz, line 483:
```suggestion
	i.config.Log.Infow(fmt.Sprintf("deleted a tc filter for service %s with priority %d", tcFilter.service, tcFilter.priority), "interfaces", interfaces)
```

injector/network_disruption.go
```diff
@@ -552,6 +552,28 @@ func (i *networkDisruptionInjector) buildServiceFiltersFromService(service v1.Se
 	return endpointsToWatch
 }
 
+func (i *networkDisruptionInjector) extractValidPortListInServicePorts(serviceSpec v1beta1.NetworkDisruptionServiceSpec, k8sService *v1.Service) []v1.ServicePort {
+	ports := k8sService.Spec.Ports
```
luphaz, line 556: WDYT of providing the intent this way:
* if `len(serviceSpec.Ports) == 0` then `return k8sService.Spec.Ports`
  * early quit, our most common case for now I think

Then, you could build a dictionary of "allowedPorts" from `serviceSpec.ports` with a key similar to:
* `<name|port>-<protocol(default=tcp)>`

Then, you compare `k8sService.Spec.Ports` to such dictionary and add it to `ports` if matched.

api/v1beta1/validations.go
```diff
@@ -35,6 +35,36 @@ func validateServices(k8sClient client.Client, services []NetworkDisruptionServi
 			return fmt.Errorf("error retrieving the specified network disruption service: %w", err)
 		}
 
+		for _, port := range service.Ports {
```
luphaz, line 38: I would tend to avoid imbricated loops when it's possible to build a dictionary before instead.

I made a suggestion on network_disruption implementation that seems could be reused here in another way here.

WDYT?

chart/templates/crds/chaos.datadoghq.com_disruptions.yaml
clairecng: To me the default value means we don't try to find a servicePort with an associated protocol, so it can be anything.
We actually don't use the value except for filtering / finding the servicePort, so it makes no sense to me having a default. Meaning that if I specify this:

```
port: 8080
name: hello-world
```

and I have this set up in my service:
```
port: 8080
name: hello-world
protocol: udp
```

it finds this servicePort.

WDYT?

luphaz: 👍 ok 😅 

So, let me try to rephrase to ensure I understood:
* If I do not specify a protocol value, the service port associated, no matter it's protocol, will be "matched"
* If I specify a protocol value, the service port associated need to have specified protocol to be "matched"

🤔 

With this new enlightening I'm wondering why we should have protocol in the end (I know, it's me that suggested it...)

It seems it provide a filtering layer that is useless IMO (protect against an unexpected port protocol change?)

It feels, it's overly complex in the end, for a relative added value nobody asked for.

Hence, I would nearly get back to your original implementation, a simple array...

Except, this array could contain `intOrString` int being the port value, string being the port name.

And the only task we do, is we match against this array values (either port or name)

The protocol is automatically retrieved from the Kubernetes service definition (and we might emphasis this, as I forgot it during the review TBH).

Sorry for the back and forth, and THANKS for highlighting the misunderstanding I had 🤗

luphaz: If you have a strong opinion on keeping the protocol to provide additional filtering capabilities, I'm also fine with it 🤗 

Just wanted to highlight my comprehension and challenge the benefits from my POV

clairecng: I agree with you on the fact that maybe specifying the protocol is overkill 😅. I would however prefer to have a servicePort definition with an optional name and/or port instead of having an array of either port or names. I feel like it would be clearer for the user but maybe having the added logic of specifying name: or port: would be even more confusing? What do you think?

So, it would look like:

```
services:
  - name: demo # service name
    namespace: chaos-demo # service namespace
    servicePorts:
       - port: 8080
       - name: special-port
```

instead of:

```
services:
  - name: demo # service name
    namespace: chaos-demo # service namespace
    servicePorts: 
       - 8080
       - special-port
```

I'll also add more documentation.

ptnapoleon: Is it possible in k8s to have a service set up that re-uses the same port but for different protocols?

luphaz: > Is it possible in k8s to have a service set up that re-uses the same port but for different protocols?

At least nothing is preventing you to do so (tested locally)

That imply the destination application is able to differentiate traffic received on the same port (multiplexer?), I'm not sure it's somehow possible 🤔 I do not know TBH.

The suggested implementation (without protocol but with a dedicated object) enables us to extend the implementation later if we have feedbacks requiring such use case.

Would it be enough for you?

api/v1beta1/network_disruption.go
```diff
@@ -458,3 +454,46 @@ func (h NetworkDisruptionHostSpec) Validate() error {
...
+
+	servicePortsDic := map[string]v1.ServicePort{}
+	goodPorts, notFoundPorts := []v1.ServicePort{}, []NetworkDisruptionServicePortSpec{}
+
+	// Convert service ports from found k8s service to a dictionary in order to facilitate the filtering of the ports
+	for _, port := range k8sService.Spec.Ports {
+		servicePortsDic[strconv.Itoa(int(port.Port))] = port
+		if port.Name != "" {
+			servicePortsDic[port.Name] = port
+		}
```
luphaz, line 471: if someone is weirdly defining a name of the port as a number into the k8s service, you might end up with an issue.

Maybe use a prefix to guarantee you know what you are looking for (either a number or a name, that can also be a number)

```suggestion
		servicePortsDic[fmt.Sprintf("port-%d", port.Port)] = port
		if port.Name != "" {
			servicePortsDic[fmt.Sprintf("name-%s", port.Name)] = port
		}
```

api/v1beta1/network_disruption.go
```diff
@@ -458,3 +454,46 @@ func (h NetworkDisruptionHostSpec) Validate() error {
...
+
+	for _, allowedPort := range s.Ports {
+		if allowedPort.Port != 0 {
+			servicePort := servicePortsDic[strconv.Itoa(allowedPort.Port)]
+
+			if servicePort.Port == 0 || (allowedPort.Name != "" && allowedPort.Name != servicePort.Name) {
+				notFoundPorts = append(notFoundPorts, allowedPort)
+
+				continue
+			}
```
luphaz, line 482:
```suggestion
			servicePort, ok := servicePortsDic[fmt.Sprintf("port-%d", allowedPort.Port)]

			if !ok || (allowedPort.Name != "" && allowedPort.Name != servicePort.Name) {
				notFoundPorts = append(notFoundPorts, allowedPort)

				continue
			}
```

I'm not sure about discarding a port that is found TBH 🤔 (found by port number, but not matching name)

What is your reasoning/feeling that led you to do this?

api/v1beta1/network_disruption.go
```diff
@@ -458,3 +454,46 @@ func (h NetworkDisruptionHostSpec) Validate() error {
...
+
+			goodPorts = append(goodPorts, servicePort)
+		} else if allowedPort.Name != "" {
+			servicePort := servicePortsDic[allowedPort.Name]
+
+			if servicePort.Port == 0 || (allowedPort.Port != 0 && servicePort.Port == int32(allowedPort.Port)) {
+				notFoundPorts = append(notFoundPorts, allowedPort)
+
+				continue
+			}
```
luphaz, line 492:
```suggestion
			servicePort := servicePortsDic[fmt.Sprintf("name-%d", allowedPort.Name)]

			if !ok {
				notFoundPorts = append(notFoundPorts, allowedPort)

				continue
			}
```

`allowedPort != 0` can't happen because part of first if branch.

cli/chaosli/cmd/create.go
```diff
@@ -527,29 +527,39 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
 
-				if confirmOption("Would you like to specify the protocol for this port?", "This field is optional and is used to find the right port to be affected in case the service has multiple ports") {
-					port.Protocol, _ = selectInput("Please choose then (or ctrl+c to go back)", []string{"tcp", "udp", "sctp"}, "")
+				if confirmOption("Would you like to specify the port value for this port?", "This field is optional and is used to find the right port to be affected in case the service has multiple ports") {
+					port.Port, _ = strconv.Atoi(getInput("Please enter the name of the port for this service (or ctrl+c to go back)", "", survey.WithValidator(integerValidator), survey.WithValidator(survey.Required)))
 				}
 
-				port.Port, _ = strconv.Atoi(getInput("What port would you like to target?", "", survey.WithValidator(integerValidator), survey.WithValidator(survey.Required)))
+				if port.Name == "" && port.Port == 0 {
+					fmt.Println(`No port name or port value was specified.`)
```
luphaz, line 547:
```suggestion
					fmt.Println(`No port name or port value was specified, all Kubernetes service ports will be impacted`)
```

cli/chaosli/cmd/create.go
```diff
@@ -527,29 +527,39 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
 
 		hasPorts := confirmOption("Do you want to provide the ports affected by the disruption? (In case of no port defined, all ports of the service will be affected)", "")
 
-		ports := []v1beta1.NetworkDisruptionServicePortSpec{}
-
 		if hasPorts {
-			getServicePort := func() v1beta1.NetworkDisruptionServicePortSpec {
+			ports := []v1beta1.NetworkDisruptionServicePortSpec{}
+
+			getServicePort := func() *v1beta1.NetworkDisruptionServicePortSpec {
```
luphaz, line 533:
```suggestion
			getServicePort := func() v1beta1.NetworkDisruptionServicePortSpec {
```

cli/chaosli/cmd/create.go
```diff
@@ -527,29 +527,39 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
 
-				port.Port, _ = strconv.Atoi(getInput("What port would you like to target?", "", survey.WithValidator(integerValidator), survey.WithValidator(survey.Required)))
+				if port.Name == "" && port.Port == 0 {
+					fmt.Println(`No port name or port value was specified.`)
 
-				return port
+					return nil
+				}
+
+				return &port
```
luphaz, line 552:
```suggestion
				return port
```

cli/chaosli/cmd/create.go
```diff
@@ -527,29 +527,39 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
 			}
 
-			ports = append(ports, getServicePort())
+			if port := getServicePort(); port != nil {
+				ports = append(ports, *port)
```
luphaz, line 556:
```suggestion
				ports = append(ports, port)
```

cli/chaosli/cmd/create.go
```diff
@@ -527,29 +527,39 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
-		ports := []v1beta1.NetworkDisruptionServicePortSpec{}
-
 		if hasPorts {
-			getServicePort := func() v1beta1.NetworkDisruptionServicePortSpec {
+			ports := []v1beta1.NetworkDisruptionServicePortSpec{}
+
+			getServicePort := func() *v1beta1.NetworkDisruptionServicePortSpec {
 				port := v1beta1.NetworkDisruptionServicePortSpec{}
 
+				fmt.Println(`We can identify a service port by name or by port value. At least one of those fields are required.`)
```
luphaz, line 536:
```suggestion
				fmt.Println(`We can identify a service port by name or by port value. Providing on of those fields is required to limit impact on specified port.`)
```

luphaz (approved):

api/v1beta1/network_disruption.go
clairecng: I feel like being strict in the way we find things will eliminate possible user errors, so I'd like the user provided fields to be exact. If the user is misnaming their servicePort I don't want them to be able to use this port in case it's actually the other one they wanted to use. I'm not that set on it though, if you strongly believe it would be best to allow the user to disrupt the provided port even if the provided name does not match then I can change the behavior.

luphaz: With the documentation you added it's pretty clear and this behaviour seems right to me.

Thanks for the explanation, it definitely makes sense for me also now 🤗

cli/chaosli/cmd/create.go
clairecng: Here the user didn't provide any port or value where they should have. To my understanding this case would only happen if the user made a mistake while responding to the questions, but the user can still add another service port after this, so not all ports would be impacted here. 
We could however interpret this lack of response to signify that they don't want any service ports and the message would be relevant here. WDYT?

cli/chaosli/cmd/create.go
clairecng: using a pointer to be able to detect if there is a port provided or not, I'm not sure to understand why I should not do that. Could you provide more context to your suggested change?

cli/chaosli/cmd/create.go
clairecng: Yes, definitely

cli/chaosli/cmd/create.go
luphaz: 🤔 I would continue your previous reasoning about being strict.
If it chooses to provide a port or name and don't, he should receive an error message and be re-asked (or able to exit).

Maybe we could clarify in this message how to get back to not select any port.

```suggestion
					fmt.Println(`No port name or port value provided, please provide a port/name or hit CTRL^C to skip this filtering and impact all ports.`)
```

clairecng: This was my error message but yes it's probably not enough. Thanks!

cli/chaosli/cmd/create.go
luphaz: TLDR: when you have a struct, using a pointer or not should be consistent among all calls (receiver methods, returned values, etc)

In depth explanation here:
* https://dd.slack.com/archives/G01PMMNMHEZ/p1680018175024769

Here, if you concern is to validate is it empty or not you can check the returned value against an empty struct (non nil).

cli/chaosli/cmd/create.go
```diff
@@ -525,6 +525,46 @@ func getServices() []v1beta1.NetworkDisruptionServiceSpec {
...
 
+		hasPorts := confirmOption("Do you want to provide the ports affected by the disruption? (In case of no port defined, all ports of the service will be affected)", "")
+
+		if hasPorts {
+			ports := []v1beta1.NetworkDisruptionServicePortSpec{}
+
+			getServicePort := func() v1beta1.NetworkDisruptionServicePortSpec {
+				port := v1beta1.NetworkDisruptionServicePortSpec{}
+
+				fmt.Println(`We can identify a service port by name or by port value. Providing on of those fields is required to limit impact on specified port.`)
```
ptnapoleon, line 536:
```suggestion
				fmt.Println(`We can identify a service port by name or by port value. Providing one of those fields is required to limit impact on specified port.`)
```

docs/network_disruption/hosts-and-services.md
ptnapoleon, line 1: If you're renaming this file, we link to it in three places throughout docs that need to be updated

ptnapoleon (approved):
There are some broken doc links that need to be fixed, and two typos in chaosli. Approving so you can still merge after fixing those

ptnapoleon (approved):

taihuynh167 (approved):

PR #681, opened 2023-04-03
CHAOS-489: Chaos pod watcher + refacto watcher

aymericDD:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- Create the chaos pods watcher. It sends notifications if chaos pods are failing.
- New watcher architecture to facilitate the maintains and the creation of new watchers:

chart

## Code Quality Checklist

## Testing

    - [x] locally.


controllers/watchers/chaos_pod.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+// OnAdd is a handler function for the add event of the controller.
+func (w ChaosPodHandler) OnAdd(_ interface{}) {
+	// Do nothing on add event
+}
+
+// OnUpdate is a handler function for the update event of the controller.
+func (w ChaosPodHandler) OnUpdate(oldObj, newObj interface{}) {
```
aymericDD, line 47: This function is really simple. It sends notification if the pod is failing during the update.

The notification is not sent if: 
- The phase of the pod is not the same as before to avoid spamming.
- The previous phase of the pod is not in running.

If we want to catch all failures cases, we have to check the status of the chaos pod containers and its events and the events of the disruption to. It will complexify a lot the code. To know more you can refer to the [DisruptionTargetHanlder](https://github.com/DataDog/chaos-controller/pull/681/files#diff-ee1090926ca1d28cf941203c3d7e4cb7dcc035ff38a510f99b5b4876455d9699R60).

api/v1beta1/events.go
```diff
@@ -27,30 +29,30 @@ type DisruptionEvent struct {
 	OnTargetTemplateMessage        string                  // Template message to attach to the target resource (pod or node). Empty if the event should not be sent to a target (DisruptEvent only)
 	OnDisruptionTemplateMessage    string                  // We want to separate the aggregated message from the single message to include more info in the single message
 	OnDisruptionTemplateAggMessage string                  // Template message to attach to the disruption. Empty if the event should not be sent on a disruption
-	Category                       DisruptionEventCategory // Either TargetEvent or DisruptionEvent
+	Category                       DisruptionEventCategory // Either TargetEvent or DisruptionEvent or ChaosPodEvent
```
clairecng, line 32:
```suggestion
	Category                       DisruptionEventCategory // Either TargetEvent, DisruptionEvent or ChaosPodEvent
```

controllers/watchers/chaos_pod.go
```diff
@@ -0,0 +1,137 @@
...
+	oldPod, _ := oldObj.(*corev1.Pod)
+	newPod, _ := newObj.(*corev1.Pod)
+
+	// If the old and new phase are the same, do nothing
+	if oldPod.Status.Phase == newPod.Status.Phase {
+		return
+	}
+
+	// If the old pod is running and the new one is failed, do nothing
+	// The disruption is full injected if the pod has been in a running phase.
```
clairecng, line 58:
```suggestion
	// The disruption is fully injected if the pod is in a running phase.
	// The pod goes from running to failed when the finalizer is deleted, snd we don't want to alert on that.
```

controllers/watchers/chaos_pod.go
```diff
@@ -0,0 +1,137 @@
...
+		newPod.Status.Phase,
+		newPod.Status.Reason,
+	)
+
+	// Send event to notify user
+	w.recorder.Event(w.disruption, eventType, eventReason, eventMessage)
+
+	w.log.Debugw("ChaosPodHandler UPDATE - Send event",
+		"eventMessage", eventMessage,
+		"eventType", eventType,
```
clairecng, line 93: would be worth it to add the disruption name and namespace, and the chaos pod name as well here!

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,520 @@
...
+	MetricsSink    metrics.Sink
+	Log            *zap.SugaredLogger
+}
+
+// On new target
+func (d DisruptionTargetHandler) OnAdd(obj interface{}) {
+	pod, okPod := obj.(*v1.Pod)
+	node, okNode := obj.(*v1.Node)
+
+	d.Log.Debugw("DisruptionTargetHandler ADD")
```
clairecng, line 44: Should we add name/namespace of the target and disruption name/namespace?

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,520 @@
...
+
+	d.OnChangeHandleMetricsSink(pod, node, okPod, okNode)
+}
+
+// On target deletion
+func (d DisruptionTargetHandler) OnDelete(obj interface{}) {
+	pod, okPod := obj.(*v1.Pod)
+	node, okNode := obj.(*v1.Node)
+
+	d.Log.Debugw("DisruptionTargetHandler DELETE")
```
clairecng, line 54: Should we add name/namespace of the target and disruption name/namespace?

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,520 @@
...
+}
+
+// On updated target
+func (d DisruptionTargetHandler) OnUpdate(oldObj, newObj interface{}) {
+	oldPod, okOldPod := oldObj.(*v1.Pod)
+	newPod, okNewPod := newObj.(*v1.Pod)
+	oldNode, okOldNode := oldObj.(*v1.Node)
+	newNode, okNewNode := newObj.(*v1.Node)
+
+	d.Log.Debugw("DisruptionTargetHandler UPDATE")
```
clairecng, line 66: Should we add name/namespace of the target and disruption name/namespace?

controllers/watchers/target_pod.go
aymericDD: Normally the disruption name /namespace is already injected [here](https://github.com/DataDog/chaos-controller/blob/main/controllers/disruption_controller.go#L101), but I think, it is a good idea to override the value here  to ensure the disruption name and namespace are valid in case the watcher is used by another package.

api/v1beta1/events.go
```diff
@@ -228,13 +234,13 @@ var Events = map[string]DisruptionEvent{
...
 		Reason:                      EventDisruptionCreated,
-		OnDisruptionTemplateMessage: "Disruption created",
+		OnDisruptionTemplateMessage: "disruption created",
 		Category:                    DisruptEvent,
 	},
 	EventDisruptionFinished: {
 		Type:                        corev1.EventTypeNormal,
 		Reason:                      EventDisruptionFinished,
-		OnDisruptionTemplateMessage: "Disruption finished",
+		OnDisruptionTemplateMessage: "disruption finished",
```
ptnapoleon, line 243: Why lower case these? that isn't consistent throughout the file, afaict

controllers/disruption_controller.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DataDog/chaos-controller/controllers/watchers"
```
ptnapoleon, line 17: nit: not sorted with the rest of the same project imports

controllers/watcher_handler.go
```diff
@@ -0,0 +1,96 @@
...
+var watcherList = map[string]newWatcher{
+	"chaos-pod": func(name string, r *DisruptionReconciler, instance *chaosv1beta1.Disruption) (watchers.Watcher, error) {
+		return watchers.NewChaosPodWatcher(name, r.Recorder, instance, r.log)
+	},
+	"disruption-target": func(name string, r *DisruptionReconciler, instance *chaosv1beta1.Disruption) (watchers.Watcher, error) {
+		return watchers.NewDisruptionTargetWatcher(name, r.Recorder, r.Reader, r.EnableObserver, instance, r.MetricsSink, r.log)
+	},
+}
+
+// manageInstanceWatchers manage watcher instance
```
ptnapoleon, line 27:
```suggestion
// manageInstanceWatchers manages the watchers for each disruption instance
```

?

controllers/watchers/chaos_pod.go
```diff
@@ -0,0 +1,140 @@
...
+		log:        l,
+	}
+}
+
+// OnAdd is a handler function for the add event of the controller.
+func (w ChaosPodHandler) OnAdd(_ interface{}) {
+	// Do nothing on add event
+}
+
+// OnUpdate is a handler function for the update event of the controller.
```
ptnapoleon, line 46: These godocs say the `__ event of the controller`, but what controller is that referring to?

controllers/watchers/chaos_pod.go
```diff
@@ -0,0 +1,140 @@
...
+func (w ChaosPodHandler) OnUpdate(oldObj, newObj interface{}) {
+	// Convert oldObj and newObj to Pod objects
+	oldPod, _ := oldObj.(*corev1.Pod)
+	newPod, _ := newObj.(*corev1.Pod)
```
ptnapoleon, line 50: Should we check these ignored `ok` values, and error if they're false?

controllers/watchers/chaos_pod.go
ptnapoleon: Okay, I see that its the events sent by the pod Informer used by the chaos-controller. I think ending these with "of the controller" could be a bit confusing, as these events are really "of the chaos pod"

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+type DisruptionTargetHandler struct {
+	Recorder       record.EventRecorder
+	Reader         client.Reader
+	EnableObserver bool
+	Disruption     *v1beta1.Disruption
+	MetricsSink    metrics.Sink
+	Log            *zap.SugaredLogger
+}
+
+// On new target
```
ptnapoleon, line 39: nit: godoc comments should start with the method name

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	k8scache "sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// DisruptionTargetHandler struct used to manage what to do when changes occur on the watched objects in the cache
+type DisruptionTargetHandler struct {
+	Recorder       record.EventRecorder
```
ptnapoleon, line 31: Why are all of these members public? For `watchers/chaos_pod.go#ChaosPodHandler`, all the members are private. I don't see any of these accessed outside of this file.

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+
+		disruptionEvents, err := d.getEventsFromCurrentDisruption("Node", newNode.ObjectMeta, d.Disruption.CreationTimestamp.Time)
+		if err != nil {
+			d.Log.Warnf("couldn't get the list of events from the target. Might not be able to notify on error changes: %s", err.Error())
+		}
+
+		// we detect and compute the error / warning events, status changes, conditions of the updated node
+		eventsToSend = d.buildNodeEventsToSend(*oldNode, *newNode, disruptionEvents)
+	default:
+		d.Log.Debugw("target observer couldn't detect what type of changes happened on the targets")
```
ptnapoleon, line 140:
```suggestion
		d.Log.Warnw("target observer couldn't detect what type of changes happened on the targets")
```

At least Warn, if not Error?

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+
+					switch {
+					case strings.Contains(lowerCasedMessage, "liveness probe"):
+						eventsToSend[v1beta1.EventTargetLivenessProbeChange] = true
+					case strings.Contains(lowerCasedMessage, "readiness probe"):
+						// If the object of the disruption is in the list of targets, it means it has been injected.
+						// The readiness probe is failing during the injection
+						if d.Disruption.Status.HasTarget(event.InvolvedObject.Name) {
+							eventsToSend[v1beta1.EventTargetReadinessProbeChangeDuringDisruption] = true
+						} else {
```
ptnapoleon, line 261: Is this `else` guaranteed to be true? Could it be the case that we no longer have the Target in our Targets list, but we're still getting events about it from the watcher? If this watcher is for our targets, how are we getting an event about this pod but it isn't in our Targets list?

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+func (d DisruptionTargetHandler) buildPodEventsToSend(oldPod v1.Pod, newPod v1.Pod, disruptionEvents []v1.Event) map[string]bool {
+	var recoverTimestamp *time.Time // keep track of the timestamp of a recovering event / state
+
+	eventsToSend := make(map[string]bool)
+	runningState := "Running"
+
+	// compare statuses between old and new pod to detect changes
+	for _, container := range newPod.Status.ContainerStatuses {
+		for _, oldContainer := range oldPod.Status.ContainerStatuses {
+			if container.Name != oldContainer.Name { // When restarting, container can change of ID, so we need to verify by name
```
ptnapoleon, line 334: It's also possible for the name to change, but I think that's rare?

controllers/watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+				continue
+			}
+
+			// Warning events
+			if container.RestartCount > (oldContainer.RestartCount + 2) {
```
ptnapoleon, line 339: Can you leave a comment explaining why `+ 2` here?

controllers/watchers/watcher_test.go
```diff
@@ -0,0 +1,316 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023 Datadog, Inc.
+
+package watchers_test
+
+import (
+	"context"
```
ptnapoleon, line 9: nit: could you sort the imports with goimports?

controllers/watchers/watcher_test.go
```diff
@@ -0,0 +1,316 @@
...
+	namespaceName := types.NamespacedName{Name: "namespace-name", Namespace: "namespace"}
+
+	BeforeEach(func() {
+		targetObjectType = &v1.Pod{}
+	})
+
+	JustBeforeEach(func() {
+		// Arrange
+		log, err := logger.NewZapLogger()
+		Expect(err).ShouldNot(HaveOccurred())
```
ptnapoleon, line 46: Why not create a single logger in test setup, instead of one per test case?

controllers/watchers/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+			return fmt.Errorf("error creating cache: %s", err)
+		}
+
+		w.cache = cache
+	}
+
+	// get informer from cache
+	info, err := w.cache.GetInformer(context.Background(), w.config.ObjectType)
+	if err != nil {
+		return fmt.Errorf("error getting informer form cache. Error: %s", err)
```
ptnapoleon, line 169:
```suggestion
		return fmt.Errorf("error getting informer from cache. Error: %s", err)
```

controllers/watchers/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// add event handler to informer
+	_, err = info.AddEventHandler(w.config.Handler)
+	if err != nil {
+		return fmt.Errorf("error adding event handler to the informer. Error: %s", err)
```
ptnapoleon, line 175:
```suggestion
		return fmt.Errorf("error adding event handler to the informer. Error: %w", err)
```

controllers/watchers/target_pod_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/DataDog/chaos-controller/metrics/noop"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+var _ = Describe("target pod", func() {
+	Describe("Handler", func() {
+		//	@ToDo
```
ptnapoleon, line 22: Still a TODO?

controllers/watchers/target_pod_test.go
```diff
@@ -0,0 +1,77 @@
...
+			watcher, err = watchers.NewDisruptionTargetWatcher(watcherName, &mocks.EventRecorderMock{}, &mocks.ReaderMock{}, true, &disruption, &noop.Sink{}, logger)
+		})
+
+		Context("with a valid disruption", func() {
+			BeforeEach(func() {
+				// Arrange
+				disruption = chaosv1beta1.Disruption{
+					Spec: chaosv1beta1.DisruptionSpec{
+						Selector: labels.Set{
+							"Lorem": "ispum",
```
ptnapoleon, line 50:
```suggestion
							"Lorem": "ipsum",
```

ptnapoleon (commented):
One thing to note, is that because they're in the controllers package, right now all of the new tests run in `make e2e-tests` instead of `make test`. Given they're all mocked unit tests, that's suboptimal. Could you edit the `test:` and `e2e-test:` targets in the `Makefile` to change where these are included?

api/v1beta1/events.go
aymericDD: Good catch! It is a typo' error.

controllers/watchers/chaos_pod.go
aymericDD: Yes we can, but if we encounter it is not normal at all. Chaos pods Handler should listen pods, and that's all.

aymericDD:
> One thing to note, is that because they're in the controllers package, right now all of the new tests run in `make e2e-tests` instead of `make test`. Given they're all mocked unit tests, that's suboptimal. Could you edit the `test:` and `e2e-test:` targets in the `Makefile` to change where these are included?

Alternatively, I could relocate the 'watchers' package to the root directory of the project. What do you think @ptnapoleon ?

controllers/watchers/target_pod_test.go
aymericDD: Indeed! 
IMO: this watcher is really complex and it will require a lot of reverse engineering to cover all the cases... To implement the unit tests it could more faster to pair with @clairecng and/or @nathantournant

ptnapoleon:
> > One thing to note, is that because they're in the controllers package, right now all of the new tests run in `make e2e-tests` instead of `make test`. Given they're all mocked unit tests, that's suboptimal. Could you edit the `test:` and `e2e-test:` targets in the `Makefile` to change where these are included?
> 
> Alternatively, I could relocate the 'watchers' package to the root directory of the project. What do you think @ptnapoleon ?

That works just as well, and absolutely solves the problem.

Fantastic PR!

controllers/watchers/target_pod.go
aymericDD: Yes, it is not normal if the default condition is true.

controllers/watchers/target_pod.go
aymericDD: > Could it be the case that we no longer have the Target in our Targets list, but we're still getting events about it from the watcher?
To be honest, I don't know... I just move the **DisruptionTargetHandler** form the cache_handler.go...

controllers/watchers/target_pod.go
aymericDD: FAK: 
- Deployment: If the name of the container is updated a new replicaset is created and a new pod is scheduled. If we have a new pod, is it considered as a new target? If yes, I think it will not be a problem.
- Node: I don't know the behavior.

controllers/watchers/watcher_test.go
aymericDD: Of course! It has no sense to be repeated each time.

controllers/watchers/target_pod.go
ptnapoleon: If it isn't a change in behavior, it's fine if its wrong.

watchers/watcher_test.go
```diff
@@ -0,0 +1,317 @@
...
+		cacheMock        mocks.CacheMock
+		informerMock     mocks.InformerMock
+		handlerMock      mocks.HandlerMock
+		targetObjectType client.Object
+		err              error
+		cacheContextFunc func() (ctx context.Context, cancel context.CancelFunc)
+	)
+
+	// Arrange
+	namespaceName := types.NamespacedName{Name: "namespace-name", Namespace: "namespace"}
```
ptnapoleon, line 39: absolutely ridiculous nitpick from me, but using "namespace-name" as the name in one test, but as the namespace in another test is confusing

watchers/target_pod.go
```diff
@@ -0,0 +1,563 @@
...
+	"strings"
+	"time"
+
+	"github.com/DataDog/chaos-controller/api/v1beta1"
+	"github.com/DataDog/chaos-controller/metrics"
+	"github.com/DataDog/chaos-controller/targetselector"
+	chaostypes "github.com/DataDog/chaos-controller/types"
+	"go.uber.org/zap"
+	v1 "k8s.io/api/core/v1"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
ptnapoleon, line 21: Any chance you could change these two aliases to `corev1` and `metav1`, respectively? It would make it consistent with other uses in the codebase

controllers/watcher_handler.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if err := r.WatcherManager.RemoveExpiredWatchers(); err != nil {
+		return err
+	}
+
+	for wName, newWatcherFunc := range watcherList {
+		watcherName, err := getInstanceWatcherName(wName, instance)
+		if err != nil {
+			return fmt.Errorf("failed to initialize %s watcher of the disruption %s: could not get the watcher name. Error: %s", wName, instance.Name, err.Error())
```
ptnapoleon, line 42: I believe that errors should always be interpolated with `%w`, https://go.dev/blog/go1.13-errors
```suggestion
			return fmt.Errorf("failed to initialize %s watcher of the disruption %s: could not get the watcher name. Error: %w", wName, instance.Name, err)
```

ptnapoleon (approved):
Other than two nits, looks great!! I do think we should let this bake in staging to confirm no mem leaks, but very readable, tests are well laid out, great improvement

watchers/manager.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		m.deleteOrphanWatcher(w)
+	}
+
+	return nil
+}
+
+// deleteOrphanWatcher deletes a Watcher instance if the associated Disruption object does not exist
+func (m manager) deleteOrphanWatcher(watcher Watcher) bool {
```
luphaz, line 132: This feels to me it may not be the same thing as expired watchers

Why hiding this method instead of providing it in the interface and being able to schedule both separately?

watchers/mocks/cache.go
luphaz, line 1: How did you created all those mocks?

You might want to use `go:generate` func and use mockery to do it for you.

Example here:
* https://github.com/DataDog/chaos-controller/blob/8d4a392eb786c80b6dc8669f388b514bd08fce45/eventnotifier/slack/slack.go#L30

Then you will be able to use the `New...` method and use [EXPECT](https://github.com/DataDog/chaos-controller/blob/8d4a392eb786c80b6dc8669f388b514bd08fce45/eventnotifier/slack/slack_test.go#L125) that is fluent and nice to use from my POV

To generate mock once `go:generate` is added, use `make generate-mock`

watchers/target_pod_test.go
```diff
@@ -0,0 +1,77 @@
...
+		var (
+			disruption chaosv1beta1.Disruption
+			err        error
+			watcher    watchers.Watcher
+		)
+
+		JustBeforeEach(func() {
+			// Arrange
+			var logger *zap.SugaredLogger
+			logger, err = log.NewZapLogger()
```
luphaz, line 37: FYI if you need to "test" logs you can use this:
* https://github.com/DataDog/chaos-controller/blob/8d4a392eb786c80b6dc8669f388b514bd08fce45/injector/python_test.go#L117

You may also want to use instead this:
* https://pkg.go.dev/go.uber.org/zap/zaptest

That creates a logger for testing.

watchers/target_pod_test.go
```diff
@@ -0,0 +1,77 @@
...
+						Selector: labels.Set{
+							"Lorem": "ipsum",
+						},
+					},
+				}
+			})
+
+			It("should not return an error", func() {
+				Expect(err).ShouldNot(HaveOccurred())
+			})
```
luphaz, line 58: FYI should succeed also exists (and does the same)
```suggestion
			It("should succeed", func() {
				Expect(err).Should(Succeed())
			})
```

watchers/chaos_pod_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+		Describe("on Add", func() {
+			var newPod v1.Pod
+
+			JustBeforeEach(func() {
+				// Action
+				chaosPodHandler.OnAdd(&newPod)
+			})
+
+			It("should to nothing", func() {})
```
luphaz, line 56:
```suggestion
			It("should do nothing", func() {})
```

luphaz: 🤔 not sure it worth to have this if there is no expectations associated

watchers/chaos_pod_test.go
```diff
@@ -0,0 +1,243 @@
...
+						expectEventMessage := fmt.Sprintf(chaosv1beta1.Events[chaosv1beta1.EventChaosPodFailedState].OnDisruptionTemplateMessage,
+							newPod,
+							v1.PodFailed,
+							"OutOfpods",
+						)
+
+						Eventually(eventRecorder.(*record.FakeRecorder).Events).Should(Receive(ContainSubstring(expectEventType), ContainSubstring(expectEventMessage)))
+					})
+
+					It("should not send two times the event", func() {
```
luphaz, line 105:
```suggestion
					It("should send the event once", func() {
```

watchers/manager_test.go
```diff
@@ -0,0 +1,465 @@
...
+	"github.com/stretchr/testify/mock"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"net/http"
+)
+
+var _ = Describe("Manager of watchers", func() {
+	// Watcher methods
+	const methodWatcherStart string = "Start"
```
luphaz, line 25: Use mockery and Expect and all this consts will disappear 🪄 🤗

watchers/chaos_pod_test.go
aymericDD: It just to be explicit about it, do nothing and cover implementation. I practice TDD, so I often defined my unit test before the implementation.

controllers/watcher_handler.go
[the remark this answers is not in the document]
aymericDD: Indeed, it is a good idea 👍

watchers/watcher_test.go
aymericDD: Indeed, it's often a good idea to be consistent 👍

controllers/watcher_handler.go
aymericDD: Thank you for the source! Really interesting, I understand now why it is interesting to use `%w` 👍

aymericDD: After reading the article and in particular this sentence :
image

Should I have to use the `%w` into the watcher implementation? @ptnapoleon 

IMO: the watcher is an abstraction layer, so if I understand the article it is not recommended to wrap error.

controllers/watcher_handler.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		if w := r.WatcherManager.GetWatcher(watcherName); w != nil {
+			continue
+		}
+
+		watcherInstance, err := newWatcherFunc(watcherName, r, instance)
+		if err != nil {
+			r.log.Errorf("failed to initialize the %s watcher of the %s disruption. Error: %s", wName, instance.Name, err.Error())
+			return err
```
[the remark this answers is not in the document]
aymericDD: 🙈 Oupsy... I am totally agree with you 👍

controllers/watcher_handler.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if len(watcherList) == 0 {
+		return nil
+	}
+
+	if err := r.WatcherManager.RemoveExpiredWatchers(); err != nil {
+		return err
+	}
+
+	for wName, newWatcherFunc := range watcherList {
```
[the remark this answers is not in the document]
aymericDD: If a factory is used instead, this code becomes obsolete.

controllers/watcher_handler.go
```diff
@@ -0,0 +1,96 @@
...
+
+// manageInstanceWatchers manage the watchers for each disruption instance
+func (r *DisruptionReconciler) manageInstanceWatchers(instance *chaosv1beta1.Disruption) error {
+	r.log.Debugf("start to manage watchers for the %s disruption", instance.Name)
+
+	if len(watcherList) == 0 {
+		return nil
+	}
+
+	if err := r.WatcherManager.RemoveExpiredWatchers(); err != nil {
```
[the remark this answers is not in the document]
luphaz: For me it's an issue to have this on the critical path of creating a new cache handler.
If this returns an error, you can't create any cache anymore.

This feels counter intuitive.

Let's see how it will look like with the suggested factory change and we will see.

controllers/watcher_handler.go
```diff
@@ -0,0 +1,96 @@
...
+			if err := r.WatcherManager.RemoveWatcher(w); err != nil {
+				r.log.Errorf("could not remove the %s watcher of the %s disruption: Error: %s", wName, instance.Name, err.Error())
+			}
+
+			r.log.Debugf("the %s watcher of the %s disruption is removed", wName, instance.Name)
+		}
+	}
+}
+
+func getInstanceWatcherName(watcherPrefixName string, instance *chaosv1beta1.Disruption) (string, error) {
```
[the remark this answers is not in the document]
aymericDD: Currently, the `WatcherManager` just manage `Watcher` instances, but it does not have a notion of `Disruption`. The factory should take in argument a disruption and create a watcher so it could handle the naming of the watcher based on the disruption. I agree with you it will be simpler to have a `map[string][]Watcher` inside the WatcherManager, but it should be renamed `DisuptionsWatchersManger` or another name _i am not comfortable to name things..._. I will think about a new implementation, but It will introduce more changes.

watchers/chaos_pod.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// If the old and new phase are the same, do nothing
+	if oldPod.Status.Phase == newPod.Status.Phase {
+		return
+	}
+
+	// If the old pod is running and the new one is failed, do nothing
+	// The disruption is fully injected if the pod is in a running phase.
+	if oldPod.Status.Phase == corev1.PodRunning && newPod.Status.Phase == corev1.PodFailed {
```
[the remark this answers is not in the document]
aymericDD: This watcher does not cover all cases. We could improve the watcher in another PR to be sure to handle all cases, but it already cover cases like **OutOfPods, ImagePullBackOff** etc etc...
I conducted a CPU pressure test and observed that the injection status was activated immediately upon running the chaos pod.

watchers/mocks/cache.go
aymericDD: So nice! I have to test this!

watchers/target_pod_test.go
aymericDD: For me, `Expect(err).Should(Succeed())` means the error has been raised. I prefer `ShouldNot(HaveOccurred())`, for my it is less ambiguous.

controllers/watcher_handler.go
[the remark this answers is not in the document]
aymericDD: After investigation, the method `RemoveExpiredWatchers` of the `WatcherManager` does not return an error. So I have removed the return error and now it is not blocker anymore.

watchers/manager.go
aymericDD: Good point! I will change that to 👍

controllers/watcher_handler.go
ptnapoleon: Given this is all internal to us, and the logs are all read by us, I think it's good to pierce the abstraction layer within the logs

watchers/target_pod_test.go
luphaz: Yeah, with the error here it's definitely misleading, this helper has been introduced when you have the method call into the expectation:

```golang
Expect(MyMethodReturnOnlyErrorOrNil()).Should(Succeed())
```

Here it's readable, with only `err` it's definitely not.

Thanks for challenging 🤗

api/v1beta1/events.go
```diff
@@ -19,6 +19,8 @@ const (
 	TargetEvent DisruptionEventCategory = "TargetEvent"
 	// Event only attached to the disruption
 	DisruptEvent DisruptionEventCategory = "DisruptionEvent"
+	// Event only attached to an injection
```
clairecng, line 22:
```suggestion
	// Event only attached to a chaos pod
```

watchers/chaos_pod_test.go
```diff
@@ -0,0 +1,243 @@
...
+	oldPod = v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "oldPod"},
+		Status:     oldPodStatus,
+	}
+	newPod = v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "newPod"},
+		Status:     newPodStatus,
+	}
+	return
+}
```
clairecng, line 243: Thanks a lot for all of those tests Aymeric 👍

watchers/target_pod_test.go
aymericDD: I agree with `Expect(MyMethodReturnOnlyErrorOrNil()).Should(Succeed())` because the action should succeed in a nominal case. Thanks for your comprehension :bow:

controllers/watcher_handler.go
aymericDD: It makes sense! I will adapt all the logs 👍

watchers/cache_mock_test.go
```diff
@@ -0,0 +1,71 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023 Datadog, Inc.
+
+package watchers_test
```
ptnapoleon, line 6: What's going on with this file? It isn't labeled as autogenerated, but it seems like it is. It's named `_test`, but it seems to just be a mock?

controllers/disruption_controller.go
```diff
@@ -153,8 +155,14 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	if err := r.manageInstanceSelectorCache(instance); err != nil {
-		r.log.Errorw("error managing selector cache", "error", err)
+	if err := r.DisruptionsWatchersManager.RemoveAllOrphanWatchers(); err != nil {
+		r.log.Errorw("error during the deletion of orphan watchers", "error", err)
+	}
+
+	r.DisruptionsWatchersManager.RemoveAllExpiredWatchers()
```
ptnapoleon, line 162: Matthieu mentioned this in the last review, but even though it doesn't return an error anymore, I think we can move this work out of the reconcile loop and into async code. The check for expired and orphaned watchers aren't specific to a given instance, and I don't think we need to run them everytime we reconcile. wdyt?

watchers/watcher.go
```diff
@@ -0,0 +1,191 @@
...
+	ObjectType client.Object
+}
+
+// CtxTuple is a struct that holds a context and its cancel function, as well as a NamespacedName that identifies a Disruption resource.
+type CtxTuple struct {
+	// CancelFunc is a function that can be called to cancel the associated context.
+	CancelFunc context.CancelFunc
+
+	// Ctx is a context.Context object that is used to manage the lifetime of an operation.
+	Ctx context.Context
```
ptnapoleon, line 69: Seconding Matthieu, can't we store the cancelFunc and the namespacedname inside the context, using `context.WithValue` and `ctx.Value` functions??

watchers/target_pod.go
```diff
@@ -0,0 +1,506 @@
...
+
+	// remove recovering event if node has another warning condition / phase
+	if eventsToSend[v1beta1.EventTargetNodeRecoveredState] && len(eventsToSend) > 1 {
+		eventsToSend[v1beta1.EventTargetNodeRecoveredState] = false
+		recoverTimestamp = nil
+	}
+
+	eventsToSend, cannotRecoverYet := d.findNotifiableEvents(eventsToSend, targetEvents, recoverTimestamp, newNode.Name)
+
+	// if other warning events has been detected, the target hasn't recovered
```
ptnapoleon, line 474:
```suggestion
	// if other warning events have been detected, the target hasn't recovered
```

watchers/target_pod.go
```diff
@@ -0,0 +1,506 @@
...
+
+	// remove recovering event if pod has another warning container state
+	if eventsToSend[v1beta1.EventTargetPodRecoveredState] && len(eventsToSend) > 1 {
+		eventsToSend[v1beta1.EventTargetPodRecoveredState] = false
+		recoverTimestamp = nil
+	}
+
+	eventsToSend, cannotRecoverYet := d.findNotifiableEvents(eventsToSend, disruptionEvents, recoverTimestamp, fmt.Sprintf("%s/%s", newPod.Namespace, newPod.Name))
+
+	// if other warning events has been detected, the target hasn't recovered
```
ptnapoleon, line 383:
```suggestion
	// if other warning events have been detected, the target hasn't recovered
```

watchers/disruptions_watchers_manager.go
```diff
@@ -0,0 +1,217 @@
...
+		if ok := watcherManager.GetWatcher(watcherNameHash); ok != nil {
+			continue
+		}
+
+		// Otherwise add the new watcher for the disruption
+		if err := d.addWatcher(disruption, watcherName, watcherNameHash, cacheMock, watcherManager); err != nil {
+			return err
+		}
+
+		d.log.Infof("Watcher %s created for disruption %s with %s namespace", watcherName, disruption.Name, disruption.Namespace)
```
ptnapoleon, line 102: This should be using the `disruptionName` and `disruptionNamespace` tags, rather than embedding them in the string. I also think it should be ok for this to be at debug?

watchers/disruptions_watchers_manager.go
```diff
@@ -0,0 +1,217 @@
...
+	if watcherManager != nil {
+		watcherManager.RemoveAllWatchers()
+
+		// Remove the Watcher Manager from the map.
+		delete(d.watchersManagers, namespacedName)
+
+		d.log.Infof("all watchers have been removed for the %s disruption with %s namespace", disruption.Name, disruption.Namespace)
+	}
+
+	d.log.Debugw(fmt.Sprintf("could not remove all watchers for the %s disruption with %s namespace", disruption.Name, disruption.Namespace), "disruption", disruption, "namespacedName", namespacedName)
```
ptnapoleon, line 125: We don't want to dump the entire disruption struct as a log tag. 
```suggestion
	d.log.Debugw(fmt.Sprintf("could not remove all watchers for the %s disruption with %s namespace", disruption.Name, disruption.Namespace), "disruptionName", disruption.Name, "disruptionNamespace", namespacedName.Namespace)
```

watchers/disruptions_watchers_manager.go
ptnapoleon: ditto on the other places in this file where we use Infof instead of Infow

watchers/chaos_pod.go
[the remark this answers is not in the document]
ptnapoleon: I don't think I follow if we settled this. Let's leave a TODO or jira ticket as a follow-up to see if we want to improve this?

ptnapoleon (commented):
I have some concerns about logging tags in the watchers, but otherwise this seems ready to merge? We can deploy to staging initially to bake for a while, this is a lot of changes at once, its hard for me to hold in my head at once. But great job adding a lot of tests

watchers/cache_mock_test.go
aymericDD: Good question 👍  It's a custom mock file which uses the 'sync.WaitGroup' to avoid concurrent memory accesses during the execution of the unit tests.

LICENSE-3rdparty.csv
```diff
@@ -1041,4 +1041,4 @@ sigs.k8s.io/structured-merge-diff/v4,sigs.k8s.io/structured-merge-diff/v4/fieldp
 sigs.k8s.io/structured-merge-diff/v4,sigs.k8s.io/structured-merge-diff/v4/schema,Apache-2.0
 sigs.k8s.io/structured-merge-diff/v4,sigs.k8s.io/structured-merge-diff/v4/typed,Apache-2.0
 sigs.k8s.io/structured-merge-diff/v4,sigs.k8s.io/structured-merge-diff/v4/value,Apache-2.0
-sigs.k8s.io/yaml,sigs.k8s.io/yaml,MIT
+sigs.k8s.io/yaml,sigs.k8s.io/yaml,Apache-2.0
```
ptnapoleon, line 1044: A bit confused, because https://github.com/kubernetes-sigs/yaml/blob/master/LICENSE shows this should be MIT, did you manually edit this, or did `make license-check` come up with this?

ptnapoleon (commented):
Why are there so many new `mock` files in the new `mocks` directory? I thought we'd moved to keeping mocks alongside the source they were generated from.

watchers/watcher_test.go
ptnapoleon: Unresolving because they're still inconsistent

watchers/factory_test.go
```diff
@@ -0,0 +1,140 @@
...
+			watcher, err = watcherFactory.NewChaosPodWatcher(watcherName, &disruption, cacheMock)
+		})
+
+		Context("with a valid disruption", func() {
+			BeforeEach(func() {
+				// Arrange
+				disruption = chaosv1beta1.Disruption{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "disruption-name",
+						Namespace: "namespace-name",
```
ptnapoleon, line 59: Here is the inconsistency

ptnapoleon (approved):
Left small questions and concerns, but I think this looks as good as I can tell with my eyes

watchers/chaos_pod_test.go
```diff
@@ -0,0 +1,181 @@
...
+			bufferSize = 10
+		)
+
+		var (
+			chaosPodHandler watchers.ChaosPodHandler
+			disruption      *v1beta1.Disruption
+			eventRecorder   record.EventRecorder
+		)
+
+		JustBeforeEach(func() {
```
luphaz, line 32: Prefer `BeforeEach` whenever possible, in particular here it's configuration/setup, not a creation.

Use `JustBeforeEach` probably only once to perform the creation and associated assertions.

https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach

watchers/disruptions_watchers_manager_test.go
```diff
@@ -0,0 +1,467 @@
...
+		disruptionsWatchersManager                                       watchers.DisruptionsWatchersManager
+		disruptionTargetWatcherMock                                      *watchers.WatcherMock
+		err                                                              error
+		expectedDisruptionTargetWatcherName, expectedChaosPodWatcherName string
+		watchersManagerMock                                              *watchers.ManagerMock
+		watcherFactoryMock                                               *watchers.FactoryMock
+		readerMock                                                       *mocks.ReaderMock
+	)
+
+	twoDisruptions := []*chaosv1beta1.Disruption{
```
luphaz, line 35: FYI if inside your setup or assertion nodes you modify those at any time you should NOT assign them here but in a `BeforeEach` node.

To avoid confusion and potential later usage and bug, you may still want to move this into a `BeforeEach` now so later contributors won't have to fix this.

Only case it's OK to declare + assign is for const IMO.

LICENSE-3rdparty.csv
aymericDD: I don't update the file manually and to be honest I don't really know why the license has changed...

.vendor.mockery.yaml
```diff
@@ -0,0 +1,56 @@
+all: True
+case: underscore
+disable-version-string: True
+filename: "{{.InterfaceNameSnake}}_mock.go"
```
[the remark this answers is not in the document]
aymericDD: For sure, no need to duplicate the mock keyword. 👍

controllers/disruption_controller.go
```diff
@@ -86,7 +88,7 @@ type CtxTuple struct {
 //+kubebuilder:rbac:groups=core,resources=nodes,verbs=list;watch
 //+kubebuilder:rbac:groups=core,resources=services,verbs=list;watch
 
-func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
+func (r *DisruptionReconciler) Reconcile(_ context.Context, req ctrl.Request) (res ctrl.Result, err error) {
```
luphaz, line 91: Usually it's worrying to me to have a context not used.
It's a code smell we are not handling things correctly IMO.

We should create a dedicated task to ask ourselves in each Kubernetes API call, should it be cancelled or not (during a deploy as an example)

And add required management.

Not in this PR, but to keep in mind...

luphaz (approved):

controllers/disruption_controller.go
ptnapoleon: Agreed, I do think we should start using this `ctx`, but that's a separate PR

ptnapoleon (approved):

aymericDD:
Thanks for all reviewers 🙏

PR #687, opened 2023-04-06
CHAOS-553: Improve logging and prevent reinjection of static node failure

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Mostly logging improvements and changes
- I've realized the issue is that with staticTargeting and a pod level disruption, we will continually shut down the nodes hosting that pod as it is rescheduled. We never update the target list after the first selection because of staticTargeting, but whats in the target list is a target pod. After the first injection, our chaos pod is gone, because the node is gone. So we skip over selectTargets, go to startInjection, and find a target w/ no chaos pod and create a new one

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


controllers/disruption_controller.go
```diff
@@ -462,6 +468,7 @@ func (r *DisruptionReconciler) createChaosPods(instance *chaosv1beta1.Disruption
 		// get IDs of targeted containers or all containers
 		targetContainers, err = utils.GetTargetedContainersInfo(&pod, instance.Spec.Containers)
 		if err != nil {
+			r.log.Debugw("unable to get target pod container ID", "targetPodStatus", pod.Status, "targetPodName", target, "targetPodNamespace", instance.Namespace)
```
ptnapoleon, line 471: The error will get logged at error in the top-level reconcile loop, but I don't want to stuff all this pod info into returns so that we can log it up there, so I'm debug logging it here. reasonable?

controllers/disruption_controller.go
```diff
@@ -493,7 +500,14 @@ func (r *DisruptionReconciler) createChaosPods(instance *chaosv1beta1.Disruption
 		// create injection pods if none have been found
 		switch len(found) {
 		case 0:
-			r.log.Infow("creating chaos pod", "target", target)
+			chaosPodArgs := []string{}
+			if len(chaosPod.Spec.Containers) > 0 && chaosPod.Spec.Containers[0].Name == "injector" {
```
ptnapoleon, line 504: i'm fairly confident this will always be true, but wanted to be safe.

controllers/disruption_controller.go
```diff
@@ -264,7 +264,7 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		// start injections
 		if err := r.startInjection(instance); err != nil {
-			r.log.Errorw("error injecting the disruption", "error", err)
+			r.log.Errorw("error creating chaos pods to start the disruption", "error", err)
 
 			return ctrl.Result{}, fmt.Errorf("error injecting the disruption: %w", err)
```
craig-seeman, line 269: Should the returned text be updated as well?

chart/values.yaml
```diff
@@ -44,7 +44,7 @@ controller:
     datadog: # datadog cloud provider config
       enabled: true # enable the provider
       ipRangesURL: "https://ip-ranges.datadoghq.com/" # URL to the IP ranges file (format must be the expected one, defaults is the public file provided by the cloud provider)
-  defaultDuration: 1h # default spec.duration for a disruption with none specified
+  defaultDuration: 5m # default spec.duration for a disruption with none specified
```
craig-seeman, line 47: 5 minutes seems a bit short for chaos testing we've been performing. Would it be possible to keep this 1h as a default from an end-user default perspective as it's referenced in the docs as such as well?

Perhaps a compromise could be to have something like a local-values.yaml for development specific defaults.

ptnapoleon: I hadn't considered that users were directly using this `values.yaml`, I'll look at adding a separate local dev values

Makefile
```diff
@@ -149,6 +149,7 @@ lima-install: manifests
 	helm template \
 		--set controller.enableSafeguards=false \
 		--set controller.expiredDisruptionGCDelay=${EXPIRED_DISRUPTION_GC_DELAY} \
+		--values ./chart/values/dev.yaml \
```
luphaz, line 152: WDYT of moving the `controller.enableSafeguards=false` to this dedicated values file?

api/v1beta1/disruption_webhook.go
```diff
@@ -95,7 +95,8 @@ var _ webhook.Validator = &Disruption{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *Disruption) ValidateCreate() error {
-	logger.Debugw("validating created disruption", "instance", r.Name, "namespace", r.Namespace)
+	logger := logger.With("instance", r.Name, "namespace", r.Namespace)
```
luphaz, line 98:
```suggestion
	logger := logger.With("instance", r.Name, "namespace", r.Namespace)
```

Should we continue to align our logging keys,`disruptionName` and `disruptionNamespace` are used in the reconcile loop and injector, could we use such keys here also?

Maybe we could have a logger wrapper/builder that return a logger from a disruption with expected keys so we spread/reuse it whenever needed.

api/v1beta1/disruption_webhook.go
```diff
@@ -95,7 +95,8 @@ var _ webhook.Validator = &Disruption{}
...
+	logger.Debugw("validating created disruption", "spec", r.Spec)
```
luphaz, line 99: ❤️ could/should we also add this to the reconcile loop or will it be "too much logs"?

controllers/disruption_controller.go
```diff
@@ -131,9 +131,9 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// handle any chaos pods being deleted (either by the disruption deletion or by an external event)
 	if err := r.handleChaosPodsTermination(instance); err != nil {
 		if isModifiedError(err) {
-			r.log.Warnw("error handling chaos pods termination", "error", err)
+			r.log.Infow("retryable error handling chaos pods termination", "error", err)
 
-			return ctrl.Result{}, nil
+			return ctrl.Result{Requeue: true}, err
```
luphaz, line 136: I think the `Requeue: true` is not needed as error is not nil:
* https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

```go
	// Reconcile performs a full reconciliation for the object referred to by the Request.
	// The Controller will requeue the Request to be processed again if an error is non-nil or
	// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```

controllers/disruption_controller.go
```diff
@@ -424,6 +436,12 @@ func (r *DisruptionReconciler) startInjection(instance *chaosv1beta1.Disruption)
 				continue
 			}
 
+			if disKind == chaostypes.DisruptionKindNodeFailure && instance.Spec.StaticTargeting && injection.InjectionStatus != chaostypes.DisruptionInjectionStatusNotInjected {
```
luphaz, line 439: Maybe we could create a small helper method to clarify the intent, something like `shouldSkipNodeFailureReinjection` or similar

controllers/disruption_controller.go
luphaz, line 1: The whole reconcile loop feels cluttered by logs from my POV, I would suggest to:
* ensure all errors are wrapped as expected
* remove all logs that are error handling logs, usually duplicating the exact same message that is inside the error wrapping (warn, error, info)
* create a defer method as the top that will handle in a consistent manner all of our errors
  * in particular now that we are no longer skipping to return an error in case of `isModifiedError`, we can safely handle errors here (in the defer) and log the context because this is what we are doing in the end

controllers/disruption_controller.go
luphaz: Shouldn't this be part of the error informations instead?
Could we create a custom error type to encapsulate error and provide contextual informations that will be logged as expected?

It feels we are always logging + erroring, that is a code smell we are missing something from my POV.

The error itself should host all informations expected to troubleshoot, or a log should be here and no error returned.

Here we are always doing both, it feels weird from my POV.

controllers/disruption_controller.go
```diff
@@ -210,9 +212,15 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		// the injection is being created or modified, apply needed actions
 		controllerutil.AddFinalizer(instance, chaostypes.DisruptionFinalizer)
 		if err := r.Update(context.Background(), instance); err != nil {
+			if isModifiedError(err) {
```
aymericDD, line 215: Does this part of the code is coverage by unit test?

controllers/disruption_controller.go
aymericDD, line 439: Does this part of the code is coverage by unit test?

aymericDD (approved):
LGTM! Good catch! 👍

controllers/disruption_controller.go
ptnapoleon: If there's a nice way to stuff the tags we want into the error, so that it can be logged at the top level, that would look cleanest, I agree.

Would that be alright in a follow-up? I primarily want to get the node failure change merged, this logging was all just in pursuit of debugging that

controllers/disruption_controller.go
ptnapoleon: Yes, that is true. It's fine to be redundant, though?

luphaz: yep it is fine indeed, I only tend to use the minimal instruction to achieve expected result, only a personal matter though

api/v1beta1/disruption_webhook.go
luphaz: 👍

controllers/disruption_controller.go
luphaz: What I was thinking about was creating a custom error type with proper contextual properties or random `map[string]string` contextual fields.

api/v1beta1/disruption_webhook.go
luphaz: It depends, are you talking about logs or metrics :trollface: 

https://github.com/DataDog/chaos-controller/blob/4cfb483c2af30fe139ccee39f1fc59b93a7d7ed2/controllers/disruption_controller.go#L101

clairecng: Metrics yes! But having everything aligned as well would be nice too 😅

luphaz: Yes definitely, and here I would complain about the fact metric tags are "lowercased" and it's more readable to me to have words separated by underscore :trollface: hence `disruption_name` and `disruption_namespace` are my favs.

controllers/disruption_controller.go
ptnapoleon: > remove all logs that are error handling logs, usually duplicating the exact same message that is inside the error wrapping (warn, error, info)

I can fix this up for all of our inner functions. But the `errors` that we're returning from reconcile are not logged by anything else, and so we definitely want our own logs that duplicate the message, right?

ptnapoleon:
Looks like the logging changes are going well, just going to address Aymeric's comments then re-open for review

controllers/disruption_controller_test.go
```diff
@@ -596,6 +596,53 @@ var _ = Describe("Disruption Controller", func() {
 		})
 	})
 
+	Context("don't reinject a static node disruption", func() {
```
ptnapoleon, line 599: This test takes almost a minute

controllers/disruption_controller.go
ptnapoleon: It is now

controllers/disruption_controller.go
ptnapoleon: No, it is not

controllers/disruption_controller.go
```diff
@@ -114,6 +114,29 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		if err == nil {
+			return
+		}
+
+		unwrappedError, ok := err.(chaostypes.DisruptionError)
+		if ok {
+			errorContext := unwrappedError.Context()
+
+			if isModifiedError(unwrappedError) {
+				r.log.Infow(fmt.Sprintf("retryable %s", unwrappedError.Error()), errorContext)
```
luphaz, line 127: Aren't we suppose to have key values here?
```suggestion
				r.log.Infow("a retryable error occured in reconcile loop", "err", unwrappedError, "context", errorContext)
```

An alternative could be:
```suggestion
                                 wraps := make([]interface{}, 0, len(errorContext)+2)
                                 wraps = append(wraps, "err", unwrappedError)
                                 for key, value := range errorContext {
                                   wraps = append(wraps, key, value)
                                 }
				r.log.Infow("a retryable error occured in reconcile loop", wraps...)
```

controllers/disruption_controller.go
```diff
@@ -114,6 +114,29 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			} else {
+				r.log.Errorw(unwrappedError.Error(), errorContext)
+			}
+		}
+
+		if isModifiedError(err) {
+			r.log.Infow(fmt.Sprintf("retryable %s", err.Error()))
+		} else {
+			r.log.Errorw(err.Error())
```
luphaz, line 136:
```suggestion
			r.log.Infow("a retryable error occured in reconcile loop", "err", err)
		} else {
			r.log.Errorw("an error occured in reconcile loop", "err", err)
```

luphaz (approved):

controllers/disruption_controller.go
taihuynh167: please consider to log something here. I feel there are opportunities to simplify  both the wrapping for loop (`	for _, chaosPod := range targetChaosPods {`) and this `case 0` condition

controllers/disruption_controller.go
```diff
@@ -210,9 +217,7 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		// the injection is being created or modified, apply needed actions
 		controllerutil.AddFinalizer(instance, chaostypes.DisruptionFinalizer)
 		if err := r.Update(context.Background(), instance); err != nil {
-			r.log.Errorw("error adding disruption finalizer", "error", err)
-
-			return ctrl.Result{Requeue: true}, err
+			return ctrl.Result{}, fmt.Errorf("error adding disruption finalizer: %w", err)
```
taihuynh167, line 220: why is this no longer a `Requeue: true`?

luphaz: FYI https://github.com/DataDog/chaos-controller/pull/687#discussion_r1162354861

ptnapoleon:
Testing this in staging currently

PR #690, opened 2023-04-11
o11y: add basic tracer and profiler config

nathantournant:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug

- Adds a configless tracer setup in the chaos-controller, along with an option to activate the profiler, leveraging datadog's tooling to better monitor the chaos controller performance.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


o11y/tracer/datadog/datadog.go
```diff
@@ -0,0 +1,48 @@
...
+// New initiated datadog tracer sink
+func New(cfg types.SinkConfig) (*Sink, error) {
+	var err error
+
+	switch {
+	case cfg.EnableTracer:
+		tracer.Start(
+			tracer.WithSampler(tracer.NewRateSampler(cfg.SampleRate)),
+		)
+	case cfg.EnableTracer && cfg.EnableProfiler:
```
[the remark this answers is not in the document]
nathantournant: Basic mistake, will fix now.

main.go
```diff
@@ -358,6 +375,14 @@ func main() {
 		}
 	}()
 
+	// tracer sink
+	tracer, err := tracer.GetSink(tracertypes.SinkDriver(cfg.Controller.TracerSink), cfg.Controller.Tracer)
+	defer tracer.Stop()
+
+	if err != nil {
+		logger.Fatal(err)
```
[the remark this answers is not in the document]
nathantournant: That's a good point; we do not kill the controller on other sink init errors. I'll move to an error.

o11y/tracer/datadog/datadog.go
```diff
@@ -0,0 +1,49 @@
...
+func New(cfg types.SinkConfig) (Sink, error) {
+	var err error
+
+	if cfg.EnableTracer {
+		tracer.Start(
+			tracer.WithSampler(tracer.NewRateSampler(cfg.SampleRate)),
+		)
+	}
+
+	if cfg.EnableTracer && cfg.EnableProfiler {
```
clairecng, line 28: we can't have a profiler without a tracer?

o11y/tracer/types/types.go
```diff
@@ -0,0 +1,24 @@
...
+	EnableProfiler bool
+	SampleRate     float64
+}
+
+// SinkDriver represents a sink driver to use
+type SinkDriver string
+
+const (
+	// SinkDriverDatadog is the Datadog driver
+	SinkDriverDatadog SinkDriver = "datadog"
```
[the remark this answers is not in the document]
nathantournant: I think the reason is to ease the config flag setup; flag/configmap input is a string, and that allows to easily follow through to the option in the code. It's the pattern used for the metrics sink and consistency can be appreciated.
Sure it could be done another way, but this works decently in my opinion.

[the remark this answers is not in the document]
luphaz: Yes indeed, I did not make the correlation with the configuration file, hence string is appropriate

Thanks!

o11y/tracer/datadog/datadog.go
nathantournant: It can; I'll remove the condition.

main.go
```diff
@@ -358,6 +380,22 @@ func main() {
 		}
 	}()
 
+	// tracer sink
+	tracer, err := tracer.GetSink(cfg.Controller.Tracer)
```
luphaz, line 384: It feels weird to me we have the enable flags and do not use them to control the initialization in the main 🤔 

Why aren't we doing this?

main.go
```diff
@@ -358,6 +380,22 @@ func main() {
...
+	defer tracer.Stop()
+
+	if err != nil {
+		logger.Errorw("error while creating tracer sink", "error", err)
+	}
```
luphaz, line 389: You can't expect the tracer to be initialized on error, at least usual defer pattern do not defer if error
```suggestion
	if err != nil {
		logger.Errorw("error while creating tracer sink", "error", err)
	} else {
		defer tracer.Stop()
	}
```

main.go
```diff
@@ -358,6 +380,22 @@ func main() {
...
+
+	// profiler sink
+	profiler, err := profiler.GetSink(cfg.Controller.Profiler)
+	defer profiler.Stop()
+
+	if err != nil {
+		logger.Errorw("error while creating profiler sink", "error", err)
+	}
```
luphaz, line 397: Same here, only defer on non error

o11y/profiler/datadog/datadog.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// Sink describes a datadog profiler sink
+type Sink struct{}
+
+// New initiated datadog profiler sink
+func New(cfg types.SinkConfig) (Sink, error) {
+	var err error
+
+	if cfg.Enable {
```
luphaz, line 20: TBH I would move this outside of the profiler and let the main handle the if enable.

o11y/tracer/datadog/datadog.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// Sink describes a datadog tracer sink
+type Sink struct{}
+
+// New initiated datadog tracer sink
+func New(cfg types.SinkConfig) (Sink, error) {
+	var err error
+
+	if cfg.Enable {
```
luphaz, line 21: Same, let's align on where the enable is handled, creating a not enabled object seems weird to me.

luphaz (commented):
I'm not sure I completely understand what we are doing with Enable/noop here.
It seems both try to achieve the same goal to me 🤔 

We may need a fake instance, so we don't need to check everywhere if metrics/tracer/profiler is enabled, however, we could simply check if `enable` and use `noop` implicitly.

I'm not sure it's OK to create an instance that is not enabled on metrics side, hence, what I would suggest would be to handle this on `(metrics/tracer/profiler).GetSink` where we would check if it's enable, and return `noop` if it is not.

main.go
nathantournant: fixed !

o11y/profiler/datadog/datadog.go
nathantournant: Removed the `enabled` flags. I agree that `noop` is the better way to deactivate tracing and profiling.

chart/templates/configmap.yaml
```diff
@@ -16,11 +16,9 @@ data:
       metricsSink: {{ .Values.controller.metricsSink | quote }}
       tracer:
         sink: {{ .Values.controller.tracer.sink | quote }}
-        enable: {{ .Values.controller.tracer.enable }}
         sampleRate: {{ .Values.controller.tracer.sampleRate }}
       profiler:
         sink: {{ .Values.controller.profiler.sink | quote }}
```
luphaz: Hence I would follow the metrics approach as a single parameter is availble
```suggestion
      profilerSink: {{ .Values.controller.profilerSink | quote }}
```

luphaz (approved):

chart/templates/configmap.yaml
nathantournant: sounds reasonable

aymericDD (approved):
Good job 👏

nathantournant:
I removed the tracer config for now, as I'll use the latest APM updates regarding Datadog's integration with OpenTelemetry's API. In that regard, the proposed Sink was not well-suited.
It will come through a separate PR with a more convenient design.
This only includes the profiler we need to analyse memory and goroutines performance.

PR #694, opened 2023-04-12
[Helm] Parameterize resources for controller

craig-seeman:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- `x` Sets resource allocation via helm variables instead of being hardcoded in the chart template.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


ptnapoleon (approved):

ptnapoleon (commented):
We don't use these limit values in production, just for dev, we set them according to our typical usage in real clusters. It would be better to move these to `values.yaml` and configure them however you'd like, without needing to change `deployment.yaml` in the future

craig-seeman:
> We don't use these limit values in production, just for dev, we set them according to our typical usage in real clusters. It would be better to move these to `values.yaml` and configure them however you'd like, without needing to change `deployment.yaml` in the future

I like this approach! I'll shoot that into a lil' refactor tomorrow here and do that~ Thanks for the feedback!

Devatoria:
Thanks for the contribution. We also have people here working on improving the overall memory usage of the controller to make it more stable because it can spike for large experiments indeed. We used to have to temporarily increase that limit during some injections on a few thousands targets at Datadog otherwise it would OOM too. Hopefully we have a bunch of fixes soon to help with that.

craig-seeman:
Any strong feelings on the level of implementation? 

After looking at the config with the resources defined, I'm sort of leaning more towards a simpler config file (A) and just specifying the limit and request in the same setting as opposed to allow both the request and limit to be set individually (B).

(A)
```yaml
resources:
  cpu: 100m
  memory: 700Mi
```

(B)
```yaml
resources:
  limits:
    cpu: 100m
    memory: 700Mi
  requests:
    cpu: 100m
    memory: 700Mi
```

craig-seeman:
Went with the (A) Approach outlined above to keep simplicity around the configs. I also went ahead and set the default back to 300Mi after realizing _we_ might be the exception vs the norm in our deployment size.

chart/values.yaml
```diff
@@ -57,6 +57,9 @@ controller:
     enable: false
     namespaceThreshold: 80
     clusterThreshold: 66
+  resources: # resources assinged to the controller pod. may need to be increased when deploying to larger scale clusters
```
ptnapoleon, line 60:
```suggestion
  resources: # resources assigned to the controller pod. may need to be increased when deploying to larger scale clusters
```

ptnapoleon:
One nit and then I'll merge. Definitely like the approach of keeping limits and requests identical for the controller

PR #695, opened 2023-04-17
Fix race condition in cpu_pressure_test.go

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- The test was failing frequently in CI and locally. I found that in `cpu_pressure.go`:
```
if i.prioritizeStresserProcess(core, tid) {
				mutex.Lock()
				i.routines++
				mutex.Unlock()
				wg.Done()
				i.startStresser(core, tid)
			}
```
The `wg.Done()` is called immediately before `i.startStresser`. This code block is running in a goroutine. Thus in the test code we might Assert that `Stress` was called before we have guaranteed it would have been called by the app code. Thus I wrap it in an Eventually.

- I also changed all of these `It`s into `By`s. Why? Well `JustBeforeEach` and `BeforeEach` actually run before each `It`, thus we're making a series of assertions about a single injection, while actually doing five injections (and cleanups). This doesn't affect the correctness, but is a waste of time.

- I added a TODO and need to ask @taihuynh167, what `AfterEach` are we relying on in the second test? I couldn't find one when I checked out the commit that added the second test.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


ptnapoleon:
Rats, the test still failed for me after ~30 runs. Less often, but will keep looking

luphaz (approved):

ptnapoleon:
Eventually can't stop that Assert from failing the test, need to change the approach

luphaz (approved):

PR #698, opened 2023-04-17
feat(make): templates and target vars

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- This PR aims to use make capabilities to reduce duplication:
  - [functions](https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html)
  - [substitutions](https://www.gnu.org/savannah-checkouts/gnu/make/manual/html_node/Text-Functions.html)
  - [target specific variables](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html)
  - And templating 🙈 

NB: this was mostly to experiment what we can do with make, it might not be more maintainable this way, let me know WDYT? 

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: manager injector handler release
+.PHONY: manager injector handler release generate generate-mocks clean-mocks all lima-push-all lima-redeploy lima-all
 .SILENT: release
 
 # Lima requires to have images built on a specific namespace to be shared to the Kubernetes cluster when using containerd runtime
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -37,23 +37,65 @@ else
...
+_handler:;
+_manager: generate
+
+_docker-build-injector: docker-build-ebpf
+_docker-build-handler:;
+_docker-build-manager:;
+
+# we define the template we expect for each target
+define TARGET_template
+$(1): BINARY_NAME=$(1)
```
luphaz, line 65: `$(1)` is the target name `injector|handler|manager`

Makefile
```diff
@@ -37,23 +37,65 @@ else
...
+docker-build-ebpf:
+	docker build --platform linux/$(OS_ARCH) --build-arg ARCH=$(OS_ARCH) -t ebpf-builder-$(OS_ARCH) -f bin/ebpf-builder/Dockerfile ./bin/ebpf-builder/
+	-rm -r bin/injector/ebpf/
+	docker run --rm -v $(shell pwd):/app ebpf-builder-$(OS_ARCH)
+
+lima-push-injector lima-push-handler lima-push-manager: FAKE_FOR=COMPLETION
+
+_injector:;
+_handler:;
+_manager: generate
```
luphaz, line 57: As we use a template, the template must be valid for all targets, we hence have empty targets when no dependencies before build

Makefile
```diff
@@ -37,23 +37,65 @@ else
...
+docker-build-injector: IMAGE_TAG=$(INJECTOR_IMAGE)
+docker-build-handler: IMAGE_TAG=$(HANDLER_IMAGE)
+docker-build-manager: IMAGE_TAG=$(MANAGER_IMAGE)
+
+docker-build-ebpf:
+	docker build --platform linux/$(OS_ARCH) --build-arg ARCH=$(OS_ARCH) -t ebpf-builder-$(OS_ARCH) -f bin/ebpf-builder/Dockerfile ./bin/ebpf-builder/
+	-rm -r bin/injector/ebpf/
+	docker run --rm -v $(shell pwd):/app ebpf-builder-$(OS_ARCH)
+
+lima-push-injector lima-push-handler lima-push-manager: FAKE_FOR=COMPLETION
```
luphaz, line 53: This aims to provide bashcompletion or similar as they are parsing the Makefile, not necessarily the templates

Makefile
ptnapoleon: This would be helpful as a direct comment right above here

ptnapoleon (approved):
works for me? The Makefile was getting fairly verbose

Makefile
luphaz:
```suggestion
# $(1) is the target name: injector|handler|manager
$(1): BINARY_NAME=$(1)
```

PR #701, opened 2023-04-18
test(gingko): update to v2

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- This PR aims to upgrade `ginkgo` to `v2`
- Changes some implementations to fullfill new Ginkgo expectations and ensure expectations can be asserted
  - some test implementations (injector dns|network) are doing a setup of ALL mocks methods as expectations, hence short term solution is to use `Maybe` so the call can happen, or not

NB: previously, mock expectations WERE NOT ASSERTED because in v1 `t.Cleanup` is a [no-op](https://github.com/onsi/ginkgo/blob/d38b9d946d52cd175495d30143fbecc5aff98f13/internal/testingtproxy/testing_t_proxy.go#L34), and this is what is used by testify, hence if no explicit call to `mock.AssertExpectations` was made, assertions WERE NOT CHECKED 🙈 

And more changes:
- This PR also switch the test runner from `go` to `ginkgo` and uses their [suggested](https://onsi.github.io/ginkgo/#recommended-continuous-integration-configuration) configuration for CI 
- This PR also uploads the junit test report to circle ci to ease debugging and detect long tests, flakky tests etc
- This PR also changes the resource_class of VM we are using in the hope to speed up tests a little
- This PR also changes CI definition to avoid duplication and add more parallelism to speed up our pipelines (still not enough IMO... it's a start from 15min to 12min)
  - mostly it will build the containers before the e2e-test and reuse the generated tars and use [DSL](https://circleci.com/docs/docker-layer-caching/#limitations)
- This PR also replace cache by [workspace](https://circleci.com/docs/persist-data/) to [share](https://circleci.com/docs/workspaces/) code between steps

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


Makefile
```diff
@@ -49,7 +55,15 @@ docker-build-manager: IMAGE_TAG=$(MANAGER_IMAGE)
 docker-build-ebpf:
 	docker build --platform linux/$(OS_ARCH) --build-arg ARCH=$(OS_ARCH) -t ebpf-builder-$(OS_ARCH) -f bin/ebpf-builder/Dockerfile ./bin/ebpf-builder/
 	-rm -r bin/injector/ebpf/
-	docker run --rm -v $(shell pwd):/app ebpf-builder-$(OS_ARCH)
+# create a dummy container with volume to store files
```
luphaz, line 58: Need to use this approach due to circleci:
* https://circleci.com/docs/building-docker-images/#mounting-folders

luphaz:
NB: just before merging this PR I will update status checks accordingly as they were renamed

.circleci/config.yml
```diff
@@ -7,62 +7,76 @@ version: 2.1
 
 commands:
   go_save_cache:
-    description: "Save $GOHOME directory into the cache for the actual revision"
+    description: Save $GOPATH directory as workspace
     steps:
-      - save_cache:
-          key: env-{{ .Revision }}
+      - persist_to_workspace:
```
luphaz, line 12: Workspace are meant to be used to share code with downstream steps and provide retry due to their persistence.

.circleci/config.yml
```diff
@@ -7,62 +7,76 @@ version: 2.1
...
+  ebpf_builder_save_cache:
+    description: Save ebpf-builder tar for later use
+    parameters:
+      path:
+        type: string
+        default: bin/ebpf-builder/ebpf-builder.tar.gz
+        description: ebpf-builder tar location
+    steps:
+      - run: docker save ebpf-builder-amd64 -o <<parameters.path>>
+      - save_cache:
```
luphaz, line 46: Here we want to store this for long term and reuse it in next builds, it aims to speed up the docker build time, like it does already locally.

.circleci/config.yml
```diff
@@ -7,62 +7,76 @@ version: 2.1
...
+            - v1-ebpf-builder-{{ checksum "bin/ebpf-builder/Dockerfile" }}
+            # if cache for exact version of docker image is not present, then load any most recent one
+            - v1-ebpf-builder-
+      - run: docker load -i <<parameters.path>> || true
 
 templates:
   # sets the working directory to the project path
   working_directory: &working_directory
-    working_directory: /go/src/github.com/DataDog/chaos-controller
+    working_directory: /home/circleci/go/src/github.com/DataDog/chaos-controller
```
luphaz, line 68: Due to transition to circle CI next gen images, it's better to move into a place where we don't need to be root to write files:
* https://circleci.com/docs/circleci-images/#next-gen-language-images

.circleci/config.yml
```diff
@@ -7,62 +7,76 @@ version: 2.1
...
     docker:
       # TODO: Publicize the chaos controller ci image to the datadog docker registry
       #- image: datadog/chaos-controller-runner-circle:<< pipeline.parameters.CURRENT_CI_IMAGE >>
-      - image: golang:1.18-buster
-    environment:
-      KUBEBUILDER_VERSION: 3.1.0
-      KUBEBUILDER_ARCH: amd64
-      GOLANGCI_LINT_VERSION: 1.51.0 ## this value should be the same as in the Makefile
-      HELM_VERSION: 3.5.4
+      - image: cimg/go:1.18
```
luphaz, line 76: This is circle ci images, provides default tool installed (like docker) to ease step definition and avoid apt-get/update things

.circleci/config.yml
```diff
@@ -7,62 +7,76 @@ version: 2.1
...
   golang_large:
-    <<: *working_directory
-    docker:
-      # TODO: Publicize the chaos controller ci image to the datadog docker registry
-      #- image: datadog/chaos-controller-runner-circle:<< pipeline.parameters.CURRENT_CI_IMAGE >>
-      - image: golang:1.18-buster
-    resource_class: medium+
-    environment:
-      KUBEBUILDER_VERSION: 3.1.0
```
luphaz, line 62: moved to make

.circleci/config.yml
```diff
@@ -72,181 +86,146 @@ executors:
...
+                command: git diff --exit-code
   # runs the tests
   test:
-    executor: golang
+    executor: golang_large
     steps:
       - go_restore_cache
       - run: make test
+      - store_test_results:
+          path: report.xml
```
luphaz, line 173: upload test reports to circle ci to enjoy [this](https://app.circleci.com/insights/github/DataDog/chaos-controller/workflows/test_and_build/tests?branch=matthieu.bono/CHAOS-503/upgrade-gingko-v2)

.circleci/config.yml
```diff
@@ -72,181 +86,146 @@ executors:
...
+          command: make lima-install-cert-manager KUBECTL="minikube kubectl --"
       - run:
           name: Build and load images
-          <<: *working_directory
           command: make minikube-load-all
       - run:
           name: Run e2e tests
-          <<: *working_directory
           command: |
+            export PATH="/home/circleci/go/bin:${PATH}"
```
luphaz, line 194: we put all binaries into this path to ease their usage

api/v1beta1/disruption_webhook_test.go
```diff
@@ -180,7 +183,7 @@ var _ = Describe("Disruption", func() {
 					// Assert
 					Expect(err).Should(HaveOccurred())
 					Expect(err.Error()).Should(HavePrefix("the controller is currently in delete-only mode, you can't create new disruptions for now"))
-					Expect(ddmarkMock.AssertNotCalled(GinkgoT(), "ValidateStructMultierror", mock.Anything, mock.Anything)).To(BeTrue())
+					Expect(ddmarkMock.AssertNumberOfCalls(GinkgoT(), "ValidateStructMultierror", 0)).To(BeTrue())
```
aymericDD, line 186: 👏

cloudservice/manager_test.go
```diff
@@ -236,7 +236,6 @@ var _ = Describe("New function", func() {
 func NewCloudServiceMock(isNewVersionMockValue bool, isNewVersionError error, convertToGenericIPRangesVersion string, convertToGenericIPRangesServiceList []string, convertToGenericIPRanges map[string][]string, convertToGenericIPRangesError error) *CloudProviderIPRangeManagerMock {
 	cloudProviderIPRangeMock := NewCloudProviderIPRangeManagerMock(GinkgoT())
 
-	cloudProviderIPRangeMock.EXPECT().IsNewVersion(mock.Anything, mock.Anything).Return(isNewVersionMockValue, isNewVersionError)
```
aymericDD, line 239: Are you sure this call is not expected? Maybe they are an issue in the implementation or the coverage is not enough?

taihuynh167 (approved):
💯

injector/dns_disruption_test.go
```diff
@@ -36,31 +36,32 @@ var _ = Describe("Failure", func() {
...
+		netnsManagerMock.EXPECT().Enter().Return(nil).Maybe()
+		netnsManagerMock.EXPECT().Exit().Return(nil).Maybe()
 
 		// container
 		ctn := mocks.NewContainerMock(GinkgoT())
 
 		// pythonRunner
 		pythonRunner := mocks.NewPythonRunnerMock(GinkgoT())
-		pythonRunner.EXPECT().RunPython(mock.Anything, mock.Anything, mock.Anything).Return(nil)
+		pythonRunner.EXPECT().RunPython(mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe().Maybe()
```
aymericDD, line 54: Why use twice `Maybe`?

injector/dns_disruption_test.go
```diff
@@ -36,31 +36,32 @@ var _ = Describe("Failure", func() {
 	BeforeEach(func() {
 		// cgroup
 		cgroupManagerMock = mocks.NewCGroupManagerMock(GinkgoT())
-		cgroupManagerMock.EXPECT().RelativePath(mock.Anything).Return("/kubepod.slice/foo")
-		cgroupManagerMock.EXPECT().Write(mock.Anything, mock.Anything, mock.Anything).Return(nil)
+		cgroupManagerMock.EXPECT().RelativePath(mock.Anything).Return("/kubepod.slice/foo").Maybe()
```
aymericDD, line 39: I am not really a fan of the `Maybe` keyword, because it does not check if the method has been called during the `AssertExpectation` and let the suit test be successful. I think the tests should be reworked in order to define `EXPECT` when it is needed.

aymericDD: In a first time we can allow the `Maybe`, but it must be refactored in another PR.

controllers/cache_handler_test.go
```diff
@@ -39,27 +38,27 @@ var _ = Describe("Cache Handler verifications", func() {
 	var disruption *v1beta1.Disruption
 	var targetLabels map[string]string
 
-	AfterEach(func() {
+	AfterEach(func(ctx SpecContext) {
```
aymericDD, line 41: Noice! Good usage of the context 👍

aymericDD (approved):
Really good job! 👏

cloudservice/manager_test.go
luphaz: I am yes, I looked at the code.
This is expected when there is already something stored.
Indeed, some tests might be missing

injector/dns_disruption_test.go
luphaz: Mistake, fixing
```suggestion
		pythonRunner.EXPECT().RunPython(mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
```

injector/dns_disruption_test.go
luphaz: Indeed, I won't rework those tests here, the PR is big enough (an `Maybe` is exactly was the initial implementation of those tests was doing, hence correct)

I will create an issue to rework those and the network ones, thanks for pointing that out.

luphaz: done

.circleci/config.yml
luphaz, line 12:
```suggestion
      # circleci workspace are meant to be used to share data between jobs
      - persist_to_workspace:
```

luphaz:
```suggestion
      # circleci workspace are meant to be used to share data between jobs
      # https://circleci.com/docs/workspaces/
      - persist_to_workspace:
```

PR #709, opened 2023-05-02
CHAOS-437: fix node failure examples, add new ddmark rule

nathantournant:
## What does this PR do?

- [X] Adds new functionality
- [X] Fixes a bug

- The example around `nodeFailure` can seem confusing; it's currently easy to start a nodeFailure disruption defaulting to the pod level, with a node-oriented selector, that will most likely find no target (or worse, undesired targets).
- This PR adds a related DDMark rule allowing to solve that issue, by checking a given struct subfield value as a trigger to expect other subfield presence or specific value. Read `ddmark/README.md` for more details.

## Code Quality Checklist

## Testing

    - [X] by adding new `unit` tests or `end-to-end` tests.
    - play with `examples/node_failure.yaml`: remove the `nodeFailure` and `level: node` fields and validate with `chaosli` to see the difference. Do not forget to build an update chaosli (`make chaosli`) on the branch, and test with the command`./bin/chaosli/chaosli_darwin_arm64 validate --path examples/node_failure.yaml`
    - [X] locally.

### DIFF
--- a/ddmark/README.md
+++ b/ddmark/README.md
@@ -52,10 +52,26 @@ It also allows us to define custom rules to apply to our structures, and focus t
   - `// +ddmark:validation:ExclusiveFields={<fieldName1>,<fieldName2>,...}`
   - Applies to: any `<struct>` type
   - Asserts that `<fieldname1>` can only be non-nil iff all of the other fields are `nil`
-- LinkedFields:
-  - `// +ddmark:validation:LinkedFields={<fieldName1>,<fieldName2>,...}`
-  - Applies to: any `<struct>` type
-  - Asserts the fields in the list are either all `nil` or all non-`nil`
+- LinkedFieldsValue:
+  - without values:
+    -  `// +ddmark:validation:LinkedFieldsValue={<fieldName1>,<fieldName2>,...}`
+    - Applies to: any `<struct>` type
+    - Asserts the fields in the list are either all `nil` or all non-`nil`
+  - with values:
+    - `// +ddmark:validation:LinkedFieldsValue={<fieldName1>=<value>,<fieldName2>,...}`
+    - Applies to: any `<struct>` type -- value checks are only for `string` and `int` subfields
+    - Asserts the fields in the list are either all `nil` (or are *specifically* set to the indicated value) or all non-`nil` (or are ***specifically not*** set to the given value)
+    - Fields with values and without values can be mixed
+- LinkedFieldsValueWithTrigger
+  - without values:
+    - `// +ddmark:validation:LinkedFieldsValueWithTrigger={<fieldName1>, <fieldName2>,...}`
+    - Applies to: any `<struct>` type
+    - Asserts the following: if the first field is non-`nil`, all the following fields have to also be non-`nil`
+  - with values:
+    - `// +ddmark:validation:LinkedFieldsValueWithTrigger={<fieldName1>=<value>, <fieldName2>,...}`
+    - Applies to: any `<struct>` type -- value checks are only for `string` and `int` subfields 
+    - Asserts the following: if the first field is non-`nil` OR is set to the given `<value>`, all the following fields have to also be non-`nil` OR set to their given `<value>`.
+    - Fields with values and without values can be mixed
 - AtLeastOneOf:
   - `// +ddmark:validation:AtLeastOneOf={<fieldName1>,<fieldName2>,...}`
   - Applies to: any `<struct>` type
--- a/ddmark/validation.go
+++ b/ddmark/validation.go
@@ -141,11 +141,10 @@ func (e Enum) ApplyRule(fieldvalue reflect.Value) error {
 	fieldInterface := fieldvalue.Interface()
 
 	for _, markerInterface := range e {
-		if !reflect.ValueOf(markerInterface).Type().ConvertibleTo(fieldvalue.Type()) {
+		markerInterface, isOk := attemptConversion(markerInterface, fieldvalue)
+		if !isOk {
 			return e.TypeCheckError(fieldvalue)
 		}
-
-		markerInterface = reflect.ValueOf(markerInterface).Convert(fieldvalue.Type()).Interface()
 
 		if fieldInterface == markerInterface || reflect.ValueOf(fieldInterface).IsZero() {
 			return nil
@@ -341,6 +340,16 @@ func ruleName(i interface{}) string {
 // genericTypeError returns a generic error for wrong type marker attempt
 func genericTypeCheckError(i interface{}, fieldValue reflect.Value, expectedTypes string) error {
 	return fmt.Errorf("%s: marker applied to wrong type: currently %T, can only be %s", ruleName(i), fieldValue, expectedTypes)
+}
+
+// attemptConversion tries to convert markerInterface to fieldValue's interface type
+func attemptConversion(markerInterface interface{}, fieldValue reflect.Value) (any, bool) {
+	convertible := reflect.ValueOf(markerInterface).Type().ConvertibleTo(fieldValue.Type())
+	if !convertible {
+		return nil, false
+	}
+
+	return reflect.ValueOf(markerInterface).Convert(fieldValue.Type()).Interface(), true
 }
 
 // structValueToMap takes a struct value and turns it into a map, allowing more flexible field and value parsing
--- a/ddmark/validation.go
+++ b/ddmark/validation.go
@@ -62,14 +62,24 @@ func (m Maximum) ApplyRule(fieldvalue reflect.Value) error {
 	fieldInt, ok := parseIntOrUInt(fieldvalue)
 
 	if !ok {
-		return fmt.Errorf("%v: marker applied to wrong type: currently %v, can only be %v", ruleName(m), fieldvalue.Type(), "int or uint")
+		return m.GenTypeCheckError(fieldvalue)
 	}
 
 	if int(m) < fieldInt {
-		return fmt.Errorf("%v: field has value %v, max is %v (included)", ruleName(m), fieldInt, m)
+		return m.GenValueCheckError(fieldInt)
 	}
 
 	return nil
+}
+
+func (m Maximum) GenValueCheckError(fieldInt int) error {
+	template := "%v: field has value %v, max is %v (included)"
+	return fmt.Errorf(template, ruleName(m), fieldInt, m)
+}
+
+func (m Maximum) GenTypeCheckError(fieldValue reflect.Value) error {
+	template := "%v: marker applied to wrong type: currently %v, can only be %v"
+	return fmt.Errorf(template, ruleName(m), fieldValue.Type(), "int or uint")
 }
 
 func (m Minimum) ApplyRule(fieldvalue reflect.Value) error {
@@ -77,14 +87,24 @@ func (m Minimum) ApplyRule(fieldvalue reflect.Value) error {
 	fieldInt, ok := parseIntOrUInt(fieldvalue)
 
 	if !ok {
-		return fmt.Errorf("%v: marker applied to wrong type: currently %v, can only be %v", ruleName(m), fieldvalue.Type(), "int or uint")
+		return m.GenTypeCheckError(fieldvalue)
 	}
 
 	if int(m) > fieldInt {
-		return fmt.Errorf("%v: field has value %v, min is %v (included)", ruleName(m), fieldInt, m)
+		return m.GenValueCheckError(fieldInt)
 	}
 
 	return nil
+}
+
+func (m Minimum) GenValueCheckError(fieldInt int) error {
+	template := "%v: field has value %v, min is %v (included)"
+	return fmt.Errorf(template, ruleName(m), fieldInt, m)
+}
+
+func (m Minimum) GenTypeCheckError(fieldValue reflect.Value) error {
+	template := "%v: marker applied to wrong type: currently %v, can only be %v"
+	return fmt.Errorf(template, ruleName(m), fieldValue.Type(), "int or uint")
 }
 
 func (e ExclusiveFields) ApplyRule(fieldvalue reflect.Value) error {
@@ -106,10 +126,15 @@ func (e ExclusiveFields) ApplyRule(fieldvalue reflect.Value) error {
 	}
 
 	if matchCount >= 1 {
-		return fmt.Errorf("%v: some fields are incompatible, %s can't be set alongside any of %v", ruleName(e), e[0], e[1:])
+		return e.GenValueCheckError()
 	}
 
 	return nil
+}
+
+func (e ExclusiveFields) GenValueCheckError() error {
+	template := "%v: some fields are incompatible, %s can't be set alongside any of %v"
+	return fmt.Errorf(template, ruleName(e), e[0], e[1:])
 }
 
 func (e Enum) ApplyRule(fieldvalue reflect.Value) error {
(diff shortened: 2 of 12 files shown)


ddmark/validation.go
```diff
@@ -241,3 +297,53 @@ func parseIntOrUInt(value reflect.Value) (int, bool) {
 
 	return fieldInt, ok
 }
+
+// checkValueExistOrIsValid checks if a given string marker item name value exist in a unmarshalled struct (converted to a map by structValueToMap)
+// it returns 1 if the value is found and -if applicable- the required value is valid, 0 otherwise
+func checkValueExistsOrIsValid(markerItem string, structMap map[string]interface{}, ruleName string) (int, error) {
```
luphaz, line 303: Why using an int instead of a bool?
Even if the method is used by incrementing a counter, it feels to me this method should return a boolean instead. 
It's up to the caller to decide to increment a counter, not to this method purpose IMO

ddmark/validation.go
```diff
@@ -241,3 +297,53 @@ func parseIntOrUInt(value reflect.Value) (int, bool) {
...
+	// marketItem can either be a fieldName, or fieldName=fieldValue
+	markerItemArr := strings.Split(markerItem, "=")
```
luphaz, line 305: You may want to use cut instead:
* https://pkg.go.dev/strings#Cut

```suggestion
	before, after, found := strings.Cut(markerItem, "=")
```

https://go.dev/play/p/py5xzqG_SgL

ddmark/validation.go
```diff
@@ -241,3 +297,53 @@ func parseIntOrUInt(value reflect.Value) (int, bool) {
...
+			case reflect.Int:
+				vInt := v.Convert(tv).Interface().(int)
+				vStr = strconv.Itoa(vInt)
+			case reflect.String:
+				vStr = v.Convert(t).Interface().(string)
+			default:
+				return 0, fmt.Errorf("%v: please do not apply this marker to anything else than int or string. Current type: %v", ruleName, tv.Name())
+			}
+
+			if strings.Compare(markerSubfieldValue, vStr) == 0 {
```
luphaz, line 337:
```suggestion
			if markerSubfieldValue == vStr {
```

Using string.Compare is only relevant if you use the various output (-1/1/0) here as it's the simple equality it feels more natural to do a natural comparison (case sensitive).

You could add the option to check for equality without case sensitivity (EqualFold)

ddmark/validation.go
```diff
@@ -241,3 +297,53 @@ func parseIntOrUInt(value reflect.Value) (int, bool) {
...
+			return 1, nil
+		}
+	// a value was required => check item has described value
+	case 2:
+		markerSubfieldName, markerSubfieldValue := markerItemArr[0], markerItemArr[1]
+		if structMap[markerSubfieldName] == nil {
+			break
+		}
+
+		if reflect.Indirect(reflect.ValueOf(structMap[markerSubfieldName])).Type().ConvertibleTo(reflect.TypeOf(markerSubfieldValue)) {
```
luphaz, line 320: There is a lot of nesting here that makes the "golden path" hard to understand.

In general you should favour escaping when it's not the golden path to reduce nesting.

More details:
* https://www.youtube.com/watch?v=CFRhGnuXG-4&ab_channel=CodeAesthetic
* https://drew.thecsillags.com/posts/2014-04-14-refactoring-flow-control-and-conditionals/

luphaz: I would even extract this to a method TBH (extract or even compare int or string or something similar)

ddmark/validation_integration_test.go
```diff
@@ -166,70 +166,148 @@ exclusivefieldstest:
...
+		})
 		It("checks out valid all-nil values", func() {
-			linkedfieldsYaml := `
-linkedfieldstest:
+			linkedfieldsvalueYaml := `
+linkedfieldsvaluetest:
   randomintfield: 1
   strfield:
   pstrfield:
   intfield: 0 # is nil
```
luphaz, line 200:
```suggestion
  intfield: 0 # is empty
```

ddmark/validation_integration_test.go
```diff
@@ -166,70 +166,148 @@ exclusivefieldstest:
...
+			linkedfieldsvaluewithtriggerYaml := `
+linkedfieldsvaluewithtriggertest:
+  strfield: aa      # invalid
+  pstrfield: bb     # invalid
+  intfield: 12
+  pintfield: 1
+  aintfield: [1,2]
+`
+			err := validateString(linkedfieldsvaluewithtriggerYaml)
+			Expect(err.Errors).To(HaveLen(0))
```
luphaz, line 258:
```suggestion
			Expect(err.Errors).To(BeEmpty())
```

FYI this can be used also

ddmark/validation_integration_test.go
```diff
@@ -166,70 +166,148 @@ exclusivefieldstest:
...
 		})
 	})
 
+	Context("LinkedFieldsValueWithTrigger Marker", func() {
+		// +ddmark:validation:LinkedFieldsValueWithTrigger={StrField=aaa,IntField=2}
+		// +ddmark:validation:LinkedFieldsValueWithTrigger={PStrField=bbb,PIntField=12,AIntField}
+		It("checks out valid all-incorrect initial values", func() {
+			linkedfieldsvaluewithtriggerYaml := `
+linkedfieldsvaluewithtriggertest:
+  strfield: aa      # invalid
```
luphaz, line 251: I'm sorry it's a little confusing for me here 🤔 

How can we have invalid as a comment and no error?

ddmark/validation_integration_test.go
```diff
@@ -166,70 +166,148 @@ exclusivefieldstest:
...
+			linkedfieldsvaluewithtriggerYaml := `
+linkedfieldsvaluewithtriggertest:
+  strfield: aaa
+  pstrfield: bbb
+  intfield:  # is nil
+  pintfield: # is nil
+  aintfield: # is nil
+`
+			err := validateString(linkedfieldsvaluewithtriggerYaml)
+			Expect(err.Errors).To(HaveLen(2))
```
luphaz, line 283: I would probably check the expected string to confirm validation rules that triggered is the one expected.

One of the simplest way would be:
```suggestion
			Expect(err.Errors).To(HaveLen(2))
			Expect(err.Errors[0]).To(MatchError("what will be in the Error() string"))
			Expect(err.Errors[1]).To(MatchError("what will be in the other Error() string"))
```

ddmark/validation_teststruct.go
```diff
@@ -67,9 +68,20 @@ type ExclusiveFieldsTestStruct struct {
 	CField    int
 }
 
-// +ddmark:validation:LinkedFields={StrField,IntField}
-// +ddmark:validation:LinkedFields={PStrField,PIntField,AIntField}
-type LinkedFieldsTestStruct struct {
+// +ddmark:validation:LinkedFieldsValue={StrField=aaa,IntField}
+// +ddmark:validation:LinkedFieldsValue={PStrField,PIntField,AIntField}
+type LinkedFieldsValueTestStruct struct {
+	RandomIntField int // allows to actually check all-empty structs
```
luphaz, line 74: 🤔 not sure I get it. If we remove this field what happens?

ddmark/validation_unit_test.go
```diff
@@ -233,6 +233,68 @@ var _ = Describe("Validation Rules Cases", func() {
 			fakeObj.Field2 = 0
 			Expect(linked.ApplyRule(ValueOf(fakeObj))).ToNot(BeNil())
 		})
+
+		It("rejects object with one missing field (Field2 is 3, expected value was 2)", func() {
+			fakeObj.Field2 = 3
+			Expect(linked.ApplyRule(ValueOf(fakeObj))).ToNot(BeNil())
```
luphaz, line 239:
```suggestion
			Expect(linked.ApplyRule(ValueOf(fakeObj))).ToNot(Succeed())
```

It's more readable to use `Succeed` when function returns an error. Then use `To/ToNo` to express what is expected

In general when you expect an error, I tend to prefer checking the EXACT error message:

```suggestion
			Expect(linked.ApplyRule(ValueOf(fakeObj))).To(MatchError("what is inside the error message"))
```

That will check not only error exists AND is the expected one.

ddmark/validation_unit_test.go
```diff
@@ -233,6 +233,68 @@ var _ = Describe("Validation Rules Cases", func() {
...
+		}
+
+		// we expect that; if Field1 value is "aaa" then Field 2 value is 12 and Field3 is not nil
+		arr := []string{"Field1=aaa", "Field2=12", "Field3"}
+		linked := LinkedFieldsValueWithTrigger(arr)
+		var fakeObj dummyStruct
+
+		BeforeEach(func() {
+			i := 3
+			var pi *int = &i
```
luphaz, line 257:
```suggestion
			pi := &i
```

ddmark/validation_unit_test.go
```diff
@@ -233,6 +233,68 @@ var _ = Describe("Validation Rules Cases", func() {
...
+				Field1: "aaa",
+				Field2: 12,
+				Field3: pi,
+			}
+		})
+		It("validates object with all correct fields", func() {
+			Expect(linked.ApplyRule(ValueOf(fakeObj))).To(BeNil())
+		})
+
+		It("validates object with all nil fields", func() {
```
luphaz, line 269:
```suggestion
		It("validates object with all zero fields", func() {
```

luphaz (commented):
Nice addition 👏 

It will definitely be helpful to express which field is required or not.

I'm wondering if we can express this in the OpenAPI schema also 🤔

ddmark/validation.go
nathantournant: agreed

ddmark/validation.go
nathantournant: I forgot about Cut -- nice. I'll adapt.

nathantournant: there's also a point to make about lack of testing for a value field requiring an empty value (`"fieldEmptyValue="`). I'll add test to assert the marker's behaviour.

nathantournant: done ! along with an ongoing test rework to adapt to new codebase standards

ddmark/validation_unit_test.go
nathantournant: I agree it adds a greater level of precision and is better in general.
These error messages are customised to every marker definition because of the included field names -- and I don't like the idea of expecting a string hard-coded error, as it adds toil and I think it provides little added value in this situation.
I'll experiment with a error-generation functions in the DDValidationMarker interface, that would allow to standardise the error generation and reduce the amount of hard-coding to do this. It'll allow for a clearer code base and more precise tests.

ddmark/validation_teststruct.go
nathantournant: This might seem a bit intricate; the struct-level marker is only run on a given struct if it has at lease 1 existing subfield. If no subfield is set, the struct is never unmarshalled and just skipped.
All considered this is an expected and desirable behavior, otherwise every possible marked struct subfields would become de-facto Required in the validation of a parent struct.
This field is not here to be tested, but to allow for all-empty struct testing (ie. all 5 testing fields are empty or nil) without seeing the struct be skipped in the integration test.

ddmark/validation.go
nathantournant: Came from a switch refactoring -- just realised how horrific that code was, adjusting !

ddmark/validation_integration_test.go
nathantournant: phrasing is bad, I'll edit; if the trigger values are not fitting the marker requirement (ie. invalid, incorrect, unfit), the marker validation is valid whatever the other fields' values

ddmark/validation_unit_test.go
nathantournant: In the end, I compromised with type-interfaced error generation functions for unit tests, and hard-coded strings for integration tests

docs/node_disruption.md
```diff
@@ -5,4 +5,13 @@ The `nodeFailure` field triggers a kernel panic on the node. Because the node wi
 * `/proc/sys/kernel/sysrq` > `/mnt/sysrq`
 * `/proc/sysrq-trigger` > `/mnt/sysrq-trigger`
 
-> :warning:️ Node behavior when using this disruption can differ depending on the cloud provider (node may or may not be replaced, restarted, cordoned, etc.).
\ No newline at end of file
+> :warning:️ Node behavior when using this disruption can differ depending on the cloud provider (node may or may not be replaced, restarted, cordoned, etc.).
+
+## Targeting
+
+For clarity purpose, it is mandatory to explicitely set the disruption's `level` field to either `pod` or `node`.
```
ptnapoleon, line 12:
```suggestion
For clarity purpose, it is mandatory to explicitly set the disruption's `level` field to either `pod` or `node`.
```

docs/node_disruption.md
```diff
@@ -5,4 +5,13 @@ The `nodeFailure` field triggers a kernel panic on the node. Because the node wi
...
+
+The `nodeFailure` disruption acts on a node. However, the `level` field needs to be adapted for the selector:
+
+- if `level: node` is set, the selector will target nodes and impact those directly nodes.
```
ptnapoleon, line 16:
```suggestion
- if `level: node` is set, the selector will target nodes and impact those nodes directly.
```

docs/node_disruption.md
```diff
@@ -5,4 +5,13 @@ The `nodeFailure` field triggers a kernel panic on the node. Because the node wi
...
+- if `level: pod` is set, the selector will targets pods and impact the nodes that hosts them.
```
ptnapoleon, line 17:
```suggestion
- if `level: pod` is set, the selector will targets pods and impact the nodes that host them.
```

examples/node_failure.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
   annotations:
     chaos.datadoghq.com/environment: "lima"
 spec:
+  level: node # selector targets nodes instead of pods. Targeting a pod will impact the node hosting given pod.
   selector:
     app: demo-curl
```
ptnapoleon, line 16: If you're changing the level, you need to change the selector to not be a pod selector anymore

examples/node_failure_shutdown.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
   annotations:
     chaos.datadoghq.com/environment: "lima"
 spec:
+  level: node # selector targets nodes instead of pods. Targeting a pod will impact the node hosting given pod.
```
ptnapoleon, line 14: If you're changing the level, you need to change the selector to not be a pod selector anymore

ptnapoleon (changes requested):
The new examples no longer work in our demo environment

ddmark/validation.go
```diff
@@ -151,33 +199,102 @@ func (l LinkedFields) ApplyRule(fieldvalue reflect.Value) error {
...
+
+	if matchCount != 0 && matchCount != len(l) {
+		return l.GenValueCheckError()
+	}
+
+	return nil
+}
+
+func (l LinkedFieldsValueWithTrigger) GenValueCheckError() error {
+	template := "%v: all of the following fields need to be aligned; if the first value is valid / exists, all the following need to either exist or have the indicated value: %v"
```
ptnapoleon, line 274: Can we possibly extract "the first value" out from `l` here, so it looks like:
```
all of the following fields need to be aligned; if nodeFailure is set, all the following need to either exist or have the indicated value: [level]
```

I think "valid / exists" might be too vague, we can probably assume that if we got here, we're dealing with something valid (or at least something parseable). These messages are primarily for inexperienced users, after all

ddmark/README.md
```diff
@@ -52,10 +52,26 @@ It also allows us to define custom rules to apply to our structures, and focus t
...
+  - without values:
+    -  `// +ddmark:validation:LinkedFieldsValue={<fieldName1>,<fieldName2>,...}`
+    - Applies to: any `<struct>` type
+    - Asserts the fields in the list are either all `nil` or all non-`nil`
+  - with values:
+    - `// +ddmark:validation:LinkedFieldsValue={<fieldName1>=<value>,<fieldName2>,...}`
+    - Applies to: any `<struct>` type -- value checks are only for `string` and `int` subfields
+    - Asserts the fields in the list are either all `nil` (or are *specifically* set to the indicated value) or all non-`nil` (or are ***specifically not*** set to the given value)
+    - Fields with values and without values can be mixed
+- LinkedFieldsValueWithTrigger
```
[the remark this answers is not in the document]
nathantournant: I am not _happy_ with it, but couldn't really think of better either

ddmark/validation.go
nathantournant: good point; I'll adapt that now

nathantournant: done !

ptnapoleon (approved):

ddmark/validation_teststruct.go
luphaz, line 1: if this file is only for test suffix it with `_test.go` so it's not available in the built package.

docs/node_disruption.md
```diff
@@ -5,4 +5,13 @@ The `nodeFailure` field triggers a kernel panic on the node. Because the node wi
...
\ No newline at end of file
+> :warning:️ Node behavior when using this disruption can differ depending on the cloud provider (node may or may not be replaced, restarted, cordoned, etc.).
+
+## Targeting
+
+For clarity purpose, it is mandatory to explicitly set the disruption's `level` field to either `pod` or `node`.
+
+The `nodeFailure` disruption acts on a node. However, the `level` field needs to be adapted for the selector:
+
+- if `level: node` is set, the selector will target nodes and impact those nodes directly.
```
luphaz, line 16: Hence level is somehow misleading and should be part of selector potentially.

If we revamp the spec we should probably have something like:

```yaml
  selector:
    labels:
      test: value
    annotations:
      test: value
    level: node|pod
```

luphaz (approved):

PR #710, opened 2023-05-02
CHAOS-534: Re-resolve hosts while injected during network disruption

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Tested in staging with no notable problems even after more than a week, works as expected in unit test, with close log tracing, and with local testing.
- Not sure if we wanted to modify `docs/changes_handling.md` to mention this updated functionality? 

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


injector/network_disruption.go
```diff
@@ -89,6 +96,11 @@ type serviceWatcher struct {
 	servicesResourceVersion        string
 }
 
+type hostsWatcher struct {
+	// The only identifying info we need are the ip and filter priority
+	hostFilterMap map[v1beta1.NetworkDisruptionHostSpec][]*tcFilter
```
ptnapoleon, line 101: I'll be quite honest that I wasn't sure if it's better to store a map of pointers to (slice of) pointers, structs to (slice of) pointers, or structs to (slice of) structs. I suppose these filter structs are only one `uint32` larger than their own pointer, so probably fine to store directly? The hostspecs should be okay to store directly, if i'm correctly understanding maps

injector/network_disruption.go
```diff
@@ -70,6 +71,12 @@ type tcServiceFilter struct {
 	priority uint32 // one priority per tc filters applied, the priority is the same for all interfaces
 }
 
+// tcFilter describes a tc filter
+type tcFilter struct {
```
ptnapoleon, line 75: Originally I wanted to re-use tcServiceFilter, but the contents of `networkDisruptionService` were a little too coupled with what the k8s api returns when describing a filter, so I thought it was best to define a smaller type with exactly what I need

injector/network_disruption.go
```diff
@@ -479,16 +491,25 @@ func (i *networkDisruptionInjector) addServiceFilters(serviceName string, filter
...
+	i.config.Log.Infow(fmt.Sprintf("deleted a tc filter for service %s with priority %d", tcFilter.service, tcFilter.priority), "interfaces", interfaces)
+
+	return nil
+}
+
+// delete tc filters using only its priority
+func (i *networkDisruptionInjector) removeTcFilter(interfaces []string, priority uint32) error {
 	for _, iface := range interfaces {
-		if err := i.config.TrafficController.DeleteFilter(iface, tcFilter.priority); err != nil {
+		if err := i.config.TrafficController.DeleteFilter(iface, priority); err != nil {
```
luphaz, line 508: Should we halt early and do not try to remove other filter when we are failing to remove one of them?

injector/network_disruption.go
```diff
@@ -479,16 +491,25 @@ func (i *networkDisruptionInjector) addServiceFilters(serviceName string, filter
...
 }
 
-// removeServiceFilter delete tc filters using its priority
+// removeServiceFilter delete tc filters for a k8s service
 func (i *networkDisruptionInjector) removeServiceFilter(interfaces []string, tcFilter tcServiceFilter) error {
+	if err := i.removeTcFilter(interfaces, tcFilter.priority); err != nil {
+		return err
+	}
+
+	i.config.Log.Infow(fmt.Sprintf("deleted a tc filter for service %s with priority %d", tcFilter.service, tcFilter.priority), "interfaces", interfaces)
```
luphaz, line 500:
```suggestion
	i.config.Log.Infow("tc filter deleted for all interfaces", "tcServiceFilter", tcFilter, "interfaces", interfaces)
```

injector/network_disruption.go
```diff
@@ -891,17 +912,110 @@ func (i *networkDisruptionInjector) handleFiltersForServices(interfaces []string
...
+	hosts := hostsWatcher{}
+
+	hostFilterMap, err := i.addFiltersForHosts(interfaces, i.spec.Hosts, flowid)
+	if err != nil {
+		return err
+	}
+
+	hosts.hostFilterMap = hostFilterMap
+
+	go i.watchHostChanges(interfaces, hosts, flowid)
```
luphaz, line 926: You might want to have a context or a channel to stop this goroutine in case of reinject.
You may not want several goroutine trying to apply filters concurrently.

injector/network_disruption.go
```diff
@@ -891,17 +912,110 @@ func (i *networkDisruptionInjector) handleFiltersForServices(interfaces []string
...
+
+	return nil
+}
+
+// watchHostChanges watches for changes to the resolved IP for hosts
+func (i *networkDisruptionInjector) watchHostChanges(interfaces []string, hosts hostsWatcher, flowid string) {
```
luphaz, line 932: Could we maybe only Clean/Reinject periodically?
Potentially with only a hasChange detection mechanism to guard useless reinjection.

IMO it's best to have 1 way to do things, 1 to inject, 1 to clean.

And we trigger those when relevant.

We might do a little more work (checking services DNS twice) but it's less error prone IMO and less tests and less code to maintain.

WDYT?

injector/network_disruption.go
luphaz: I would probably keep the simplest list to be able to reliably and consistently detect changes and changes only.

Then, if any changes, I would trigger a reinject.

Depending on the duration of such process (reinjection, that we will have to measure with the profiler or APM), this might be challenged to ensure the disruption is stable enough.

injector/network_disruption.go
ptnapoleon: We are halting early if any error. We could choose not to, just to remove everything we can? Probably safer to just halt on unexpected errors

injector/network_disruption.go
ptnapoleon: Discussed on slack, but putting here, I'd prefer to keep this approach.
1. I'd rather be consistent with how watchServiceChanges works
2. Adding another async way of triggering clean and reinject is similarly complex to what we're doing here
3. A clean/reinject approach would leave tiny blips of connectivity during disruptions, everytime we had to update a single filter. I'd prefer to try this approach until it actually proves to be buggy

ptnapoleon (commented):
Still re-writing, I'll open for review once I'm done and happy w/ testing

ptnapoleon:
Local testing isn't going terribly well, not worth a re-review yet

ptnapoleon:
I'm happy with how it's working locally, I'm going to begin adding unit tests and deploy it for testing tomorrow

injector/network_disruption_test.go
```diff
@@ -354,6 +355,29 @@ var _ = Describe("Failure", func() {
...
+						Port: 80,
+					},
+				}
+			})
+
+			It("should update the filters", func() {
+				tc.AssertCalled(GinkgoT(), "AddFilter", []string{"lo", "eth0", "eth1"}, "1:0", "", nilIPNet, buildSingleIPNet(testHostIP), 0, 80, network.TCP, network.ConnStateUndefined, "1:4")
+
+				const newTestHostIP = "2.2.2.2"
+				dns.EXPECT().Resolve("testhost").Return([]net.IP{net.ParseIP(newTestHostIP)}, nil).Maybe()
```
ptnapoleon, line 372: No matter how long I wait, I don't ever see this change take "effect" within `network_disruption.go`. It continues to resolve "testhost" as 1.1.1.1. /cc @luphaz

injector/network_disruption.go
```diff
@@ -66,6 +69,30 @@ type tcServiceFilter struct {
...
+func (t tcFilter) String() string {
+	ip := ""
+	if t.ip != nil {
+		ip = t.ip.String()
+	}
+
+	return fmt.Sprintf("ip=%s; priority=%s", ip, strconv.FormatUint(uint64(t.priority), 10))
+}
+
+func FiltersToStrings(t []tcFilter) []string {
```
ptnapoleon, line 87: I don't really understand the reason, but even though `tcFilter` has a String(), and I can print one in a log just fine, trying to print a slice of them just gives me `{[], [], []}`, thus this function

injector/network_disruption.go
```diff
@@ -886,17 +933,133 @@ func (i *networkDisruptionInjector) handleFiltersForServices(interfaces []string
...
+					// If we have more or fewer IPs than before, we obviously have a change and need to update the tc filters
+					changedHosts = append(changedHosts, host)
+				}
+			}
+
+			if len(changedHosts) > 0 {
+				for _, changedHost := range changedHosts {
+					for _, filter := range hosts.hostFilterMap[changedHost] {
+						err := i.removeTcFilter(interfaces, filter.priority)
+						if err != nil {
```
luphaz, line 1009:
```suggestion
						
						if err := i.removeTcFilter(interfaces, filter.priority); err != nil {
```

injector/network_disruption_test.go
```diff
@@ -354,6 +354,33 @@ var _ = Describe("Failure", func() {
...
+			BeforeEach(func() {
+				spec.Hosts = []v1beta1.NetworkDisruptionHostSpec{
+					{
+						Host: "testhost",
+						Port: 80,
+					},
+				}
+
+				dns.EXPECT().Resolve("testhost").Return([]net.IP{net.ParseIP(testHostIP)}, nil).Once()
+			})
```
luphaz, line 367: Could we maybe add more tests to illustrate the statements we have in the code:
* LEN oldIPS <> LEN newIPs
* ip does not exists in new ip (maybe this test?)
* no changes in IPs

injector/network_disruption.go
luphaz: It seems due to the fact behing the scene zap uses json marshalling of the struct:
* https://go.dev/play/p/4aOAhZSh7Va

Hence you should have exported fields at least to have something displayed when using it.

You could still have a dedicated type `[]tcFilter` and associate a custom marshaller to reuse the `Stringer` method of the underlying private type if you prefer.

luphaz (approved):
💪 great we are now reactive to host changes.

Thanks for the changes added there that match the existing services implementation we have.

IMO, this whole update mechanism should NOT be part of the injector but a dedicated controller instead.
Such controller would watch those changes and create the resource as simple as we need (a configmap with entries) to be reused by our pod.

This is a deeper refactoring related to the 101 pres you did not had yet, I'm fine to first merging this to improve the existing situation, and would love to discuss about next iterations to move all external calls made by `injectors` somewhere else (in a dedicated reconciliation loop)

Hence ✅

PR #704, opened 2023-04-20
feat(injector/cpu-pressure): revamp

luphaz:
## What does this PR do?

- [x] Adds new functionality
- [x] Alters existing functionality
- [x] Fixes a bug
- [x] Improves documentation or testing

Please briefly describe your changes as well as the motivation behind them:
- This PR aims to adapt the `cpu-pressure` injector to be able to react to container restarts
- This PR also fixes couple of bugs seen along the way:
  - Exit without error
  - Stress all containers not only the last
  - Stress all cores on all containers
  - split DisruptionInjectionStatus and DisruptionTargetInjectionStatus to avoid miss understanding of which is allowed to have what.
  - add some new DisruptionInjectionStatus to avoid having a misleading `PreviouslyInjected` status (when it was partially injected or not injected, an expired disruption changed to the same status, it's no longer the case)
  - fixes when the `previously` injection status is calculated to ensure when injection pods exited successfully near the deadline we are not transitioning back to `PartiallyInjected`

This PR also takes the opportunity to (re)-introduce a `make local` rule to run the controller outside of Kubernetes.
This PR also introduce a `make pre-debug` that will like `make local` except it won't launch the binary and will let you use you favorite debugger method (IDE, dlv, ...)

This PR also review how we are running e2e-test to guarantee pods are created for each test ran and ease their duplication.
E2E-tests are hence now ran in parallel.

## Code Quality Checklist

- [x] The documentation is up to date.
- [x] My code is sufficiently commented and passes continuous integration checks.
- [x] I have signed my commit (see [Contributing Docs](../CONTRIBUTING.md)).

## Testing

- [x] I leveraged continuous integration testing
    - [x] by depending on existing `unit` tests or `end-to-end` tests.
    - [x] by adding new `unit` tests or `end-to-end` tests.
- [x] I manually tested the following steps:
    - [x] locally.
    - [ ] as a canary deployment to a cluster.

PR #712, opened 2023-05-04
CHAOS-618: Proposed layout for controlling the triggering of chaos pod creation and injection

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- My use case is it takes a lot of time to schedule hundreds of chaos pods. If I want them all to start their disruption at the same time, I need to be passing a time, not a duration.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


examples/trigger.yaml
```diff
@@ -0,0 +1,27 @@
...
+  namespace: chaos-demo
+  annotations:
+    chaos.datadoghq.com/environment: "lima"
+spec:
+  level: pod
+  selector:
+    app: demo-curl
+  count: 100%
+  staticTargeting: true
+  trigger:
```
ptnapoleon, line 19: @luphaz How do you feel about this structure? Let's us control injection and pod creation separately.

luphaz: Ok, so here inject means when the disruption is effectively applied, that relates to what we have today in the status as `Injected`?

And `pods` is, when will I create the pods for the given disruption?

Inject feels OK, however `pods` I'm not sure to what it relates to 🤔 

It's seems OK for me to split both however as they are not the same thing at all, and it make sense IMO to clearly isolate them.

Maybe `create` instead of pods? `deploy`, `run`?

ptnapoleon: The names are definitely insufficient, I always put that last. But seems like you're good with splitting them like this, so I'll move forward with implementing and we can debate names as a team

api/v1beta1/disruption_types.go
```diff
@@ -313,6 +331,47 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() (retErr error) {
...
+			if !subTrigger.NotBefore.IsZero() {
+				now := metav1.Now()
+				if subTrigger.NotBefore.Before(&now) {
+					retErr = multierror.Append(retErr,
+						fmt.Errorf("you should not set spec.trigger.*.notBefore to a time in the past. spec.trigger.*.notBefore: %s, current timestamp: %s",
+							subTrigger.NotBefore.String(),
+							now.String()))
+				}
+
+				if subTrigger.NotBefore.Sub(now.Time) > s.Duration.Duration() {
```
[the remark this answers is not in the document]
luphaz: Same for pod not before in the end IMO.

I would say:
* podNotBefore
* injectionNotBefore
* duration starting

If only one of those is defined, duration starts right after.

api/v1beta1/disruption_webhook_test.go
```diff
@@ -252,6 +252,74 @@ var _ = Describe("Disruption", func() {
...
+
+			JustBeforeEach(func() {
+				newDisruption = makeValidNetworkDisruption()
+				controllerutil.AddFinalizer(newDisruption, chaostypes.DisruptionFinalizer)
+			})
+
+			AfterEach(func() {
+				k8sClient = nil
+				newDisruption = nil
+			})
```
luphaz, line 269: Favour DeferCleanup usage to have cleanup close to setup.

I'm not sure why it's needed TBH as you are already re-creating those everytime.

Maybe, create both the client AND the disruption in the `BeforeEach` hence you will be able to modify them in other `BeforeEach` 

And you can remove the just before each (as finalizers can be moved to `BeforeEach` also IMO)

api/v1beta1/disruption_webhook_test.go
```diff
@@ -252,6 +252,74 @@ var _ = Describe("Disruption", func() {
...
+					newDisruption.Spec.Duration = "10m"
+					newDisruption.Spec.Trigger = &DisruptionTriggers{
+						Inject: &DisruptionTrigger{
+							NotBefore: metav1.NewTime(time.Now().Add(time.Minute * 15)),
+						},
+					}
+
+					err := newDisruption.ValidateCreate()
+
+					Expect(err).Should(HaveOccurred())
```
luphaz, line 282: Prefer:
```suggestion
					Expect(newDisruption.ValidateCreate()).Should(MatchError("you should not set spec.trigger.*.notBefore to a time farther in the future than the disruption duration"))
```

It guarantees error has occurred AND contains expected error text.

luphaz: You might want to replace all your tests by [Ginkgo table testing](https://onsi.github.io/ginkgo/#table-specs) (you can still use BeforeEach with that):

```suggestion
DescribeTable("spec.trigger possible errors", func(duration string, pods, inject *DisruptionTrigger, wantError string){
  newDisruption.Spec.Duration = duration
					newDisruption.Spec.Trigger = &DisruptionTriggers{
						Pods: pods,
						Inject: inject,
					}

  Expect(newDisruption.ValidateCreate()Should(MatchError(wantError))
},
Entry("notBefore is set past duration", "10m", nil, &DisruptionTrigger{
							NotBefore: metav1.NewTime(time.Now().Add(time.Minute * 15)),
						}, "you should not set spec.trigger.*.notBefore to a time farther in the future than the disruption duration"),
Entry("offset is set larger than duration", "10m", &DisruptionTrigger{
							NotBefore: metav1.NewTime(time.Now().Add(time.Minute * 15)),
						}, nil, "you should not set spec.trigger.*.offset higher than the disruption duration"))
```

api/validator_test.go
```diff
@@ -34,6 +34,24 @@ var _ = Describe("Validator", func() {
 		errList = ValidateDisruptionSpecFromString(yamlDisruptionSpec.String())
 	})
 
+	Describe("validating disruption triggers", func() {
+		Context("both offset and notBefore are set", func() {
+			BeforeEach(func() {
+				yamlDisruptionSpec.WriteString("\nnetwork:")
```
luphaz, line 40: for readability you might want to use HEREDOC style strings, just be careful to NOT let your autoformatter re-indent the string for you.:

```suggestion
yamlDisruptionSpec.WriteString(`
network:
  corrupt: 100
  duration: 87600h
  triggers
    pods:
      notBefore: 2040-01-02T15:04:05-04:00
      offset: 1m
`)
```

controllers/disruption_controller.go
```diff
@@ -260,6 +260,25 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+
+			if instance.Spec.Trigger.Pods.Offset.Duration() > 0 {
+				noPodsBefore = metav1.NewTime(instance.CreationTimestamp.Add(instance.Spec.Trigger.Pods.Offset.Duration()))
+			}
+
+			if now.Before(&noPodsBefore) {
+				r.log.Debugw("requeuing disruption as we haven't yet reached trigger.pods", "trigger.pods", noPodsBefore)
+
+				return ctrl.Result{Requeue: false}, nil
```
luphaz, line 278: You might want to specify a requeue time instead.

controllers/disruption_controller.go
```diff
@@ -260,6 +260,25 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{Requeue: false}, nil
 		}
 
+		// check if we have reached trigger.pods. If not, skip the rest of reconciliation.
```
luphaz, line 263: I would extract the logic to a method to ease readability and test it.

controllers/disruption_controller.go
```diff
@@ -1246,6 +1265,29 @@ func (r *DisruptionReconciler) generateChaosPods(instance *chaosv1beta1.Disrupti
 			pulseDormantDuration = instance.Spec.Pulse.DormantDuration.Duration()
 		}
 
+		// spec.trigger.inject.notBefore will be used as an argument to tell the chaos pods the earliest timestamp (in milliseconds from Unix epoch) that they can inject
+		// deeply assumes that the chaos-controller and chaos pods are using the same timestamp, but that's okay because anyone responsible uses UTC for all their machines
+		var notInjectedBefore int64
+
+		if instance.Spec.Trigger != nil {
```
luphaz, line 1272: I would probably also create a method here so it will ease adaptation and testing and improve readabilty.

cli/injector/main.go
```diff
@@ -490,6 +491,18 @@ func injectAndWait(cmd *cobra.Command, args []string) {
 		return
 	}
 
+	if disruptionArgs.NotInjectedBefore > 0 {
```
luphaz, line 494: I love how simple it is in the injector to do this 💪

api/v1beta1/disruption_types.go
```diff
@@ -43,11 +43,12 @@ type DisruptionSpec struct {
...
-	DryRun          bool              `json:"dryRun,omitempty"`          // enable dry-run mode
-	OnInit          bool              `json:"onInit,omitempty"`          // enable disruption on init
-	Unsafemode      *UnsafemodeSpec   `json:"unsafeMode,omitempty"`      // unsafemode spec used to turn off safemode safety nets
-	StaticTargeting bool              `json:"staticTargeting,omitempty"` // enable dynamic targeting and cluster observation
+	Filter          *DisruptionFilter   `json:"filter,omitempty"`
+	DryRun          bool                `json:"dryRun,omitempty"`          // enable dry-run mode
+	OnInit          bool                `json:"onInit,omitempty"`          // enable disruption on init
+	Unsafemode      *UnsafemodeSpec     `json:"unsafeMode,omitempty"`      // unsafemode spec used to turn off safemode safety nets
+	StaticTargeting bool                `json:"staticTargeting,omitempty"` // enable dynamic targeting and cluster observation
+	Trigger         *DisruptionTriggers `json:"triggers,omitempty"`        // alter the pre-injection lifecycle
```
luphaz, line 51: So maybe let's keep Triggers for the field also?
```suggestion
	Triggers         *DisruptionTriggers `json:"triggers,omitempty"`        // alter the pre-injection lifecycle
```

api/v1beta1/disruption_types.go
```diff
@@ -75,9 +76,26 @@ type DisruptionSpec struct {
 	Reporting *Reporting `json:"reporting,omitempty"`
 }
 
+// DisruptionTrigger holds the options for changing when injector pods are created, and the timing of when the injection occurs
+// +ddmark:validation:ExclusiveFields={NotInjectedBefore,Offset}
+type DisruptionTriggers struct {
+	Inject *DisruptionTrigger `json:"inject,omitempty"`
+	// inject.notBefore: Normal reconciliation and chaos pod creation will occur, but chaos pods will wait to inject until NotInjectedBefore. Must be after NoPodsBefore if both are specified
+	// inject.offset: Identical to NotBefore, but specified as an offset from max(CreationTimestamp, NoPodsBefore) instead of as a metav1.Time
+	Pods *DisruptionTrigger `json:"pods,omitempty"`
```
luphaz, line 85: I would rename this to CreatePods probably.

api/disruption_kind.go
```diff
@@ -41,6 +42,7 @@ type DisruptionArgs struct {
 	PulseInitialDelay    time.Duration
 	PulseActiveDuration  time.Duration
 	PulseDormantDuration time.Duration
+	NotInjectedBefore    int64
```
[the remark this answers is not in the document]
ptnapoleon: RFC3339 is certainly easier for humans to engage with, but this is a machine <--> machine interaction, unix epoch milliseconds seems simplest? If you feel rfc3339 nice is safer, I'll go ahead and build that custom type in this PR

api/v1beta1/disruption_types.go
```diff
@@ -75,9 +76,26 @@ type DisruptionSpec struct {
 	Reporting *Reporting `json:"reporting,omitempty"`
 }
 
+// DisruptionTrigger holds the options for changing when injector pods are created, and the timing of when the injection occurs
+// +ddmark:validation:ExclusiveFields={NotInjectedBefore,Offset}
+type DisruptionTriggers struct {
+	Inject *DisruptionTrigger `json:"inject,omitempty"`
+	// inject.notBefore: Normal reconciliation and chaos pod creation will occur, but chaos pods will wait to inject until NotInjectedBefore. Must be after NoPodsBefore if both are specified
```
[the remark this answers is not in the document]
ptnapoleon: Given these fields have different semantics based on if they're `Inject` or `Pod`, but they're both the same struct type, I wasn't sure what the right place to put these comments is. Just directly in DisruptionTrigger?

controllers/disruption_controller.go
ptnapoleon: I have personally found that, in practice, the requeue times aren't respected. I see the disruptions we requeue for duration or gc delay to still get reconciled every few minutes. But I'll put in the calculation for appropriate requeue time

api/validator_test.go
ptnapoleon: I'm having a lot of trouble with this approach ^, going to leave it as is

api/disruption_kind.go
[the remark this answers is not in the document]
ptnapoleon: Looking into this, we would need to add a custom type that implements:
```
// Value is the interface to the dynamic value stored in a flag.
// (The default value is represented as a string.)
type Value interface {
	String() string
	Set(string) error
	Type() string
}
```

How will we want to handle times that don't come in a correct rfc3339 format? Do we want our custom time to wrap metav1.Time or just time.Time?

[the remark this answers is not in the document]
luphaz: My thoughts about this:
* RFC3339 is also used by [protoc.JSON](https://protobuf.dev/programming-guides/proto3/#json) marshalling hence safe for machine to machine interaction
* it's way more understandable for a human during a troubleshooting (looking at injector arguments, logs etc)
* it's safe as long as properly tested and injected
* it's a widely known and implemented format baked by a serious standard

Hence I would use this whenever possible.

controllers/disruption_controller.go
luphaz: I think the requeue is respected only if there is no other events on the object.

TBH I have not enough knowledge.

I would only say it gives an hint to what is expected and might reduce the controller (at least we can hope so).

api/disruption_kind.go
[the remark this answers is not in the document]
luphaz: FYI I have implemented a custom type with cobra baked in deserialization.

Hoping to open it soon.

I will isolate the commit so you can stole it here and advance without being blocked.

api/v1beta1/disruption_types.go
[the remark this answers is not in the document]
luphaz: 🤔 Good to know.

What I'm wondering is how common we expect those fields to be (Inject/Pod) and how could they evolve in different manner together.

If the semantic is the same and does not mean the same... Maybe I would change to have two separated structs with appropriate comments and DIFFERENT field name to enhance readability and comprehension and simplify maintenance.

api/disruption_kind.go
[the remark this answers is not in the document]
luphaz: > Looking into this, we would need to add a custom type that implements:
> 
> ```
> // Value is the interface to the dynamic value stored in a flag.
> // (The default value is represented as a string.)
> type Value interface {
> 	String() string
> 	Set(string) error
> 	Type() string
> }
> ```
> 
> How will we want to handle times that don't come in a correct rfc3339 format? Do we want our custom time to wrap metav1.Time or just time.Time?

So, the implementation of such interface is [here](https://github.com/DataDog/chaos-controller/pull/704/commits/f9c83360393c908b4059470834f4432b8dec9c5d), usage [here](https://github.com/DataDog/chaos-controller/pull/704/commits/4c9fb2b5faa6ceb870c3cf6580f617679a81cda1) 

On the controller side I think we can stick to "time" (can we?) as it's already handling all the heavy parsing for us.

I would stick to use this internal type for Injector usage.

Above linked implementation might miss Disruption POV marshalling (providing a time and convert it to the expected format expected by the injector)

api/validator_test.go
ptnapoleon: Okay, all I had to do was give up on editing this file in GoLand and do it in vim.

api/validator_test.go
```diff
@@ -34,6 +34,26 @@ var _ = Describe("Validator", func() {
...
+	Describe("validating disruption triggers", func() {
+		Context("both offset and notBefore are set", func() {
+			BeforeEach(func() {
+yamlDisruptionSpec.WriteString(`
+network:
+  corrupt: 100
+duration: 87600h
+triggers:
+  createPods:
+    notBefore: 2040-01-02T15:04:05-04:00
```
ptnapoleon, line 46: This test will start failing in 2040 but I am okay with that

ptnapoleon:
I'm going to begin testing this more thoroughly in staging, but this should be robust enough for review

api/v1beta1/disruption_types.go
```diff
@@ -44,11 +44,12 @@ type DisruptionSpec struct {
...
-	DryRun          bool              `json:"dryRun,omitempty"`          // enable dry-run mode
-	OnInit          bool              `json:"onInit,omitempty"`          // enable disruption on init
-	Unsafemode      *UnsafemodeSpec   `json:"unsafeMode,omitempty"`      // unsafemode spec used to turn off safemode safety nets
-	StaticTargeting bool              `json:"staticTargeting,omitempty"` // enable dynamic targeting and cluster observation
+	Filter          *DisruptionFilter   `json:"filter,omitempty"`
+	DryRun          bool                `json:"dryRun,omitempty"`          // enable dry-run mode
+	OnInit          bool                `json:"onInit,omitempty"`          // enable disruption on init
+	Unsafemode      *UnsafemodeSpec     `json:"unsafeMode,omitempty"`      // unsafemode spec used to turn off safemode safety nets
+	StaticTargeting bool                `json:"staticTargeting,omitempty"` // enable dynamic targeting and cluster observation
+	Triggers        *DisruptionTriggers `json:"triggers,omitempty"`        // alter the pre-injection lifecycle
```
luphaz, line 52: We do we need a pointer?
Once disruption is created is it allowed to modify triggers?

api/v1beta1/disruption_types.go
```diff
@@ -76,9 +77,27 @@ type DisruptionSpec struct {
 	Reporting *Reporting `json:"reporting,omitempty"`
 }
 
+// DisruptionTriggers holds the options for changing when injector pods are created, and the timing of when the injection occurs
+type DisruptionTriggers struct {
+	Inject     *DisruptionTrigger `json:"inject,omitempty"`
```
luphaz, line 82: Once trigger is defined, I think it's immutable, hence no pointer needed IMO

api/v1beta1/disruption_types.go
```diff
@@ -314,6 +333,33 @@ func (s *DisruptionSpec) validateGlobalDisruptionScope() (retErr error) {
...
+		triggerTypes := []*DisruptionTrigger{s.Triggers.Inject, s.Triggers.CreatePods}
+
+		for _, subTrigger := range triggerTypes {
+			if subTrigger == nil {
+				continue
+			}
+
+			if !subTrigger.NotBefore.IsZero() {
+				now := metav1.Now()
+				if subTrigger.NotBefore.Before(&now) {
```
luphaz, line 353: How close to now can we be?
How far from now do we allow?

Maybe let's enforce at least 1m from now, and no more than Xh ?

WDYT?

api/v1beta1/disruption_webhook_test.go
```diff
@@ -250,6 +250,22 @@ var _ = Describe("Disruption", func() {
...
+					newDisruption.Spec.Triggers = &DisruptionTriggers{
+						Inject: &DisruptionTrigger{
+							NotBefore: metav1.NewTime(time.Now().Add(time.Minute * 5)),
+						},
+						CreatePods: &DisruptionTrigger{
+							NotBefore: metav1.NewTime(time.Now().Add(time.Minute * 15)),
+						},
+					}
+
+					Expect(newDisruption.ValidateCreate().Error()).Should(ContainSubstring("inject.notBefore must come after createPods.notBefore if both are specified"))
```
luphaz, line 266:
```suggestion
					Expect(newDisruption.ValidateCreate()).Should(MatchError("inject.notBefore must come after createPods.notBefore if both are specified"))
```

This works also

controllers/disruption_controller.go
```diff
@@ -260,6 +260,14 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{Requeue: false}, nil
 		}
 
+		// check if we have reached trigger.createPods. If not, skip the rest of reconciliation.
+		requeueAfter := time.Until(TimeToCreatePods(instance.Spec.Triggers, instance.CreationTimestamp.Time))
+		if requeueAfter > 0 {
+			r.log.Debugw("requeuing disruption as we haven't yet reached trigger.createPods", "requeueAfter", requeueAfter.String())
```
luphaz, line 266:
```suggestion
			r.log.Debugw("requeuing disruption as we haven't yet reached trigger.createPods", "requeueAfter", requeueAfter)
```

controllers/disruption_controller.go
```diff
@@ -260,6 +260,14 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+			return ctrl.Result{Requeue: false, RequeueAfter: requeueAfter}, nil
```
luphaz, line 268:
```suggestion
			return ctrl.Result{RequeueAfter: requeueAfter}, nil
```[
RequeueAfter imply](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go#LL32C31-L32C31) `requeue: true`

File: docs/features.md [hunk 21]
```diff
@@ -31,6 +31,30 @@ If a `duration` is not specified, then a disruption will receive the default dur
 After a disruption's duration expires, the disruption resource will live in k8s for a default of 10 minutes. This can be configured by altering
 `controller.expiredDisruptionGCDelay` in the controller's config map.
 
+If any of the options in `spec.triggers` are set, the `duration` will not "begin" until _after_ the injection starts. See [below for details.](#Triggers-aka-controlling-the-timing-of-chaos-pod-creation-and-injection)
+
+## Triggers, aka, Controlling the timing of chaos pod creation and injection
+
+The `Disruption` spec has a `triggers` field, with two subfields: `createPods` and `inject`. These have identical subfields of their own: `notBefore` and `offset`. `spec.triggers` as well as all of its subfields, are strictly optional.
```
luphaz, line 38: 😍 thanks for this detailed explanation 💪 

It's crystal clear

luphaz (approved):
Can't wait to use this 💪 

I made some fixes to pflag that failed due to go 1.18.

You might want to retrieve also `chore: go 1.20` to make it work here.

api/v1beta1/disruption_types.go
ptnapoleon: We call validateGlobalDisruptionScope on Updates as well as Creates, unfortunately, so we can't compare against time.Now, and we don't have access to the creationTimestamp here. We could move this check to the reconcile loop, but I think maybe we just leave it, and plug the gap if it actually becomes an issue?

ptnapoleon:
@luphaz Do you expect https://github.com/DataDog/chaos-controller/pull/712#discussion_r1192720635 to work? It fails locally for me and here in CI, because `MatchError` seems to be looking for a full string match, not a substring?

luphaz:
> @luphaz Do you expect [#712 (comment)](https://github.com/DataDog/chaos-controller/pull/712#discussion_r1192720635) to work? It fails locally for me and here in CI, because `MatchError` seems to be looking for a full string match, not a substring?

🤔 I thought I tried locally, indeed it's a full match not a substring, sorry if I wrongly test this.

api/v1beta1/disruption_types.go
luphaz: No problem with this approach 🤗

PR #714, opened 2023-05-15
test: upload test results to Datadog if available

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- This PR aims to upload junit test reports generated by Gingko to Datadog

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


api/v1beta1/disruption_types.go
```diff
@@ -101,7 +101,7 @@ type Reporting struct {
 
 // EmbeddedChaosAPI includes the library so it can be statically exported to chaosli
 //
-//go:embed *
+//go:embed *.go
```
luphaz, line 104: Not sure but building is not at all possible if I do not change this 🤔

.circleci/config.yml
```diff
@@ -111,8 +111,8 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Install golangci-lint kubebuilder+env helm
-          command: make -j4 install-lint-deps install-kubebuilder install-helm install-controller-gen
+          name: Install binaries
+          command: make -j5 install-lint-deps install-kubebuilder install-helm install-controller-gen install-datadog-ci
```
taihuynh167, line 115: `j5` is it a type of instance?

CONTRIBUTING.md
```diff
@@ -149,6 +149,26 @@ The project contains end-to-end test which are meant to run against a real Kuber
 
 The end-to-end tests will create a set of dummy pods in the `default` namespace of the Lima cluster that will be used to inject different kind of failures and make different assertions against those injections.
 
+### Uploading test results to datadog
+
+In case you have a Datadog account and want to push the tests results to it, you can do the following:
```
taihuynh167, line 154: Do I want to push the test results?
----
When this runs in Gitlab, what key does it use to push results to DD? A team account API key?

taihuynh167 (approved):

CONTRIBUTING.md
luphaz: Per documentation [here](https://docs.datadoghq.com/continuous_integration/tests/junit_upload/?tab=linux#uploading-test-reports):

> for example, `local` when uploading results from a developer workstation, or ci when uploading them from a CI provider) in the DD_ENV environment variable. For example:

I followed such guidelines.

> 1. Do I want to push the test results?

Up to you TBH, if it helps detecting issue faster and giving timing comparison upon us, etc why you wouldn't want do it?

> 2. When this runs in Gitlab, what key does it use to push results to DD?

This does not run in GitLab but in CircleCI. The key used is a team key indeed.
That I'm also using locally TBH as I'm NOT allowed to create an API key (only an application key)

---

The keys does not really matter for the underlying instrumentation that is git based, it's only needed to have a key with enough permissions to upload an xml file.

Even if not ideal, I think it's OK for now

I still asked the question to have guidance around this.

.circleci/config.yml
luphaz: `-j` is a make flag that allows to run several make target in [parallel](https://www.gnu.org/software/make/manual/html_node/Parallel.html).

Here we install all program in parallel (low CPU, better to have high concurrency for network calls)

PR #734, opened 2023-07-03
feat(container): make chaos GBI compliant

luphaz:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- This PR aims to harden the security of the container images we provide by reducing the attack surface by transitioning to "distroless" images
  - `handler` was already "scratch"
  - `manager` is now based on `gcr.io/distroless/base-debian11:nonroot` AND contains go runtime in order to be able to run our custom validation tool `ddmark`
  - `injector` is now based on `gcr.io/distroless/python3-debian11:latest` AND contains **GLIBC** tc/iptables libraries and a precise list of binaries we are using (directly or indirectly): `tc,ls,df,iptables`  

Deployment manifest has also been modified to:
- run container and proxy as `nonroot` users and groups
- activate read only root filesystem (and introduce a dedicated volume to handle file writing inside the container)
- forbid privilege escalation
- activate runAsNonRoot 

## Code Quality Checklist

## Testing

    - [x] locally.

Disruption tested following this change:
* [x] cpu failure
* [x] disk failure
* [x] disk pressure read/write
* [x] dns
* network_disruption
  * [x] drop with host/port filter
  * [x] bandwidth limitation
  * [x] corrupt
  * [x] cloud
  * [x] delay (pod/node)
  * [x] drop
  * [x] filter (service/connstate)
  * [x] duplication
  * [x] ingress
  * [x] oninit
  * [x] pulse
* [x] node failure (restart/shutdown)


aymericDD (approved):
LGTM 👍

chart/templates/generated/role.yaml
```diff
@@ -10,7 +10,6 @@ rules:
     verbs:
       - create
       - delete
-      - get
```
Devatoria, line 13: Hm there's a chance this `get` being removed from the generated file was in fact manually added previously for a reason, we might want to add it to rules defined through kubebuilder to keep it

luphaz: I think it's not relevant due to how informer works
Informer never does get, it does list+watch
all the calls are referenced here, none of the ones done through chaos-controller are using a standard k8s client

TBH, I changed this because we lacked a get (for the services) and it worked without it

I will try without on staging and see how it goes

luphaz: In the end I took the approach you suggested and [added](https://github.com/DataDog/chaos-controller/pull/734/commits/90ad0edabb1deb89421d3998ebb6bc3156cb3720) the permissions as before 😄 
I don't want to retest all disruptions again and will fix this at a later stage as it's not harming in any way today.

ptnapoleon (approved):

PR #742, opened 2023-07-17
Disruption schedules

diyarab:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

### Enhancing Chaos Controller: Scheduled Disruptions with DisruptionCron
This PR significantly enhances the chaos controller by introducing the DisruptionCron custom resource. The `DisruptionCron` resource extends the capability of chaos experiments by enabling users to schedule disruptions in a cron-like manner on long-lived Kubernetes resources such as deployments and stateful sets. This enhancement provides more flexibility for users, offering a robust and adaptable way to manage and schedule chaos experiments in Kubernetes environments.

### Highlights:
1. The `DisruptionCron` schema has been defined, including the metadata, spec, and status. The spec includes the schedule for disruptions, starting deadline, a target resource specification, and the disruption template. The status includes the last time the disruption was scheduled successfully and the time when the target resource was previously missing.
2. The `Reconcile` function for the `DisruptionCronReconciler` has been implemented, with the logic for checking target resource existence, updating the disruption cron status, and creating disruptions from the provided template.
3. The same disruption name for every execution, acting as a lock to prevent creating the disruption twice.

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// Specifies the Disruption that will be created when executing a DisruptionShedule.
+	DisruptionTemplate Disruption `json:"disruptionTemplate"`
+}
+
+// TargetResource specifies the long-lived resource to be targeted for disruptions.
+// Disruptions are intended to exist semi-permanently, and thus appropriate targets can only be other long-lived resources,
+// such as StatefulSets or Deployments.
+type TargetResource struct {
+	// Kind specifies the type of the long-lived resource. Allowed values: "Deployment", "StatefulSet".
+	// +kubebuilder:validation:Enum=Deployment;StatefulSet
```
ptnapoleon, line 53: Where possible (i think just here in this file, that i can see), you'll want to use ddmark validation in addition to kubebuilder. You can see examples in the structs for Disruption

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,68 @@
...
+// +kubebuilder:object:root=true
+
+// DisruptionScheduleList contains a list of DisruptionSchedule
+type DisruptionScheduleList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []DisruptionSchedule `json:"items"`
+}
+
+type DisruptionScheduleSpec struct {
```
ptnapoleon, line 35: I think you'll want to go through every suboption within this spec and consider whether they need to be `required`, and if not, if they should be `omitempty`. Required just means that the user has to explicitly specify them, it's fine to have "required" options that have sensible defaults when not specified.

`omitempty` is a json tag that comes after you specify the name, similar to metadata at Line 31, which means that when not specified, we don't include this in any json representations of this struct

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,68 @@
...
+	//+kubebuilder:validation:MinLength=0
+	// The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+	Schedule string `json:"schedule"`
+
+	// TargetResource specifies the resource to run disruptions against.
+	// It can only be a Deployment or StatefulSet.
+	TargetResource TargetResource `json:"targetResource"`
+
+	// Specifies the Disruption that will be created when executing a DisruptionShedule.
+	DisruptionTemplate Disruption `json:"disruptionTemplate"`
```
ptnapoleon, line 45: I'm reasonably confident that this should be a DisruptionSpec, not a Disruption. You wouldn't want to embed a Status, or Metadata within here, just the specific disruption that the user wants to run

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// Name specifies the name of the specific instance of the long-lived resource to be targeted.
+	Name string `json:"name"`
+}
+
+type DisruptionScheduleStatus struct {
+	// The last time when the disruption was last successfully scheduled.
+	// +optional
+	LastScheduleTime *metav1.Time `json:"lastScheduleTime,omitempty"`
+
+	// Time when the target resource was previously missing.
```
ptnapoleon, line 65: Why not choose a more generic field here, such as last failed schedule, given the other field is last successful schedule. This doesn't leave us space in the status to show failed schedules for reasons other than missing targets

ptnapoleon: You may even want a field like "lastFailedSchedule" that includes a time _and_ a `reason` string field

ptnapoleon: Okay, I now see why we want this specific PreviouslyMissing field, to handle cleanup of an orphaned disruption. I take back my suggestion

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,201 @@
...
+
+	instance := &chaosv1beta1.DisruptionSchedule{}
+
+	// retrieve schedule
+	if err := r.Client.Get(context.Background(), req.NamespacedName, instance); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if !instance.DeletionTimestamp.IsZero() {
+		// NOTE: add a finalizer if anything needs to be written here
```
ptnapoleon, line 49: It can be a TODO for later, but I do think we'll want a finalizer. If anyone deletes the schedule, we will want to delete any ongoing disruptions. That... should be somewhat easy to do, as they'll be in the same namespace, so we can make the disruption a child of this resource, and I think the Delete will just propagate down?

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+func (r *DisruptionScheduleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
+	r.log = r.BaseLog.With("scheduleNamespace", req.Namespace, "scheduleName", req.Name)
+	r.log.Info("Reconciling DisruptionSchedule")
+
+	instance := &chaosv1beta1.DisruptionSchedule{}
+
+	// retrieve schedule
+	if err := r.Client.Get(context.Background(), req.NamespacedName, instance); err != nil {
```
ptnapoleon, line 44:
```suggestion
	if err := r.Client.Get(ctx, req.NamespacedName, instance); err != nil {
```

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+	err := r.Client.Get(ctx, types.NamespacedName{
+		Name:      instance.Spec.TargetResource.Name,
+		Namespace: instance.Namespace,
+	}, targetObj)
+
+	return errors.IsNotFound(err), err
+}
+
+func (r *DisruptionScheduleReconciler) updateTargetResourcePreviouslyMissing(ctx context.Context, instance *chaosv1beta1.DisruptionSchedule) (bool, bool, error) {
```
ptnapoleon, line 160: This function needs a godoc explaining what the first two return values indicate

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	disruptions, err := r.getChildDisruptions(ctx, instance)
+	if err != nil {
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.updateLastScheduleTime(ctx, instance, disruptions); err != nil {
+		r.log.Errorw("unable to update LastScheduleTime of DisruptionSchedule status", "DisruptionSchedule", instance.Name, "err", err)
```
ptnapoleon, line 60:
```suggestion
		r.log.Errorw("unable to update LastScheduleTime of DisruptionSchedule status", "err", err)
```

`scheduleName: instance.Name` has already been added to every log, after the `r.log = r.BaseLog.With("scheduleNamespace", req.Namespace, "scheduleName", req.Name)` call on L38

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	BaseLog *zap.SugaredLogger
+	log     *zap.SugaredLogger
+}
+
+const (
+	ScheduledAtAnnotation       = chaosv1beta1.GroupName + "/scheduled-at"
+	DisruptionScheduleNameLabel = chaosv1beta1.GroupName + "/disruption-schedule-name"
+)
+
+func (r *DisruptionScheduleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
```
ptnapoleon, line 37: You'll want to set up a deferred handling of panics and errors:
https://github.com/DataDog/chaos-controller/blob/main/controllers/disruption_controller.go#L121-L151

So that any retryable errors get safely retried

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	if err := r.updateLastScheduleTime(ctx, instance, disruptions); err != nil {
+		r.log.Errorw("unable to update LastScheduleTime of DisruptionSchedule status", "DisruptionSchedule", instance.Name, "err", err)
+		return ctrl.Result{}, err
+	}
+
+	// _ is targetResourceNotFound, which will be used later to requeue according to the schedule
+	_, instanceDeleted, err := r.updateTargetResourcePreviouslyMissing(ctx, instance)
+	if err != nil {
+		// Error occurred during status update or deletion, requeue
+		r.log.Errorw("failed to handle target resource status", "DisruptionSchedule", instance.Name, "err", err)
```
ptnapoleon, line 68:
```suggestion
		r.log.Errorw("failed to handle target resource status", "err", err)
```

redundant with existing annotations

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if instanceDeleted {
```
ptnapoleon, line 72: At least as of right now, this `if` check and block of code do nothing, and can be removed

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		instance.Status.LastScheduleTime = &metav1.Time{Time: *mostRecentScheduleTime}
+	} else {
+		instance.Status.LastScheduleTime = nil
+	}
+
+	return r.Client.Status().Update(ctx, instance)
+}
+
+// getMostRecentScheduleTime returns the pointer to the most recent scheduled time among a list of disruptions
+func (r *DisruptionScheduleReconciler) getMostRecentScheduleTime(disruptions *chaosv1beta1.DisruptionList) *time.Time {
```
ptnapoleon, line 105: This approach is very unlikely to work in most cases. We'll only be able to `Get` existing disruptions into this list, but disruptions do expire and get garbage collected, so we'll very frequently (almost always) just have an empty list here. If I understand the Status field we're trying to populate with this calculation, it won't be useful to constantly be setting it to 0/nil. We'll only be able to know the most recently success of a disruption if we have just created it and are following its progress

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	case "StatefulSet":
+		targetObj = &appsv1.StatefulSet{}
+	}
+
+	err := r.Client.Get(ctx, types.NamespacedName{
+		Name:      instance.Spec.TargetResource.Name,
+		Namespace: instance.Namespace,
+	}, targetObj)
+
+	return errors.IsNotFound(err), err
```
ptnapoleon, line 157: This is close to being just my personal preference, but I believe it will be easier for callers of this function if the return looks more like:
```
if errors.IsNotFound(err) {
  return false, nil
} else if err != nil {
  return true, nil
} else {
  return false, err
}
```

api/v1beta1/disruption_schedule_types.go
diyarab: Is there anything in particular that you see that can be marked as `omitempty`?

diyarab: Is there anything in particular that you see that can be marked as omitempty?

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (r *DisruptionScheduleReconciler) updateTargetResourcePreviouslyMissing(ctx context.Context, instance *chaosv1beta1.DisruptionSchedule) (bool, bool, error) {
+	disruptionScheduleDeleted := false
+	targetResourceNotFound, err := r.checkTargetResourceExists(ctx, instance)
+
+	if targetResourceNotFound {
+		r.log.Warnw("target does not exist, this schedule will be deleted if that continues", "error", err)
+
+		if instance.Status.TargetResourcePreviouslyMissing == nil {
+			r.log.Warnw("target is missing for the first time, updating status", "targetPreviouslyMissing", instance.Status.TargetResourcePreviouslyMissing)
```
ptnapoleon, line 168:
```suggestion
			r.log.Warn("target is missing for the first time, updating status")
```

Why add this annotation if `targetPreviouslyMissing` will always be `nil` here?

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+			return targetResourceNotFound, disruptionScheduleDeleted, r.handleTargetResourceFirstMissing(ctx, instance)
+		}
+
+		if time.Since(instance.Status.TargetResourcePreviouslyMissing.Time) > time.Hour*24 {
```
ptnapoleon, line 173: You should define `time.Hour * 24` as a Duration constant somewhere, and use that here. That will make it much easier when we need to refer to this value elsewhere, and help us make it configurable

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,221 @@
...
+func (r *DisruptionScheduleReconciler) handleTargetResourceFirstMissing(ctx context.Context, instance *chaosv1beta1.DisruptionSchedule) error {
+	instance.Status.TargetResourcePreviouslyMissing = &metav1.Time{Time: time.Now()}
+	if err := r.Client.Status().Update(ctx, instance); err != nil {
+		return fmt.Errorf("failed to update status: %w", err)
+	}
+
+	return nil
+}
+
+func (r *DisruptionScheduleReconciler) handleTargetResourceMissingOverOneDay(ctx context.Context, instance *chaosv1beta1.DisruptionSchedule) error {
```
ptnapoleon, line 199:
```suggestion
func (r *DisruptionScheduleReconciler) handleTargetResourceMissingOverExpiration(ctx context.Context, instance *chaosv1beta1.DisruptionSchedule) error {
```

No need to couple the function name to a value we'll likely make configurable soon enough

ptnapoleon: `TooLong`, `PastExpiration`, etc are all valid

api/v1beta1/disruption_schedule_types.go
diyarab: > I think you'll want to go through every suboption within this spec and consider whether they need to be `required`, and if not, if they should be `omitempty`. Required just means that the user has to explicitly specify them, it's fine to have "required" options that have sensible defaults when not specified.
> 
> `omitempty` is a json tag that comes after you specify the name, similar to metadata at Line 31, which means that when not specified, we don't include this in any json representations of this struct

Is there anything in particular that you see that can be marked as `omitempty`?

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,85 @@
...
+// +kubebuilder:object:root=true
+
+// DisruptionScheduleList contains a list of DisruptionSchedule
+type DisruptionScheduleList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []DisruptionSchedule `json:"items"`
+}
+
+type DisruptionScheduleSpec struct {
```
aymericDD, line 35: Is it possible to add a comment here?

controllers/disruption_controller.go
```diff
@@ -51,6 +51,16 @@ import (
 	"github.com/DataDog/chaos-controller/env"
 )
 
+// +kubebuilder:rbac:groups=chaos.datadoghq.com,resources=disruptions;disruptionschedules,verbs=list;watch;create;update;patch;delete
```
aymericDD, line 54: Why this section has been moved?

aymericDD (approved):
Good job! 💯  It could be nice to have a documentation about this new kind of disruption 👍

taihuynh167 (approved):

controllers/disruption_schedule_controller.go
diyarab: I think I might need to keep it. If we scheduled the disruption, but failed to update the LastScheduleTime on creation, we need to requeue so that the status could be updated. In this scenario at least one the disruption that we have scheduled will be available for us to update the status.

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,87 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              DisruptionScheduleSpec   `json:"spec,omitempty"`
+	Status            DisruptionScheduleStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// DisruptionScheduleList contains a list of DisruptionSchedule
+type DisruptionScheduleList struct {
```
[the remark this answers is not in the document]
diyarab: I thought it had to be defined for any new CRD?

aymericDD (approved):

api/v1beta1/disruption_schedule_types.go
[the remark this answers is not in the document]
diyarab: Based on some of the research I did. This `List` struct for the CRD is automatically created by Kubebuilder. We won't need to directly work with the `List` struct. It's mostly used by the client-go library under the hood to interact with the Kubernetes API server. It is especially useful for list/watch patterns in controllers. If we remove it, it might break these patterns.

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,87 @@
...
+	// +kubebuilder:validation:Required
+	// +ddmark:validation:Required=true
+	// The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+	Schedule string `json:"schedule"`
+
+	//+kubebuilder:validation:Minimum=0
+	// Optional deadline in seconds for starting the disruption if it misses scheduled
+	// time for any reason.  Missed disruption executions will be counted as failed ones.
+	// +nullable
+	StartingDeadlineSeconds *int64 `json:"startingDeadlineSeconds,omitempty"`
```
ptnapoleon, line 47: We can allow for a literal duration, using the DisruptionDuration type, so users can specify in a more human readable string, such as "1m", "300s", "30m", rather than having to convert to seconds. We also should come up with a non-zero default, and thus don't need this to be a pointer

api/v1beta1/disruption_schedule_types.go
```diff
@@ -0,0 +1,87 @@
...
+	TargetResource TargetResourceSpec `json:"targetResource"`
+
+	// +kubebuilder:validation:Required
+	// +ddmark:validation:Required=true
+	// Specifies the Disruption that will be created when executing a DisruptionShedule.
+	DisruptionTemplate DisruptionSpec `json:"disruptionTemplate"`
+}
+
+// TargetResource specifies the long-lived resource to be targeted for disruptions.
+// Disruptions are intended to exist semi-permanently, and thus appropriate targets can only be other long-lived resources,
```
ptnapoleon, line 62:
```suggestion
// DisruptionSchedules are intended to exist semi-permanently, and thus appropriate targets can only be other long-lived resources,
```

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,428 @@
...
+	if err != nil {
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.updateLastScheduleTime(ctx, instance, disruptions); err != nil {
+		r.log.Errorw("unable to update LastScheduleTime of DisruptionSchedule status", "err", err)
+		return ctrl.Result{}, err
+	}
+
+	// Get the next scheduled run, or the time of the unproccessed run
```
ptnapoleon, line 78:
```suggestion
	// Get the next scheduled run, or the time of the unprocessed run
```

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,428 @@
...
+
+	if len(disruptions.Items) > 0 {
+		r.log.Infow(fmt.Sprintf("cannot start a new disruption as a prior one is still running, scheduling next check in %s", requeueTime), "numActiveDisruptions", len(disruptions.Items))
+		return scheduledResult, nil
+	}
+
+	r.log.Infow("processing current run", "currentRun", missedRun.Format(time.UnixDate))
+
+	tooLate := false
+	if instance.Spec.StartingDeadlineSeconds != nil {
```
ptnapoleon, line 112: Coupling this with the comment I left above, but I don't think we should allow this to be unset

controllers/disruption_schedule_controller.go
```diff
@@ -0,0 +1,428 @@
...
+	disruptionCreated, err := r.createDisruption(ctx, instance, missedRun)
+	if err != nil {
+		r.log.Warnw(fmt.Sprintf("failed to create a disruption, scheduling next check in %s", requeueTime), "err", err)
+		return scheduledResult, err
+	}
+
+	if disruptionCreated {
+		r.log.Infow("created Disruption")
+
+		instance.Status.LastScheduleTime = &metav1.Time{Time: missedRun}
```
ptnapoleon, line 130: Why are we setting this to missedRun, instead of the disruption we just created? (there will be a gap in those times, no?)

ptnapoleon (approved):
Looks like a good starting place to merge this first PR. Other things can come in later PRs, we don't want to let this feature branch drift too long

PR #751, opened 2023-08-04
CHAOS-654: Prevent ability to update/delete userInfo

aymericDD:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- Deny the user to update or delete the UserInfo annotation in order to keep the integrity of the creator of the disruption.

Modify the disruption webhook to prevent the update of the UserInfo annotation during an update.

## Code Quality Checklist

## Testing

    - [x] locally.


api/v1beta1/disruption_webhook.go
```diff
@@ -296,6 +300,24 @@ You first need to remove those chaos pods (and potentially their finalizers) to
 	return nil
 }
 
+func (r *Disruption) validateUserInfo(oldDisruption *Disruption) error {
+	userInfo, err := r.UserInfo()
+	if err != nil {
+		return err
+	}
+
+	oldUserInfo, err := oldDisruption.UserInfo()
```
ptnapoleon, line 309: Could we allow overwriting user info if it was previously empty or invalid?

ptnapoleon (approved):
Left a suggestion, if possible

api/v1beta1/disruption_webhook.go
aymericDD: Fair point! 👍 Do we allow the creation of a disruption if the user info is empty or invalid? If it is not the case, we can consider the userInfo is valid by default. Hee we avoid the edit or delete of the userInfo,

aymericDD: After having a look at the user info webhook, I think it is possible to create a disruption with an empty or an invalid user info 🤔 So I will cover this case to allow a user to override the user info if it is empty or invalid.

ptnapoleon: I am fairly sure it's possible to get an empty one, i've seen it when we fail to set user info due to service accounts not looking how we expect

ptnapoleon (dismissed):

ptnapoleon (approved):

taihuynh167 (approved):

PR #758, opened 2023-08-10
Closes #752: InjectionStatus should default to NotInjected. Notify users when we can't find a target container.

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Contains two changes due to reported issue in #752. 
- We default injectionStatus to NotInjected. This way if we fail to have start any injection, users can still see an injection status.
- We remove the type DisruptionError, add a recorded event on disruptions when we can't find a target container, and change an error message that labels a container name as "target", to now label it as "targetContainer" for clarity. We had special cased logging code to print all `Context()` that came from a DisruptionError. But checking the current behavior, and the git history, this seems to have never worked. Thus, in the sole case we used DisruptionError, I've made sure the error message contains the desired context, and removed the type and all supporting logging code for it.

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


controllers/disruption_controller.go
```diff
@@ -38,7 +38,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
ptnapoleon, line 41: This change came about while testing, happy to undo it, but I think it helps make it clear in the code that we aren't importing the std lib's `errors` package

controllers/disruption_controller.go
```diff
@@ -555,10 +538,9 @@ func (r *DisruptionReconciler) createChaosPods(instance *chaosv1beta1.Disruption
 		// get IDs of targeted containers or all containers
 		targetContainers, err = chaosv1beta1.TargetedContainers(pod, instance.Spec.Containers)
 		if err != nil {
-			dErr := chaostypes.DisruptionError{Err: fmt.Errorf("error getting target pod container ID: %w", err)}
-			dErr.AddContext("targetPodStatus", pod.Status.String())
-			dErr.AddContext("targetPodName", target)
-			dErr.AddContext("targetPodNamespace", instance.Namespace)
+			dErr := fmt.Errorf("error getting target pod's container ID: %w", err)
```
ptnapoleon, line 541: This `err` already contains the target pod's name and container name. The namespace is already contained in the disruption logs. Additionally, the `Status.String()` was not printing something as simple as "Running" or "Completed", but an enormous struct, and thus not useful

aymericDD (approved):
Well done 👍

PR #760, opened 2023-08-14
Implement DisruptionRollout CRD

diyarab:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

This PR introduces the `DisruptionRollout` CRD, designed to seamlessly integrate the Chaos Controller into the software delivery lifecycle. By monitoring Deployment and StatefulSet updates, we can now trigger disruptions to validate the resilience of the applications.

Key Highlights:
- Introduced the `DisruptionRollout` CRD definition
- Established informers for `Deployments` and `StatefulSets`
- Initialized an indexer for efficient querying based on the `targetResource` field
- Deployed an event handler for `Deployment` and `StatefulSets` resources
- Implemented `DisruptionRollout` controller logic

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


watchers/deployment_handler.go
aymericDD, line 1: Can you add some unit tests? 🙏 
You can referer to the [unit tests](https://github.com/DataDog/chaos-controller/blob/main/watchers/chaos_pod_test.go#L44) of the **[ChaosPodHandler](https://github.com/DataDog/chaos-controller/blob/main/watchers/chaos_pod.go#L18)**.

watchers/statefulset_handler.go
aymericDD, line 1: What is the purpose of this watcher? It looks to do nothing 🤔

aymericDD (commented):
Great job 👏 If you can add tests for the new handlers (DeploymentHandler and StatefulSetHandler) it will be perfect 👌

watchers/statefulset_handler.go
diyarab: The PR is work in progress. I opened it to get the initial feedback. I will implement all the logic asap 😄

aymericDD: Alright, no problem 👍  
You can create or update a PR to draft: 
image

watchers/deployment_handler.go
diyarab: I will open an new PR for the unit tests since this one got pretty big 😄

controllers/disruption_rollout_controller.go
```diff
@@ -0,0 +1,247 @@
...
+		return scheduledResult, nil
+	}
+
+	if !r.targetResourceUpdated(&instance.Status) {
+		r.log.Infow("target resource hasn't been modified yet, sleeping")
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.LastContainerChangeTime.Before(instance.Status.LastScheduleTime) || instance.Status.LastContainerChangeTime.Equal(instance.Status.LastScheduleTime) {
+		r.log.Infow("target resource update has already been tested, sleeping")
```
ptnapoleon, line 92: Could we include LastContainerChangeTime and LastScheduleTime as tags in this log?

controllers/disruption_rollout_controller.go
```diff
@@ -0,0 +1,247 @@
...
+		return scheduledResult, nil
+	}
+
+	tooLate := false
+	if instance.Spec.DelayedStartTolerance.Duration() > 0 && !instance.Status.LastContainerChangeTime.IsZero() {
+		tooLate = instance.Status.LastContainerChangeTime.Add(instance.Spec.DelayedStartTolerance.Duration()).Before(time.Now())
+	}
+
+	if tooLate {
+		r.log.Infow("missed schedule to start a disruption, sleeping")
```
ptnapoleon, line 107: Could we include LastContainerChangeTime and DelayedStartTolerance as tags in this log?

controllers/disruption_rollout_controller.go
```diff
@@ -0,0 +1,247 @@
...
+	instance.Status.TargetResourcePreviouslyMissing = nil
+	if err := r.Client.Status().Update(ctx, instance); err != nil {
+		return fmt.Errorf("failed to update status: %w", err)
+	}
+
+	return nil
+}
+
+// targetResourceUpdated checks whether the target resource has been updated or not.
+func (r *DisruptionRolloutReconciler) targetResourceUpdated(status *chaosv1beta1.DisruptionRolloutStatus) bool {
```
ptnapoleon, line 228: This seems to be a stub? We don't check if the container hashes have changed

watchers/statefulset_handler.go
```diff
@@ -0,0 +1,137 @@
...
+// OnDelete is a handler function for the delete of a statefulset
+func (h StatefulSetHandler) OnDelete(_ interface{}) {
+	// Do nothing on delete event
+}
+
+func (h StatefulSetHandler) FetchAssociatedDisruptionRollouts(statefulset *appsv1.StatefulSet) (*chaosv1beta1.DisruptionRolloutList, error) {
+	indexedValue := "statefulset" + "-" + statefulset.Namespace + "-" + statefulset.Name
+
+	disruptionRollouts := &chaosv1beta1.DisruptionRolloutList{}
+	err := h.Client.List(context.Background(), disruptionRollouts, client.MatchingFields{"targetResource": indexedValue})
```
ptnapoleon, line 98: This field selector is checking the spec? Let's leave a TODO here that it would be more efficient to use label selectors, though this would require a webhook to add those labels when new rollouts are created

watchers/statefulset_handler.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, dr := range disruptionRollouts.Items {
+		dr.Status.LatestInitContainersHash = initContainersHash
+		dr.Status.LatestContainersHash = containersHash
+		dr.Status.LastContainerChangeTime = &metav1.Time{Time: time.Now()}
+
+		err = h.Client.Status().Update(context.TODO(), &dr)
```
ptnapoleon, line 129: Why choose context.TODO over context.Background?

watchers/deployment_handler.go
```diff
@@ -0,0 +1,231 @@
...
+		if err != nil {
+			h.log.Errorw("unable to update DisruptionRollout status", "DisruptionRollout", dr.Name, "error", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func ContainersChanged(oldPodSpec, newPodSpec *corev1.PodSpec, log *zap.SugaredLogger) (bool, map[string]string, map[string]string, error) {
```
ptnapoleon, line 144: Let's move these functions used by both deployment_handler and statefulset_handler, out of deployment_handler.go

controllers/disruption_rollout_controller.go
diyarab: This check primary purpose is to ensure that the instance.Status.LastContainerChangeTime was set, so that when we check if "target resource update has already been tested" we don't try to compare a value that has not been set.

main.go
```diff
@@ -236,6 +275,31 @@ func main() {
...
 	}
 
+	// create disruption rollout reconciler
+	disruptionRolloutReconciler := &controllers.DisruptionRolloutReconciler{
+		Client:  mgr.GetClient(),
+		BaseLog: logger,
+		Scheme:  mgr.GetScheme(),
+	}
+
+	if err := disruptionRolloutReconciler.SetupWithManager(mgr); err != nil {
```
ptnapoleon, line 285: It would be really great if we could gate turning on the rollout (and cron!) controllers based on a config value. That way we can safely release the controller even as we continue to work on these

controllers/disruption_rollout_controller.go
```diff
@@ -0,0 +1,253 @@
...
+		r.log.Infow(fmt.Sprintf("target resource is missing, scheduling next check in %s", requeueTime))
+		return scheduledResult, nil
+	}
+
+	if !r.targetResourceUpdated(&instance.Status) {
+		r.log.Infow("target resource hasn't been modified yet, sleeping")
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.LastContainerChangeTime.Before(instance.Status.LastScheduleTime) || instance.Status.LastContainerChangeTime.Equal(instance.Status.LastScheduleTime) {
```
ptnapoleon, line 91: Okay, so the watcher is what diffs the hashes, and this is the check we use to ensure we aren't re-testing the same rollout? In that case, we should change the log at L92 to Debug, instead of Info, as we'll be printing it many many times per Rollout

controllers/disruption_rollout_controller.go
diyarab: Correct

controllers/disruption_rollout_controller.go
diyarab: Yes, the watcher does hashes comparison and update. The controller is only responsible for scheduling disruptions (if all the conditions are met). This check is ensures that we don't re-test the same thing over and over. 

I have updated the log. Thank you for catching it.

chart/templates/configmap.yaml
```diff
@@ -4,7 +4,7 @@
 # Copyright 2023 Datadog, Inc.
 apiVersion: v1
 kind: ConfigMap
-metadata:
+metadata:
```
ptnapoleon, line 7:
```suggestion
metadata:
```

ptnapoleon (approved):

ptnapoleon (approved):

PR #778, opened 2023-09-15
CHAOS-770: add Probability field to disk failure

aymericDD:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- The Probability field allows the user to define a percentage of failures. For example, a Probability field set to 80% will disturb 80% of syscalls.

> Example

- Disturbe `100%` of syscalls

``` yaml
---
kind: Disruption
metadata:
  name: disk-failure
  namespace: chaos-demo
  annotations:
    chaos.datadoghq.com/environment: "lima"
spec:
  level: pod
  selector:
    app: demo-curl
  count: 1
  diskFailure:
    paths: 
      - /mnt/data/disk-read
      - /mnt/data/disk-write
    openat:
      exitCode: ENOENT
    probability: 100% # New field
```

- Disturbe `50%` of syscalls

``` yaml
---
kind: Disruption
metadata:
  name: disk-failure
  namespace: chaos-demo
  annotations:
    chaos.datadoghq.com/environment: "lima"
spec:
  level: pod
  selector:
    app: demo-curl
  count: 1
  diskFailure:
    paths: 
      - /mnt/data/disk-read
      - /mnt/data/disk-write
    openat:
      exitCode: ENOENT
    probability: 50% # New field
```

## Code Quality Checklist

## Testing

    - [x] locally


api/v1beta1/disk_failure.go
```diff
@@ -25,28 +28,64 @@ type DiskFailureSpec struct {
...
 	}
 
 	return nil
 }
 
+func (s *DiskFailureSpec) validateProbability() error {
+	probabilityError := fmt.Errorf("the probability of the disk failure disruption should be a percentage within the range of 1%% to 100%%")
+
+	if !strings.HasSuffix(s.Probability, "%") {
+		return probabilityError
```
luphaz, line 72: As we are testing explicitly for all the error paths, should we have unique error instead to be explicit about why we are rejecting the disruption?
```suggestion
		return fmt.Errorf("the probability of the disk failure disruption should be a percentage with a %% suffix")
```

api/v1beta1/disk_failure.go
```diff
@@ -25,28 +28,64 @@ type DiskFailureSpec struct {
...
+	}
+
+	probabilityStr := strings.TrimSuffix(s.Probability, "%")
+	probabilityInt, err := strconv.Atoi(probabilityStr)
+
+	if err != nil {
+		return probabilityError
```
luphaz, line 79: It relates to the questioning above, maybe we should not hide to the user it's inputs (or at least what we have as it's input)

It could help understand what does not respect our expectations (or any bug sneaking in the middle if any at some point)
```suggestion
		return fmt.Errorf("Probability percentage can't be converted to int: %w", err)
```

ebpf/disk-failure/injection.bpf.c
```diff
@@ -10,6 +10,10 @@ const volatile pid_t target_pid = 0;
 const volatile pid_t exclude_pid;
 const volatile char filter_path[61];
 const volatile pid_t exit_code = ENOENT;
+const volatile int probability = 100;
+
+unsigned int hits = 0;
+unsigned  int disruptedHits = 0;
```
luphaz, line 16:
```suggestion
unsigned int disruptedHits = 0;
```

api/v1beta1/disk_failure.go
```diff
@@ -25,28 +28,64 @@ type DiskFailureSpec struct {
 	Paths []string `json:"paths"`
 	// +nullable
 	OpenatSyscall *OpenatSyscallSpec `json:"openat,omitempty"`
+	// +kubebuilder:validation:Required
+	// +ddmark:validation:Required=true
```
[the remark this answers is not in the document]
aymericDD: Yes, I want to do this at the start, but what if the user defines the field but not set any value? Should we consider adding the default value or force the user to define a value?

api/v1beta1/disk_failure.go
aymericDD: 💯 agree with you!  I prefer to be explicit rather implicit 👍

luphaz (dismissed):

luphaz (approved):

luphaz (approved):

PR #764, opened 2023-08-21
CHAOS-232: Moving chaos pod logic from the chaos controller to a dedicated service

aymericDD:
## What does this PR do?

- [ ] Adds new functionality
- [x] Alters existing functionality
- [x] Fixes a bug
- [x] Improves documentation or testing

Please briefly describe your changes as well as the motivation behind them:

- Moving the chaos pod logic from the disruption controller inside a dedicated service named **ChaosPodService**:
  * **Pros**
    * Reducing by half the size (25 > 13) of the Disruption Reconcile parameters.
    * Coverage the chaos pods with unit tests.
- Removing helpers functions.
- Update the `cloudservice` package by adding an interface to be able to mock the `CloudServicesProvidersManager`. Implies unit test update and implementation update too. 
- Propagate the context from the controller to child calls.

## Code Quality Checklist

- [x] The documentation is up to date.
- [x] My code is sufficiently commented and passes continuous integration checks.
- [x] I have signed my commit (see [Contributing Docs](../CONTRIBUTING.md)).

## Testing

- [ ] I leveraged continuous integration testing
    - [ ] by depending on existing `unit` tests or `end-to-end` tests.
    - [ ] by adding new `unit` tests or `end-to-end` tests.
- [x] I manually tested the following steps:
    - `x`
    - [x] locally.
    - [ ] as a canary deployment to a cluster.

PR #781, opened 2023-09-21
Running Chaos Pods should not be evaluated as StuckOnRemoval

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- After #764, any disruption that was deleted manually, would consider itself as StuckOnRemoval. We'd issue a delete to the chaos pod, then run handleChaosPodTermination, and immediately evaluate it as Stuck. At first, I couldn't identify the change in #764 that caused this, as we did the same back-to-back procedure there. _however_, I did ultimately identify a small change.
https://github.com/DataDog/chaos-controller/blob/c618197d023c5f88441c9bb0fe58d6def7e49c71/controllers/disruption_controller.go#L829C10-L834
The default case of this switch statement was no longer considered after the refactor. Any chaos pods that weren't yet in a terminated state, were being considered as Stuck, as we assumed their finalizers couldn't be removed. I've in-lined the function `isFinalizerNotRemovableForChaosPod`, as its `bool` return value was making it hard to handle the fact that we needed three different exit cases: (1) pod is terminated and we can remove the finalizer, (2) pod is terminated  and we cannot remove the finalizer, i.e., StuckOnRemoval, and (3) pod is still cleaning and we should come back to it next time we reconcile.

The relevant test cases have been updated, especially when we expect stuckOnRemoval to be true:
https://github.com/DataDog/chaos-controller/pull/781/files#diff-7e0579210295a9153cd7c4bde60932eca36fffac5d47fc6b8ad1646600b17ae9R406-R415
## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.

### DIFF
--- a/services/chaospod.go
+++ b/services/chaospod.go
@@ -182,6 +182,8 @@ func (m *chaosPodService) HandleChaosPodTermination(ctx context.Context, disrupt
 // DeletePod attempts to delete the specified pod from the Kubernetes cluster.
 // Returns true if deletion was successful, otherwise returns false.
 func (m *chaosPodService) DeletePod(ctx context.Context, pod corev1.Pod) bool {
+	m.config.Log.Infow("terminating chaos pod to trigger cleanup", "chaosPod", pod.Name)
+
 	if err := m.deletePod(ctx, pod); err != nil {
 		m.config.Log.Errorw("Error terminating chaos pod", "error", err, "chaosPod", pod.Name)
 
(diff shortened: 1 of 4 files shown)


services/chaospod.go
```diff
@@ -590,41 +622,6 @@ func (m *chaosPodService) removeFinalizerForChaosPod(ctx context.Context, chaosP
...
-
-			if cs.State.Terminated != nil && cs.State.Terminated.Reason == "StartError" {
-				return false
-			}
-
-			break
-		}
-	}
-
-	return true
```
aymericDD, line 625: This line has not the same behavior as if we use the default of the switch case?

aymericDD (approved):
LGTM! My apologies for this regression... 🙈

services/chaospod.go
ptnapoleon: No :(. What happens is this returns "true", telling the caller that the finalizer _cannot_ be removed. The caller then assumes that the chaos pod is stuck on removal. We need a way to distinguish the "not yet" removal, from the "not ever" removal, and rather than complicate the return value of this function, I just in-lined it.

PR #783, opened 2023-09-27
CHAOS-785: Add new config for node level allowed hosts

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Technically we considered changing the current allowedHosts field to only apply during node level disruptions, but this is fairly breaking change so I wasn't sure
- The naming of the new field is horrible, please advise
- Essentially, some users want to disrupt connection to the core components that we currently have on our allowlist. As long as they're doing this with pod level disruptions against their own apps, that should be totally fine. So we should restrict the allowlist to only node level disruptions

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


chart/templates/configmap.yaml
```diff
@@ -88,6 +88,11 @@ data:
           {{ $v := printf "%s;%v;%s;%s" ($allowedHost.host | default "") ($allowedHost.port | default "") ($allowedHost.protocol | default "") ($allowedHost.flow | default "") -}}
           - {{ tpl $v $ }}
           {{- end }}
+        nodeLevelAllowedHosts:
+          {{- range $index, $allowedHost := .Values.injector.networkDisruption.allowedHosts }}
```
luphaz, line 92:
```suggestion
          {{- range $index, $allowedHost := .Values.injector.networkDisruption.nodeLevelAllowedHosts }}
```

luphaz (dismissed):
It seems fair to introduce a new configuration instead of breaking the previous behaviour.

The new name is clear for me at least 😉

aymericDD (approved):

Devatoria:
I'm a bit confused about having one field for node level only and one field for both node and pod levels, they are overlapping. What do you think about having both `podLevelAllowedHosts` and `nodeLevelAllowedHosts` (or `allowedHostsPod/NodeLevel` if we prefer to suffix instead) as new fields? We can keep the general `allowedHosts` for now in the CRD and use the admission webhook to fill both new fields (`pod/nodeLevelAllowedHosts`) with `allowedHosts` values when specified for retro compatibility. We would deprecate that field until future release.

It means we would have a configuration like:
```
networkDisruptions:
    podLevelAllowedHosts: [] // pod level only
    nodeLevelAllowedHosts: [] // node level only
    allowedHosts: [] // DEPRECATED: both pod and node level
```

Devatoria (changes requested):
Blocking until we discussed the comment I left on the PR to avoid too fast merge

ptnapoleon:
Hmm. Well, we don't have any use cases for podLevelAllowedHosts, and we don't even have any use cases for allowedHosts anymore... Do we just want to deprecate the old field and _only_ have nodeLevelAllowedHosts?

ptnapoleon:
Rather than break our config and make it more complicated to use, we should just inform users to use `disableDefaultAllowedHosts` in their spec when necessary

PR #786, opened 2023-10-05
CHAOS-791: Clean should be idempotent, dont error when deleting a missing iptables chain

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- So far I'm unable to reproduce the original problem locally, so I'll need to test this in staging.
- Makes one serious interface change to Injector so that when we log an error about an injection failing, we include which container it failed for. It was hard for me to follow exactly what errored in the logs when targeting a five container pod given the final error doesn't specify that info.
- Make iptables.Clear more idempotent, by skipping the deletion of any iptables rule if the chain doesn't exist. we were already checking if the rule existed, but sometimes we'd get an error that the chain didn't exist. I'm not sure if the underlying behavior of the iptables dependency changed, but it's simple enough to check the chain and then check the rule

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/cli/injector/main.go
+++ b/cli/injector/main.go
@@ -354,7 +354,7 @@ func inject(kind string, sendToMetrics bool, reinjection bool) bool {
 				}
 			}
 
-			log.Errorw("disruption injection failed", "error", err)
+			log.Errorw("disruption injection failed", "error", err, "target", inj.TargetName())
 		} else {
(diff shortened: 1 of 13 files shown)


aymericDD (approved):
LGTM 👍

PR #792, opened 2023-10-13
CHAOS-797: use correct flags when checking safemode config

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- So I've learned the keys we pass to viper are entirely cosmetic (for how we use them, at least)? It's the `mainFS.Lookup` that figures out where in the configmap to look, based on the json tags we put for the config structs. You can put literally anything in these keys (I tried it), and we will still read all the config values correctly. Nonetheless, these should be fixed for consistency

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -306,28 +306,28 @@ func New(logger *zap.SugaredLogger, osArgs []string) (config, error) {
 
 	mainFS.StringVar(&cfg.Controller.SafeMode.Environment, "safemode-environment", "", "Specify the 'location' this controller is run in. All disruptions must have an annotation of chaos.datadoghq.com/environment configured with this location to be allowed to create")
 
-	if err := viper.BindPFlag("conroller.safemode.environment", mainFS.Lookup("safemode-environment")); err != nil {
+	if err := viper.BindPFlag("controller.safeMode.environment", mainFS.Lookup("safemode-environment")); err != nil {
 		return cfg, err
 	}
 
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -306,28 +306,28 @@ func New(logger *zap.SugaredLogger, osArgs []string) (config, error) {
 
 	mainFS.StringVar(&cfg.Controller.SafeMode.Environment, "safemode-environment", "", "Specify the 'location' this controller is run in. All disruptions must have an annotation of chaos.datadoghq.com/environment configured with this location to be allowed to create")
 
-	if err := viper.BindPFlag("conroller.safemode.environment", mainFS.Lookup("safemode-environment")); err != nil {
+	if err := viper.BindPFlag("controller.safeMode.environment", mainFS.Lookup("safemode-environment")); err != nil {
```
taihuynh167, line 309: tbh this type of change scares me 😅

taihuynh167 (approved):

config/config.go
Devatoria: does it really have an impact? I thought it would just lower case it internally to avoid such conflicts

ptnapoleon: It does not actually have any impact, but camelCase is what we use for the json tags and in the configmap, so we should be consistent here

PR #795, opened 2023-10-18
Dont allow users to create bandwidthLimit disruptions below 32 bytes

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- According to https://man7.org/linux/man-pages/man8/tc-tbf.8.html , there is a system-dependent minimum on the `tc tbf rate` parameter. The error when setting one too low will look like:
```
error applying tc operations: could not perform operation on newly created qdisc: encountered error (exit status 1) using args ([qdisc add dev eth0 parent 2:2 handle 3: tbf rate 1 latency 50ms burst 1]): tbf: the "rate" parameter is mandatory.
Usage: ... tbf limit BYTES burst BYTES[/BYTES] rate KBPS [ mtu BYTES[/BYTES] ]
	[ peakrate KBPS ] [ latency TIME ] [ overhead BYTES ] [ linklayer TYPE ]
```
From my testing, though 64 bytes is mentioned on the man page, `lima` can handle 32 bytes, and I've found other systems we use that can go as low as 16. Let's try controlling at 32 bytes for now and see if that's a problem for users

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/network_disruption.go
+++ b/api/v1beta1/network_disruption.go
@@ -346,6 +346,10 @@ func (s *NetworkDisruptionSpec) Validate() (retErr error) {
 		}
 	}
 
+	if s.BandwidthLimit > 0 && s.BandwidthLimit < 32 {
+		retErr = multierror.Append(retErr, fmt.Errorf("bandwidthLimits below 32 bytes are not supported"))
+	}
+
 	return multierror.Prefix(retErr, "Network:")
 }
 
--- a/api/v1beta1/network_disruption_test.go
+++ b/api/v1beta1/network_disruption_test.go
@@ -442,6 +442,22 @@ var _ = Describe("NetworkDisruptionSpec", func() {
 			})
 		})
 
+		Describe("test option limits", func() {
+			It("rejects bandwidthLimits below 32 bytes", func() {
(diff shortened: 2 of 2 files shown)


api/v1beta1/network_disruption.go
```diff
@@ -346,6 +346,10 @@ func (s *NetworkDisruptionSpec) Validate() (retErr error) {
 		}
 	}
 
+	if s.BandwidthLimit > 0 && s.BandwidthLimit < 32 {
```
aymericDD, line 349: nit: add the unit to the field name: `BandwidthBytesLimit`

aymericDD (commented):
Simple and efficient 💯 
Can you add some unit test to cover this use case?

api/v1beta1/network_disruption_test.go
```diff
@@ -442,6 +442,22 @@ var _ = Describe("NetworkDisruptionSpec", func() {
 			})
 		})
 
+		Describe("test option limits", func() {
+			It("rejects bandwidthLimits below 32 bytes", func() {
```
aymericDD, line 446: These tests are working, but In this kind of  scenario you can use the `DescribeTable` 😉

aymericDD (approved):

PR #801, opened 2023-10-25
CHAOSPLT-76: verify kernel requirements for eBPF injections

aymericDD:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- This feature helps users understand why the application of disruption using eBPF may not be possible on a targeted node.

> Key Verification Checks:

- **Kernel Config Files**: Verify the presence of kernel configuration files on the node.
- **Mandatory Kernel Features**: Verify if the node supports mandatory kernel features required for eBPF.
-  **Mandatory eBPF Maps**: Confirm whether the node supports the mandatory eBPF maps.

## Code Quality Checklist

## Testing

    - [x] locally.


injector/network_disruption.go
```diff
@@ -160,6 +169,16 @@ func (i *networkDisruptionInjector) TargetName() string {
 
 // Inject injects the given network disruption into the given container
 func (i *networkDisruptionInjector) Inject() error {
+	if i.spec.HasHTTPFilters() {
+		if err := i.config.BPFConfigInformer.ValidateRequiredSystemConfig(); err != nil {
+			return err
+		}
+
+		if !i.config.BPFConfigInformer.GetMapTypes().HaveArrayMapType {
+			return fmt.Errorf("the disk failure needs the array map type, but the kernel does not support this type of map")
```
ptnapoleon, line 178:
```suggestion
			return fmt.Errorf("the http network failure needs the array map type, but the kernel does not support this type of map")
```

ptnapoleon (dismissed):
While I have no knowledge that you're checking for the right kernel configs, I'll trust that all is good. Other than the one coped-but-not-edited error message that needs to be changed, this looks good.

My only feedback is that the error messages we return here aren't going to make it to the user at all. I know that's a much larger change, but could we open a ticket?

injector/network_disruption.go
aymericDD: Oups! 🙈

aymericDD:
Thanks for your review 🙏 
So the purpose of the new ticket you propose is to find a way to notify the user about this error?

ptnapoleon:
> So the purpose of the new ticket you propose is to find a way to notify the user about this error?

Yep! If we can get the info into the disruption's `describe`, or even directly message them via a notifier

ptnapoleon (approved):

PR #810, opened 2023-11-13
Remove environment annotation from examples

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Removes all of the safemode environment annotations from the examples
- Removes the requirement to use the safemode annotations in lima
- We've had many, many users who, quite naturally, build their disruptions by looking at examples, rather than reading any of the docs. They end up not knowing what this env annotation is for, and set it incorrectly. In environments where it is actually mandatory, the webhook rejection message explains how to use it, so we don't need to explicitly include it in the examples like this. It's only been doing more UX harm than good

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/examples/complete.yaml
+++ b/examples/complete.yaml
@@ -7,9 +7,7 @@ apiVersion: chaos.datadoghq.com/v1beta1
 kind: Disruption
 metadata:
   name: disruption-sample
-  namespace: chaos-demo # disruption resource must be in the same namespace as targeted pods
-  annotations:
-    chaos.datadoghq.com/environment: "lima"
+  namespace: chaos-demo # disruption resource must be in the same namespace as targeted pods\
 spec:
   dryRun: false # optional, enable dry-run mode (chaos pods will be created but won't inject anything)
   reporting: # optional, add custom notification for this disruption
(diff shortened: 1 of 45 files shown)


examples/complete.yaml
```diff
@@ -7,9 +7,7 @@ apiVersion: chaos.datadoghq.com/v1beta1
 kind: Disruption
 metadata:
   name: disruption-sample
-  namespace: chaos-demo # disruption resource must be in the same namespace as targeted pods
-  annotations:
-    chaos.datadoghq.com/environment: "lima"
+  namespace: chaos-demo # disruption resource must be in the same namespace as targeted pods\
```
clairecng, line 10:
```suggestion
  namespace: chaos-demo # disruption resource must be in the same namespace as targeted pods
```

clairecng (commented):
should we add just one example called "with_environment_annotation" (or something similar)?

clairecng (dismissed):

aymericDD (approved):

ptnapoleon:
> should we add just one example called "with_environment_annotation" (or something similar)?

I'll put the annotation back into `complete.yaml`, for completeness

clairecng (approved):

PR #815, opened 2023-12-11
upgrade: mockery v2.38.0

aymericDD:
Upgrade mockery to the [latest version](https://github.com/vektra/mockery/releases/tag/v2.38.0).

### DIFF
--- a/.local.mockery.yaml
+++ b/.local.mockery.yaml
@@ -1,5 +1,4 @@
 all: True
-case: underscore
 dir: "{{.InterfaceDir}}"
 inpackage: True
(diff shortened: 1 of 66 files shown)


.local.mockery.yaml
```diff
@@ -1,5 +1,4 @@
 all: True
-case: underscore
```
aymericDD, line 2: Apparently, this config field is unsupported with the latest version. After re-generate mocks, it seems the the underscore case is used by default.

Errors:
image

ptnapoleon (approved):

PR #816, opened 2023-12-11
[CHAOSPLT-147] Metrics for Cron and Rollout Controllers

Azoam:
## What does this PR do?

- [X] Adds new functionality
- [ ] Fixes a bug

1. Adds new Metrics specific to Rollout and Cron controllers
    - **MetricDisruptionScheduled**: When a new disruption is scheduled for a specific target
    - **MetricNextScheduledTime**: The time difference between now and the next scheduled disruption
    - **MetricMissingTargetFound**: When a Missing Target is "seen" by the controller once again
    - **MetricTargetMissing**: When a Target is identified as Missing (which may occur several times for the same Target)
    - **MetricTooLate**: When the controller misses its opportunity to schedule a new disruption and waits for the next opportunity.
2. Refactor MetricSink by adding a `prefix` attribute
    - Given the old sink implementation considered there to only be one controller, we now find ourselves with 3 and therefore we need a way to differentiate metrics between the 3 controllers. Using a `prefix` attribute solves this.
## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -94,7 +94,7 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	r.log = r.BaseLog.With("disruptionName", req.Name, "disruptionNamespace", req.Namespace)
 
 	// reconcile metrics
-	r.handleMetricSinkError(r.MetricsSink.MetricReconcile())
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"controller", r.MetricsSink.GetSinkName()}))
 
 	defer func(tsStart time.Time) {
 		tags := []string{}
--- a/controllers/disruption_cron_controller.go
+++ b/controllers/disruption_cron_controller.go
@@ -8,6 +8,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"github.com/DataDog/chaos-controller/o11y/metrics"
 	"time"
 
 	chaosv1beta1 "github.com/DataDog/chaos-controller/api/v1beta1"
@@ -20,17 +21,30 @@ import (
 )
 
 type DisruptionCronReconciler struct {
-	Client  client.Client
-	Scheme  *runtime.Scheme
-	BaseLog *zap.SugaredLogger
-	log     *zap.SugaredLogger
+	Client      client.Client
+	Scheme      *runtime.Scheme
+	BaseLog     *zap.SugaredLogger
+	log         *zap.SugaredLogger
+	MetricsSink metrics.Sink
 }
 
 func (r *DisruptionCronReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
 	r.log = r.BaseLog.With("disruptionCronNamespace", req.Namespace, "disruptionCronName", req.Name)
 	r.log.Info("Reconciling DisruptionCron")
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"controller", r.MetricsSink.GetSinkName()}))
+
 	instance := &chaosv1beta1.DisruptionCron{}
+
+	defer func(tsStart time.Time) {
+		tags := []string{}
+		if instance.Name != "" {
+			tags = append(tags, "cronName:"+instance.Name, "namespace:"+instance.Namespace)
+		}
+
+		r.handleMetricSinkError(r.MetricsSink.MetricReconcileDuration(time.Since(tsStart), tags))
+	}(time.Now())
 
 	// Fetch DisruptionCron instance
 	if err := r.Client.Get(ctx, req.NamespacedName, instance); err != nil {
@@ -108,6 +122,8 @@ func (r *DisruptionCronReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	if tooLate {
+		// TODO: Add MetricTooLate
+		r.handleMetricSinkError(r.MetricsSink.MetricTooLate([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}))
 		r.log.Infow(fmt.Sprintf("missed schedule to start a disruption at %s, scheduling next check in %s", missedRun, requeueTime))
 		return scheduledResult, nil
 	}
@@ -127,7 +143,8 @@ func (r *DisruptionCronReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		r.log.Warnw("unable to create Disruption for DisruptionCron", "disruption", disruption, "err", err)
 		return ctrl.Result{}, err
 	}
-
+	// TODO: Add MetricDisruptionScheduled
+	r.handleMetricSinkError(r.MetricsSink.MetricDisruptionScheduled([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name, "disruptionName:", disruption.Name}))
 	r.log.Infow("created Disruption for DisruptionCron run", "disruptionName", disruption.Name)
 
 	// ------------------------------------------------------------------ //
@@ -177,7 +194,8 @@ func (r *DisruptionCronReconciler) updateTargetResourcePreviouslyMissing(ctx con
 
 		if instance.Status.TargetResourcePreviouslyMissing == nil {
 			r.log.Warnw("target is missing for the first time, updating status")
-
+			// TODO: Add MetricTargetMissing
+			r.handleMetricSinkError(r.MetricsSink.MetricTargetMissing(0, []string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}))
 			return targetResourceExists, disruptionCronDeleted, r.handleTargetResourceFirstMissing(ctx, instance)
 		}
 
@@ -186,12 +204,14 @@ func (r *DisruptionCronReconciler) updateTargetResourcePreviouslyMissing(ctx con
 				"timeMissing", time.Since(instance.Status.TargetResourcePreviouslyMissing.Time))
 
 			disruptionCronDeleted = true
-
+			// TODO: Add MetricTargetMissing with a day as the time
+			r.handleMetricSinkError(r.MetricsSink.MetricTargetMissing(time.Since(instance.Status.TargetResourcePreviouslyMissing.Time), []string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}))
 			return targetResourceExists, disruptionCronDeleted, r.handleTargetResourceMissingPastExpiration(ctx, instance)
 		}
 	} else if instance.Status.TargetResourcePreviouslyMissing != nil {
 		r.log.Infow("target was previously missing, but now present. updating the status accordingly")
-
+		// TODO: Add MetricTargetFound
+		r.handleMetricSinkError(r.MetricsSink.MetricMissingTargetFound([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", "targetName:", instance.Spec.TargetResource.Name}))
 		return targetResourceExists, disruptionCronDeleted, r.handleTargetResourceNowPresent(ctx, instance)
 	}
 
@@ -270,7 +290,8 @@ func (r *DisruptionCronReconciler) getNextSchedule(instance *chaosv1beta1.Disrup
 			return time.Time{}, time.Time{}, fmt.Errorf("too many missed start times (> 100)")
 		}
 	}
-
+	// TODO: Add MetricNextScheduledTime
+	r.handleMetricSinkError(r.MetricsSink.MetricNextScheduledTime(time.Until(sched.Next(now)), []string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}))
 	return lastMissed, sched.Next(now), nil
 }
 
@@ -279,4 +300,11 @@ func (r *DisruptionCronReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&chaosv1beta1.DisruptionCron{}).
 		Complete(r)
+}
+
+// handleMetricSinkError logs the given metric sink error if it is not nil
+func (r *DisruptionCronReconciler) handleMetricSinkError(err error) {
+	if err != nil {
+		r.log.Errorw("error sending a metric", "error", err)
+	}
 }
--- a/controllers/disruption_rollout_controller.go
+++ b/controllers/disruption_rollout_controller.go
@@ -7,6 +7,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"github.com/DataDog/chaos-controller/o11y/metrics"
 	"math/rand"
 	"time"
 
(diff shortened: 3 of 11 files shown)


o11y/metrics/types/types.go
```diff
@@ -23,6 +23,12 @@ const (
 	// SinkAppController is the chaos controller
 	SinkAppController SinkApp = "chaos-controller"
 
+	// SinkAppRolloutController is the rollout controller
+	SinkAppRolloutController SinkApp = "chaos-rollout-controller"
+
+	// SinkAppCronController is the cront controller
```
clairecng, line 29:
```suggestion
	// SinkAppCronController is the cron controller
```

main.go
```diff
@@ -314,13 +308,17 @@ func main() {
...
+		// new metrics sink for cron controller
 		// create disruption cron reconciler
 		disruptionCronReconciler := &controllers.DisruptionCronReconciler{
-			Client:  mgr.GetClient(),
-			BaseLog: logger,
-			Scheme:  mgr.GetScheme(),
+			Client:      mgr.GetClient(),
+			BaseLog:     logger,
+			Scheme:      mgr.GetScheme(),
+			MetricsSink: initMetricsSink(cfg.Controller.MetricsSink, logger, metricstypes.SinkAppCronController),
```
clairecng, line 317: nit: the new metrics sink comment should probably go in this line

main.go
```diff
@@ -402,3 +400,24 @@ func main() {
...
+	if err != nil {
+		logger.Errorw("error while creating metric sink, switching to noop", "error", err)
+
+		metricsSink, _ = metrics.GetSink(logger, metricstypes.SinkDriverNoop, app)
+	}
+
+	return metricsSink
+}
+
+// handle metrics sink cleint close on exit
```
clairecng, line 416:
```suggestion
// handle metrics sink client close on exit
```

examples/disruption_crons/disruption_cron_network_drop.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   name: network-drop
   namespace: chaos-demo
 spec:
-  schedule: "*/15 * * * *"
```
clairecng, line 12: why that change?

controllers/disruption_rollout_controller.go
```diff
@@ -107,6 +122,7 @@ func (r *DisruptionRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	if tooLate {
+		r.handleMetricSinkError(r.MetricsSink.MetricTooLate([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}))
```
clairecng, line 125: noticed that we have `[]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}` multiple times. Could we make this a variable or a func to mutualize the code?

controllers/disruption_rollout_controller.go
```diff
@@ -32,6 +35,18 @@ func (r *DisruptionRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	instance := &chaosv1beta1.DisruptionRollout{}
 	randSource := rand.New(rand.NewSource(time.Now().UnixNano()))
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace}))
+
+	defer func(tsStart time.Time) {
```
clairecng, line 41: what does this func do?

controllers/disruption_controller.go
```diff
@@ -94,7 +94,7 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	r.log = r.BaseLog.With("disruptionName", req.Name, "disruptionNamespace", req.Namespace)
 
 	// reconcile metrics
-	r.handleMetricSinkError(r.MetricsSink.MetricReconcile())
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"controller", r.MetricsSink.GetSinkName()}))
```
clairecng, line 97: I don't think controller is the right tag to send here? Maybe I'm missing something?
Same comment for the disruption cron controller on the same call.

controllers/disruption_rollout_controller.go
```diff
@@ -32,6 +35,18 @@ func (r *DisruptionRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	instance := &chaosv1beta1.DisruptionRollout{}
 	randSource := rand.New(rand.NewSource(time.Now().UnixNano()))
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace}))
```
clairecng, line 39: I don't think those are the right tags to send here? Maybe I'm missing something / not understanding something?
Or maybe it's the other way and this is right and the ones on the cron controller and chaos controller are wrong? 😅

controllers/disruption_cron_controller.go
```diff
@@ -20,18 +22,31 @@ import (
...
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"controller", r.MetricsSink.GetSinkName()}))
+
 	instance := &chaosv1beta1.DisruptionCron{}
 
+	defer func(tsStart time.Time) {
+		tags := []string{}
+		if instance.Name != "" {
+			tags = append(tags, "cronName:"+instance.Name, "namespace:"+instance.Namespace)
```
clairecng, line 44: IMO we should use the same tags in both logs and metrics: disruptionCronsomethingsomething

controllers/disruption_cron_controller.go
```diff
@@ -108,7 +123,9 @@ func (r *DisruptionCronReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	if tooLate {
+		r.handleMetricSinkError(r.MetricsSink.MetricTooLate([]string{"rolloutName:" + instance.Name, "namespace:", instance.Namespace, "targetName:", instance.Spec.TargetResource.Name}))
```
clairecng, line 126: same comment than for the rollout controller, make this into a variable

controllers/disruption_rollout_controller.go
clairecng: also let's use disruptionRollout as a suffix.

clairecng:
Also a bit more context for 'MetricTooLate' when does this use case occurs? (I'm not very familiar with cron / rollout controllers 😅 )

examples/disruption_crons/disruption_cron_network_drop.yaml
Azoam: Ah I was using this for testing, going to revert to 15, thanks.

controllers/disruption_rollout_controller.go
Azoam: disruptionRollout instead of rolloutName?

controllers/disruption_rollout_controller.go
Azoam: This is the function for calculating reconcile duration, but I missed this, it needs to have tags associated with it to differentiate between the functions, thanks for bringing this to my attention.

controllers/disruption_rollout_controller.go
Azoam: No youre right, these are incorrect tags. Should be the name of the App, not the individual rollouts that go through the Reconcile loop.

controllers/disruption_cron_controller.go
Azoam: Yeah that consistency sounds like it would be easier to correlate metrics and logs. I'll make the change.

controllers/disruption_rollout_controller.go
clairecng: I meant disruptionRolloutSomethingSomething for all the rollout tags, so here disruptionRolloutName!

Azoam:
> Also a bit more context for 'MetricTooLate' when does this use case occurs? (I'm not very familiar with cron / rollout controllers 😅 )

It occurs when a scheduled Disruption misses its time slot to run. So lets say the next schedule disruption should occur after X UTC and before Y UTC, but because the last disruption took too long or the target was missing for some time, its now >Y UTC, therefore we missed the schedule and we should wait until the next schedule (which is not always right after Y UTC).

controllers/disruption_controller.go
ptnapoleon, line 97: If the controller is in the prefix, why also have it as a tag?

controllers/disruption_cron_controller.go
```diff
@@ -10,6 +10,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/DataDog/chaos-controller/o11y/metrics"
```
ptnapoleon, line 13: nit: I think you'll want to swap L14 and L15 here, so that all the "local" imports are in the middle group. we ought to figure out/fix our auto-importing

controllers/disruption_controller.go
Azoam: Yup, no need to have it in the tag, going to remove for all.

controllers/disruption_rollout_controller.go
```diff
@@ -32,11 +37,25 @@ func (r *DisruptionRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	instance := &chaosv1beta1.DisruptionRollout{}
 	randSource := rand.New(rand.NewSource(time.Now().UnixNano()))
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile([]string{"controller", r.MetricsSink.GetSinkName()}))
+
+	defer func(tsStart time.Time) {
+		tags := []string{}
+		if instance.Name != "" {
+			tags = append(tags, "rolloutName:"+instance.Name, "namespace:"+instance.Namespace)
```
ptnapoleon, line 46: i know it's wordier, but why not be consistent with the `name` tags for the other two controllers: `disruptionRolloutName`?

ptnapoleon: nvm, i see you do so below, this one must just have been forgotten to update

examples/disruption_crons/testing.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: chaos.datadoghq.com/v1beta1
+kind: DisruptionCron
+metadata:
+  name: network-drop-147
+  namespace: chaos-demo
+spec:
+  schedule: "*/15 * * * *"
+  targetResource:
+    kind: deployment
+    name: chaos-microservices-frontend
```
ptnapoleon, line 15: I think we only want to commit yamls that work with our example demo?

controllers/disruption_rollout_controller.go
Azoam: Yeah, I briefly forgot the prefix gives a ton of this context so there is no reason to differentiate in the tags.

Azoam: I will leave it for the logs considering those arent prefixed with any context.

examples/disruption_crons/testing.yaml
Azoam: now how did that get in there >:(

controllers/disruption_rollout_controller.go
Azoam: checkout the comment Philip made below, but I think we can just simplify to "name" because the prefix on the metric already gives us context on whether its rollout or cron.

ptnapoleon:
Metrics in the disruption_controller use `disruptionName`: https://github.com/DataDog/chaos-controller/blob/bf2054267bc6a38b83ba7c51b96b6c15f73a812c/controllers/disruption_controller.go#L218 i  think we should be consistent with all metrics and use `disruptionRolloutName` and `disruptionCronName` instead of `name`, wdyt?

controllers/disruption_rollout_controller.go
ptnapoleon: sorry, i did a bad job explaining what i meant, I should have written `*name`. See my latest comment, but I think we should be consistent with the disruption_controller's existing metric tags

controllers/disruption_rollout_controller.go
ptnapoleon: I agree with Claire, I did a bad job explaining myself

Azoam:
> Metrics in the disruption_controller use `disruptionName`:
> 
> https://github.com/DataDog/chaos-controller/blob/bf2054267bc6a38b83ba7c51b96b6c15f73a812c/controllers/disruption_controller.go#L218
> 
> i think we should be consistent with all metrics and use `disruptionRolloutName` and `disruptionCronName` instead of `name`, wdyt?

I think name may be more appropriate given that the name of controller just so happens to be the name of the CRD. Unlike the Chaos Controller which reconciles on Disruption CRDs, it would make sense that in the controller the metrics specifically mention disruption incase there is any confusion that a Chaos Controller may have a Chaos CRD like its Rollout and Cron counterparts. Given Rollout and Cron's 1:1 naming of Controller:CRD, it seems redundant to add the prefix to name.

controllers/disruption_rollout_controller.go
Azoam: EDIT: Back to what you suggested 😅

ptnapoleon (approved):

controllers/disruption_cron_controller.go
```diff
@@ -19,24 +21,41 @@ import (
...
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile())
+
 	instance := &chaosv1beta1.DisruptionCron{}
 
+	defer func(tsStart time.Time) {
+		tags := []string{}
+		if instance.Name != "" {
+			tags = append(tags, "disruptionCronName:"+instance.Name, "namespace:"+instance.Namespace)
```
clairecng, line 46:
```suggestion
			tags = append(tags, "disruptionCronName:"+instance.Name, "disruptionCronNamespace:"+instance.Namespace)
```

controllers/disruption_cron_controller.go
```diff
@@ -19,24 +21,41 @@ import (
...
+
+		r.handleMetricSinkError(r.MetricsSink.MetricReconcileDuration(time.Since(tsStart), tags))
+	}(time.Now())
+
 	// Fetch DisruptionCron instance
 	if err := r.Client.Get(ctx, req.NamespacedName, instance); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	DisruptionCronTags = []string{"disruptionCronName:" + instance.Name, "namespace:" + instance.Namespace, "targetName:" + instance.Spec.TargetResource.Name}
```
clairecng, line 57:
```suggestion
	DisruptionCronTags = []string{"disruptionCronName:" + instance.Name, "disruptionCronNamespace:" + instance.Namespace, "disruptionCronTargetName:" + instance.Spec.TargetResource.Name}
```

controllers/disruption_rollout_controller.go
```diff
@@ -32,11 +37,25 @@ func (r *DisruptionRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	instance := &chaosv1beta1.DisruptionRollout{}
 	randSource := rand.New(rand.NewSource(time.Now().UnixNano()))
 
+	// reconcile metrics
+	r.handleMetricSinkError(r.MetricsSink.MetricReconcile())
+
+	defer func(tsStart time.Time) {
+		tags := []string{}
+		if instance.Name != "" {
+			tags = append(tags, "disruptionRolloutName:"+instance.Name, "namespace:"+instance.Namespace)
```
clairecng, line 46:
```suggestion
			tags = append(tags, "disruptionRolloutName:"+instance.Name, "disruptionRolloutNamespace:"+instance.Namespace)
```

clairecng (dismissed):
I'd prefer if we use disruption{type of controller}{name of the resource} as a prefix to all metrics: `disruptionRolloutNamespace` for example, I added some suggestions
Otherwise LGTM.

ptnapoleon (approved):

PR #817, opened 2023-12-15
Tai.huynh/chaosplt 146/history

taihuynh167:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- `x`

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


controllers/cron_rollout_helpers.go
```diff
@@ -174,41 +175,36 @@ func CreateDisruptionFromTemplate(ctx context.Context, cl client.Client, scheme
 }
 
 // getScheduledTimeForDisruption returns the scheduled time for a particular disruption.
-func getScheduledTimeForDisruption(disruption *chaosv1beta1.Disruption) (*time.Time, error) {
+func getScheduledTimeForDisruption(log *zap.SugaredLogger, disruption *chaosv1beta1.Disruption) time.Time {
```
clairecng, line 178: why don't we return the error anymore?

api/v1beta1/disruption_cron_types.go
```diff
@@ -83,4 +83,14 @@ type DisruptionCronStatus struct {
...
 	// +nullable
 	TargetResourcePreviouslyMissing *metav1.Time `json:"targetResourcePreviouslyMissing,omitempty"`
+
+	History []DisruptionRun `json:"history,omitempty"`
+}
+
+const MaxHistoryLen = 5
+
+type DisruptionRun struct {
+	Name      string      `json:"name,omitempty"`
```
clairecng, line 93: can we change names / kinds between runs?

aymericDD (dismissed):
💯

controllers/cron_rollout_helpers.go
taihuynh167: this is a private method, and in the caller (Reconcile loop), we already checked for cron expression validity earlier. Hence imho, this method can now assume a valid cron description. Its no longer meaningful to check again, and just makes code longer unnecessarily.

https://github.com/DataDog/chaos-controller/blob/b28cfba90cb7a6fa7bfa9571143859fc8c4989a1/controllers/disruption_cron_controller.go#L249

api/v1beta1/disruption_cron_types.go
taihuynh167: i'm not 100% sure here. Afaiu k8s doesnt have any guarantee that object will not change, or forbid changing. I don't see other logic in our codebase regarding safeguarding these changes. 

is there a quick test i can do with `k edit` to verify this? Do you know how?

clairecng: We identify disruptions per name/namespace so to me it wouldn't make sense we'd be authorized to edit the name field, but i've never really worked with disruption crons before so I don't know if it'd be possible unfortunately...
I asked this since we store the name and kind in the disruption run here ; if they are not editable / they don't change throughout runs, I don't know if it's really useful to keep this info.

clairecng: Also I don't know how 😅 sorry

clairecng (dismissed):
Couple of questions about what we store but otherwise LGTM!

api/v1beta1/disruption_cron_types.go
aymericDD: I can say something wrong, but FWIK disruption are immutable _(to reduce the complexity I supposed)_. The only exception is for the `count` field which dans be updated.

clairecng: Disruptions are but I'm unsure about disruption crons 🤔

aymericDD: Good question. For cron disruption, I am not aware about how it really works.

ptnapoleon: > can we change names / kinds between runs?

At the very least, we need to support changing names between runs because otherwise, anytime we change the cron --> disruption naming logic, we'd break every existing cron

controllers/cron_rollout_helpers.go
```diff
@@ -174,41 +175,36 @@ func CreateDisruptionFromTemplate(ctx context.Context, cl client.Client, scheme
...
 	timeRaw := disruption.Annotations[ScheduledAtAnnotation]
 	if len(timeRaw) == 0 {
-		return nil, nil
+		return time.Time{}
 	}
 
 	timeParsed, err := time.Parse(time.RFC3339, timeRaw)
 	if err != nil {
-		return nil, err
+		log.Errorw("unable to parse schedule time for child disruption", "err", err, "disruption", disruption.Name)
```
ptnapoleon, line 186:
```suggestion
		log.Errorw("unable to parse schedule time for child disruption", "err", err, "disruptionName", disruption.Name)
```

For consistency with disruption.name tags elsewhere

clairecng: log does not have the right tags already associated? (disruptionCronName, disruptionCronNamespace)

api/v1beta1/disruption_cron_types.go
Devatoria: As far as I can understand, this struct is expected to represent a past run of a disruption stored in the status of the disruption cron resource. Changing the name or the spec of the disruption cron resource should not change what is stored in the status since they are past events anyway.

So I don't think there's anything else to be done here? Eventually renaming the struct if not explicit enough.

api/v1beta1/disruption_cron_types.go
```diff
@@ -83,4 +83,14 @@ type DisruptionCronStatus struct {
...
 	TargetResourcePreviouslyMissing *metav1.Time `json:"targetResourcePreviouslyMissing,omitempty"`
+
+	History []DisruptionCronTrigger `json:"history,omitempty"`
+}
+
+const MaxHistoryLen = 5
+
+type DisruptionCronTrigger struct {
+	Name      string      `json:"name,omitempty"`
+	Kind      string      `json:"kind,omitempty"`
```
ptnapoleon, line 94: This `kind` seems to always be the kind of the crd we're reconciling, disruptioncron. Why reiterate it here? This used to store the names of the last disruption runs, now it just stores the `lastScheduledTime`s for a given cron. What's the benefit of storing name/kind here?

ptnapoleon: https://github.com/DataDog/chaos-controller/pull/817/files#diff-67195dd0070673a7c0f061485aa4c43739add706be74ecb8fd11ff345ddaab5bR166-R168 seems different than the pr description

controllers/cron_rollout_helpers.go
```diff
@@ -174,41 +175,36 @@ func CreateDisruptionFromTemplate(ctx context.Context, cl client.Client, scheme
...
-	}
-
-	return mostRecentScheduleTime
+func GetMostRecentScheduleTime(log *zap.SugaredLogger, disruptions *chaosv1beta1.DisruptionList) time.Time {
+	length := len(disruptions.Items)
+	if length == 0 {
+		return time.Time{}
+	}
+
+	sort.Slice(disruptions.Items, func(i, j int) bool {
```
Azoam, line 200: The initial linear fashion of finding the most recent time should be faster and more efficient than sorting the items and picking out the last item.

controllers/cron_rollout_helpers.go
```diff
@@ -174,41 +175,36 @@ func CreateDisruptionFromTemplate(ctx context.Context, cl client.Client, scheme
...
-		return nil, err
+		log.Errorw("unable to parse schedule time for child disruption", "err", err, "disruptionName", disruption.Name)
+		return time.Time{}
 	}
 
-	return &timeParsed, nil
+	return timeParsed
 }
 
 // getMostRecentScheduleTime returns the most recent scheduled time from a list of disruptions.
```
Azoam, line 193: nit: GetMost... rather than getMost

controllers/disruption_cron_controller.go
```diff
@@ -161,6 +162,18 @@ func (r *DisruptionCronReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	// Add the start time of the just initiated disruption to the status
 	instance.Status.LastScheduleTime = &metav1.Time{Time: missedRun}
+	instance.Status.History = append(instance.Status.History, chaosv1beta1.DisruptionCronTrigger{
```
Azoam, line 165: nit: comment regarding history and dropping off the oldest histories.

api/v1beta1/disruption_cron_types.go
taihuynh167: what I did here is to take a snapshot of the `disruptionCron` and store in `history`. I'm unsure if `name` and `kind` of a `disruptionCron` changes are allowed in chaos-controller, but lets assume they are, then the `history` should still show the original `name, kind` at that point in time.

aymericDD (approved):
LGTM 👍

controllers/cron_rollout_helpers.go
taihuynh167: true. Its just a hunch that the previous code is buggy. 

I personally find the sorting easier to read. In practice, this list is always an almost-sorted list. I'd rather pay a small fee to ensure it's really sorted, to avoid wrong assumption during debugging.

api/v1beta1/disruption_cron_types.go
ptnapoleon: Understood, :thumbsup:

aymericDD (approved):

controllers/cron_rollout_helpers.go
Azoam: I'm not sure how it could be buggy, it seems straightforward. The sorting does seem cleaner, but you can also encapsulate a lot of the code in the previous code into smaller functions to also make it cleaner. The fee is dependent on the # of disruptions in that list, if there are 100 disruptions, then the sorting is ~6X slower, which may not be a small fee especially if it grows. Also, we dont need disruptions items to be sorted from my understanding besides in this function where you grab the last item from the sorted list, unless there is another place where having this list be sorted again makes sense.

Azoam (approved):

api/v1beta1/disruption_cron_types.go
ptnapoleon: i did look it up, and can confirm that name and kind can never be mutated

PR #827, opened 2024-01-18
chaosplt-233: check for no target containers found

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- When a target pod has non-init containers, but none have the status Running, we can incorrectly return an empty list of target containers. This causes the injectors to be invalid, which we should avoid

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_types.go
+++ b/api/v1beta1/disruption_types.go
@@ -855,6 +855,10 @@ func TargetedContainers(pod corev1.Pod, containerNames []string) (map[string]str
 			}
 		}
 
+		if len(allContainers) < 1 {
+			return nil, fmt.Errorf("couldn't find any running containers for pod '%s'", pod.Name)
+		}
+
 		return allContainers, nil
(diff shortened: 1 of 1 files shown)


aymericDD (approved):

ptnapoleon:
> Is this a race condition? Should we expect the injector to be able to reinject once containers are running? Not something we want to treat as part of this PR but worth asking the question if we want to handle such a case in the future.

Afaict, yes. Unless the target is stuck in a weird situation, we do expect to get to a reconcile loop where it has a normal status and we can find containers for it. If the target stays like this permanently, then we'd probably never finish injecting into other, legitimate targets. But given this is the first time I've ever seen this error, I think we can safely assume for now that it's almost always a temporary condition

ptnapoleon:
there have been no issues after multiple disruptions, merging

PR #829, opened 2024-01-23
CHAOSPLT-115: Warn users who set count to 100

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

-  Send the user a warning when they specify an integer count of `100`. Due to notifier aggregation, they will only receive this once

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -168,7 +168,7 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, fmt.Errorf("error handling chaos pods termination: %w", err)
 	}
 
-	if err := r.sendWarnings(instance); err != nil {
+	if err := r.sendWarningsOnDisruptionSpec(instance); err != nil {
 		r.log.Errorw("unable to send user warnings", "err", err)
 	}
 
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -891,6 +891,8 @@ func (r *DisruptionReconciler) sendWarnings(instance *chaosv1beta1.Disruption) e
 	value, isPercent, err := chaosv1beta1.GetIntOrPercentValueSafely(instance.Spec.Count)
 	if err == nil && !isPercent && value == 100 {
 		r.recordEventOnDisruption(instance, chaosv1beta1.EventInvalidSpecDisruption, "disruption count was set to the integer 100, but you likely intended it to be the string \"100%\"", "")
+	} else if err != nil {
+		return err
 	}
 
 	return nil
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -885,7 +885,7 @@ func (r *DisruptionReconciler) recordEventOnDisruption(instance *chaosv1beta1.Di
 	}
 }
 
-// sendWarnings is used to send users warnings about their disruption spec that we're unable to detect or warn about in the validation webhook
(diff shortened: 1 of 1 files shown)


controllers/disruption_controller.go
```diff
@@ -168,6 +168,10 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, fmt.Errorf("error handling chaos pods termination: %w", err)
 	}
 
+	if err := r.sendWarnings(instance); err != nil {
+		r.log.Errorw("unable to send user warnings", "err", err)
```
ptnapoleon, line 172: Ultimately this is optional, so I don't want to stop the reconcile loop if this fails

controllers/disruption_controller.go
```diff
@@ -881,6 +885,19 @@ func (r *DisruptionReconciler) recordEventOnDisruption(instance *chaosv1beta1.Di
 	}
 }
 
+// sendWarnings is used to send users warnings about their disruption spec that we're unable to detect or warn about in the validation webhook
+func (r *DisruptionReconciler) sendWarnings(instance *chaosv1beta1.Disruption) error {
```
clairecng, line 889:
```suggestion
func (r *DisruptionReconciler) sendWarningsOnDisruptionSpec(instance *chaosv1beta1.Disruption) error {
```

taihuynh167 (approved):

controllers/disruption_controller.go
ptnapoleon: I wanted to leave the more generic title in case we had other uses for sending warnings, but you're right, that's the wrong instinct, we can always rename it later

clairecng: it's fine if we do leave it generic but in that case let's also make the comment generic.

clairecng (approved):

controllers/disruption_controller.go
```diff
@@ -168,6 +168,10 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, fmt.Errorf("error handling chaos pods termination: %w", err)
 	}
 
+	if err := r.sendWarnings(instance); err != nil {
```
clairecng, line 171:
```suggestion
	if err := r.sendWarningsOnDisruptionSpec(instance); err != nil {
```

clairecng (changes requested):

controllers/disruption_controller.go
clairecng: approved too early haha :oops:

controllers/disruption_controller.go
```diff
@@ -881,6 +885,19 @@ func (r *DisruptionReconciler) recordEventOnDisruption(instance *chaosv1beta1.Di
 	}
 }
 
+// sendWarnings is used to send users warnings about their disruption spec that we're unable to detect or warn about in the validation webhook
```
clairecng, line 888:
```suggestion
// sendWarningsOnDisruptionSpec is used to send users warnings about their disruption spec that we're unable to detect or warn about in the validation webhook
```

controllers/disruption_controller.go
ptnapoleon: Honestly just silly that github can't manage to correct every use of the function when you suggested the name change. It knows where they all are!

clairecng (approved):

PR #831, opened 2024-01-31
CHAOSPLT-216: Allow for requiring specific user groups before creating a disruption

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Allow for specifying a list of user groups in the chart. ValidateCreate in the admission webhook will reject any disruption coming from a user who does not belong to at least one of these groups

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -352,6 +352,7 @@ func (r *Disruption) validateUserInfoGroup() error {
 	}
 
 	logger.Warnw("rejecting user from creating this disruption", "permittedUserGroups", permittedUserGroups, "userGroups", userInfo.Groups)
+
 	return fmt.Errorf("lacking sufficient authorization to create disruptions. your user groups are %s, but you must be in one of the following groups: %s", userInfo.Groups, permittedUserGroupWarningString)
 }
 
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -52,6 +52,7 @@ var (
 	chaosNamespace                string
 	ddmarkClient                  ddmark.Client
 	safemodeEnvironment           string
+	permittedUserGroups           map[string]struct{}
 )
 
 const SafemodeEnvironmentAnnotation = GroupName + "/environment"
@@ -80,6 +81,10 @@ func (r *Disruption) SetupWebhookWithManager(setupWebhookConfig utils.SetupWebho
 	cloudServicesProvidersManager = setupWebhookConfig.CloudServicesProvidersManager
 	chaosNamespace = setupWebhookConfig.ChaosNamespace
 	safemodeEnvironment = setupWebhookConfig.Environment
+	permittedUserGroups = map[string]struct{}{}
+	for _, group := range setupWebhookConfig.PermittedUserGroups {
+		permittedUserGroups[group] = struct{}{}
+	}
 
 	return ctrl.NewWebhookManagedBy(setupWebhookConfig.Manager).
 		For(r).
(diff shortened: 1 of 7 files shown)


chart/values.yaml
```diff
@@ -66,6 +66,8 @@ controller:
     port: 9443 # port to use to serve requests
   safeMode:
     environment: ""
+    permittedUserGroups:
+      - system:authenticated
```
ptnapoleon, line 70: Useful to use in the dev chart because it seems we have this by default in our local userinfo

api/v1beta1/disruption_webhook_test.go
```diff
@@ -502,7 +529,8 @@ func makeValidNetworkDisruption() *Disruption {
 				IntVal: 1,
 			},
 			Network: &NetworkDisruptionSpec{
-				Drop: 100,
+				Drop:  100,
+				Hosts: []NetworkDisruptionHostSpec{{Port: 80}},
```
ptnapoleon, line 533: Why add this? Well my happy case test, where i expect no errors, was giving me errors because we made it to the safety net checking for blast radius. Rather than try to disable that feature in the unit tests here, I thought it was better to just make this spec more valid

api/v1beta1/disruption_webhook.go
```diff
@@ -316,7 +329,40 @@ You first need to remove those chaos pods (and potentially their finalizers) to
...
+		return err
+	}
+
+	validGroupFound := false
+
+	for _, group := range userInfo.Groups {
+		_, ok := permittedUserGroups[group]
+		if ok {
+			validGroupFound = true
+
```
Devatoria, line 351: Mostly useless improvement but I think we can break the loop once one permitted group has been found.

api/v1beta1/disruption_webhook.go
```diff
@@ -316,7 +329,40 @@ You first need to remove those chaos pods (and potentially their finalizers) to
...
+			logger.Debugw("permitting user disruption creation, due to group membership", "group", group)
+		}
+	}
+
+	if !validGroupFound {
+		logger.Warnw("rejecting user from creating this disruption", "permittedUserGroups", permittedUserGroups, "userGroups", userInfo.Groups)
+		return fmt.Errorf("lacking sufficient authorization to create disruptions. your user groups are %s, but you must be in one of the following groups: %s", userInfo.Groups, permittedUserGroupWarningString)
+	}
+
+	return nil
```
luphaz, line 361:
```suggestion
	for _, group := range userInfo.Groups {
		_, ok := permittedUserGroups[group]
		if ok {
			logger.Debugw("permitting user disruption creation, due to group membership", "group", group)
			
			return nil
		}
	}

	logger.Warnw("rejecting user from creating this disruption", "permittedUserGroups", permittedUserGroups, "userGroups", userInfo.Groups)
	return fmt.Errorf("lacking sufficient authorization to create disruptions. your user groups are %s, but you must be in one of the following groups: %s", userInfo.Groups, permittedUserGroupWarningString)
	}

	return nil
```

chart/values.yaml
Devatoria: Could we add a comment on this new setting to explain what it does please? Especially it's only for disruptions, and it does not impact disruption cron and rollout.

Devatoria (changes requested):

luphaz (approved):

chart/values.yaml
ptnapoleon: Yep! We will probably want to implement this for the other two resources too? But they dont have webhooks yet, so that will be a followup

chart/templates/configmap.yaml
```diff
@@ -56,6 +56,10 @@ data:
         port: {{ .Values.controller.webhook.port }}
       safeMode:
         enable: {{ .Values.controller.safeMode.enable }}
+        permittedUserGroups:
+          {{- range $index, $group := .Values.controller.safeMode.permittedUserGroups }}
```
ptnapoleon, line 60: I double checked this is safe when we have nothing specified in values.yaml

luphaz (dismissed):

luphaz (approved):

aymericDD (approved):

Devatoria (approved):

PR #838, opened 2024-02-21
chaosplt-211: reject notBefore triggers in the past

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- reject any notBefore triggers set in the past during ValidateCreate, as there's no valid reason to specify those (they'd just be ignored, at best), so it indicates probably a user mistake

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_webhook_test.go
+++ b/api/v1beta1/disruption_webhook_test.go
@@ -380,7 +380,7 @@ var _ = Describe("Disruption", func() {
 						},
 					}
 
-					Expect(newDisruption.ValidateCreate().Error()).Should(ContainSubstring("only values in the future are accepted"))
+					Expect(newDisruption.ValidateCreate()).Should(MatchError(ContainSubstring("only values in the future are accepted")))
 				})
 
 				It("triggers.createPods should return an error", func() {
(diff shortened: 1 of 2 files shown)


api/v1beta1/disruption_webhook_test.go
```diff
@@ -371,6 +371,44 @@ var _ = Describe("Disruption", func() {
...
+			When("triggers.*.notBefore is in the past", func() {
+				It("triggers.inject should return an error", func() {
+					newDisruption.Spec.Duration = "30m"
+					newDisruption.Spec.Triggers = DisruptionTriggers{
+						Inject: DisruptionTrigger{
+							NotBefore: metav1.NewTime(time.Now().Add(time.Minute * 5 * -1)),
+						},
+					}
+
+					Expect(newDisruption.ValidateCreate().Error()).Should(ContainSubstring("only values in the future are accepted"))
```
aymericDD, line 383: nit you can also use the `MatchError` clause like that:

```suggestion
					Expect(newDisruption.ValidateCreate()).Should(MatchError(ContainSubstring("only values in the future are accepted")))
```

aymericDD (dismissed):
Looks good to me 👍

aymericDD (approved):
💯

craig-seeman:
Does the controller make the final decision on when something will happen? I'm curious what happens in the event the controller and the node are in different timezones~

ptnapoleon:
> Does the controller make the final decision on when something will happen? I'm curious what happens in the event the controller and the node are in different timezones~

Well, the decision on createPods is made by the controller, but `inject` is evaluated by the injector pods. It's a time.Time that's passed from the controller to those injectors, so clock drift would certainly result in a discrepancy between the controller's expectations and when the injector pods actually start. But I believe time.Time embeds timezone information in its string form, and all computers are hopefully using UTC anyway?

PR #848, opened 2024-03-19
Upgrade controller-runtime to v0.17.2

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Upgrade our controller-runtime dependency to 0.17.2 and adapt to any breaking changes
  - All Validate functions now return admission.Warnings in addition to just an error. For now I haven't added any, just changing the signature and return statements
  - We now only get Pod metadata in the Watches call in SetupwithManager(now called WatchesRawSource), reducing our memory usage
  - Require go 1.22, and adapt our CI+makefile accordingly
  - A variety of small header changes after big breaking changes in https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0
  - the Options to ctrl.NewManager have been restructured
  - @aymericDD how did you generate this mock originally? It isn't formed by `make generate-mocks`: https://github.com/DataDog/chaos-controller/pull/848/commits/d51a651aadb14e85216f15a6390ad91661aeabbb I tried updating it manually and it seems to work
  - Update controller-tools to 0.14.0 to get newest controller-gen
  - Inject our zap logger into the singleton loggers used by ctrl-runtime and apimachinery, to get slightly better formatting from their logs, but mostly so that their Info logs no longer show up as errors for us
  - Webhooks no longer get their Decoder injected, it needs set directly
  - There's now an error handler in controller-runtime that dumps a stacktrace whenever we return an error from Reconcile. As a result, I now make sure we return `nil` whenever we get a safe error that we plan on retrying

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


watchers/chaos_pod_handler.go
```diff
@@ -42,7 +42,7 @@ func NewChaosPodHandler(disruption *chaosv1beta1.Disruption, recorder record.Eve
 }
 
 // OnAdd is a handler function for the add of the chaos pod
-func (w ChaosPodHandler) OnAdd(_ interface{}) {
+func (w ChaosPodHandler) OnAdd(_ interface{}, _ bool) {
```
clairecng, line 45: what does this bool do?

.circleci/config.yml
```diff
@@ -93,9 +93,8 @@ executors:
   ubuntu:
     <<: *working_directory
     machine:
-      image: ubuntu-2004:2022.10.1
+      image: ubuntu-2204:2024.01.2
```
clairecng, line 96: is there any reason why or this was just a nice thing to do?

clairecng:
Great changes on this pr, thanks for the time you took working on this 🙇‍♀️ 
LGTM But I'm no expert on unit tests nor controller runtime things 😬

clairecng (approved):

.circleci/config.yml
ptnapoleon: The older image came with a `go` version that was too old to build after these updates

watchers/chaos_pod_handler.go
ptnapoleon: haha, well I'm not sure. It's `isInitialList`, in the Handler interface, but there's almost no explanation of it in `client-go/tools/cache/controller.go`: `- OnAdd is called when an object is added.` 

elsewhere I can see this parameter explained briefly:
> AddFunc accepts the isInInitialList parameter, for propagating HasSynced.

So I check HasSynced
> Also used to tell if the handler is synced (has had all items in the initial list delivered).

and so I decided that worst case scenario, we're doing duplicate work? Truly though, I wasn't sure if these watchers would keep working until I ran a few big disruptions with this and it went fine

PR #860, opened 2024-04-25
Add CI Tests for Disruption Client

diyarab:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- `x`

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/install/install.go
+++ b/api/v1beta1/install/install.go
@@ -17,4 +17,5 @@ import (
 func Install(scheme *runtime.Scheme) {
 	utilruntime.Must(chaosv1beta1.AddToScheme(scheme))
 	utilruntime.Must(scheme.SetVersionPriority(v1beta1.SchemeGroupVersion))
+	utilruntime.Must(chaosv1beta1.ClientSchemeBuilder.AddToScheme(scheme))
 }
--- a/api/v1beta1/register.go
+++ b/api/v1beta1/register.go
@@ -20,6 +20,7 @@ package v1beta1
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 var (
@@ -29,6 +30,13 @@ var (
 
 func addKnownTypes(scheme *runtime.Scheme) error {
 	scheme.AddKnownTypes(GroupVersion,
+		&Disruption{},
+		&DisruptionList{},
+		&DisruptionCron{},
+		&DisruptionCronList{},
+	)
+
+	scheme.AddKnownTypes(schema.GroupVersion{Group: GroupName, Version: runtime.APIVersionInternal},
 		&Disruption{},
 		&DisruptionList{},
 		&DisruptionCron{},
--- a/clientset/v1beta1/typed/v1beta1/internalversion/disruption.go
+++ b/clientset/v1beta1/typed/v1beta1/internalversion/disruption.go
@@ -26,6 +26,7 @@ type DisruptionsGetter interface {
 // DisruptionInterface has methods to work with Disruption resources.
 type DisruptionInterface interface {
 	Create(ctx context.Context, disruption *v1beta1.Disruption, opts v1.CreateOptions) (*v1beta1.Disruption, error)
+	Update(ctx context.Context, disruption *v1beta1.Disruption, opts v1.UpdateOptions) (*v1beta1.Disruption, error)
 	Delete(ctx context.Context, name string, opts v1.DeleteOptions) error
 	Get(ctx context.Context, name string, opts v1.GetOptions) (*v1beta1.Disruption, error)
 	List(ctx context.Context, opts v1.ListOptions) (*v1beta1.DisruptionList, error)
@@ -98,6 +99,20 @@ func (c *disruptions) Create(ctx context.Context, disruption *v1beta1.Disruption
 	err = c.client.Post().
 		Namespace(c.ns).
 		Resource("disruptions").
+		VersionedParams(&opts, scheme.ParameterCodec).
+		Body(disruption).
+		Do(ctx).
+		Into(result)
+	return
+}
+
+// Update takes the representation of a disruption and updates it. Returns the server's representation of the disruption, and an error, if there is any.
+func (c *disruptions) Update(ctx context.Context, disruption *v1beta1.Disruption, opts v1.UpdateOptions) (result *v1beta1.Disruption, err error) {
+	result = &v1beta1.Disruption{}
+	err = c.client.Put().
+		Namespace(c.ns).
+		Resource("disruptions").
+		Name(disruption.Name).
 		VersionedParams(&opts, scheme.ParameterCodec).
 		Body(disruption).
 		Do(ctx).
--- a/controllers/clientset_test.go
+++ b/controllers/clientset_test.go
@@ -1,3 +1,7 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024 Datadog, Inc.
 package controllers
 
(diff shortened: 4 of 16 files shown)


api/v1beta1/register.go
```diff
@@ -35,6 +36,13 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&DisruptionCronList{},
 	)
 
+	scheme.AddKnownTypes(schema.GroupVersion{Group: GroupName, Version: runtime.APIVersionInternal},
```
aymericDD, line 39: I don't think we need this.

api/v1beta1/install/install.go
```diff
@@ -17,4 +17,5 @@ import (
 func Install(scheme *runtime.Scheme) {
 	utilruntime.Must(chaosv1beta1.AddToScheme(scheme))
 	utilruntime.Must(scheme.SetVersionPriority(v1beta1.SchemeGroupVersion))
+	utilruntime.Must(chaosv1beta1.ClientSchemeBuilder.AddToScheme(scheme))
```
aymericDD, line 20: Have you pull the latest change from the main? It was already fixed by this [PR](https://github.com/DataDog/chaos-controller/pull/857).

clientset/v1beta1/typed/v1beta1/internalversion/disruption.go
```diff
@@ -26,6 +26,7 @@ type DisruptionsGetter interface {
 // DisruptionInterface has methods to work with Disruption resources.
 type DisruptionInterface interface {
 	Create(ctx context.Context, disruption *v1beta1.Disruption, opts v1.CreateOptions) (*v1beta1.Disruption, error)
+	Update(ctx context.Context, disruption *v1beta1.Disruption, opts v1.UpdateOptions) (*v1beta1.Disruption, error)
```
aymericDD, line 29: Do we need the update function? FWIK disruption are immutable. Ping @Devatoria @ptnapoleon

aymericDD (commented):
Really nice unit tests 👏 
Just let some comments about the registration and initialisation of the clientsets.

clientset/v1beta1/typed/v1beta1/internalversion/disruption.go
aymericDD: Indeed 👍 Thanks for the reply 🙏

api/v1beta1/register.go
diyarab: It is needed for the `Watch` method to work. The watcher needs `Disruption` kind to be registered in the internal version of the group in the scheme, otherwise it throws an error. I can move into `install.go` instead maybe?

ptnapoleon: This seems like the right place to put this. But how is this different than the call to `scheme.AddKnownTypes` above? Do we know why we end up needing to add it to two versions?

controllers/suite_test.go
```diff
@@ -99,18 +100,19 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
 
-		if err := mgr.Start(bgCtx); err != nil {
+		if err := mgr.Start(suiteCtx); err != nil {
 			Fail(fmt.Sprintf("unable to start manager, test can't be ran: %v", err))
 		}
 	}()
 	DeferCleanup(cancel)
 
-	Eventually(mgr.GetCache().WaitForCacheSync).WithContext(ctx).Within(k8sAPIServerResponseTimeout).ProbeEvery(k8sAPIPotentialChangesEvery).Should(BeTrue())
+	Eventually(mgr.GetCache().WaitForCacheSync).WithContext(suiteCtx).Within(k8sAPIServerResponseTimeout).ProbeEvery(k8sAPIPotentialChangesEvery).Should(BeTrue())
```
ptnapoleon, line 115: Interesting! I'm surprised that `ctx` gets cancelled prior to the entire test suite completing. Good catch

api/v1beta1/register.go
diyarab: The `clientset` generation process is dependent on the structure of the folders in the project. In this case, it has generated a `clientset` for what it interprets as the internal version, rather than for `v1beta1`. As a result, when we use the `clientset` to interact with disruptions, it attempts to manipulate the internal version. However, it can't find the resource definition because it doesn't exist. 
If we prefer to have it in a more standard path, it would need substantial modifications to the controller.

ptnapoleon (dismissed):

api/v1beta1/register.go
aymericDD: The constant [GroupVersion](https://github.com/DataDog/chaos-controller/blob/main/api/v1beta1/groupversion_info.go#L30) used line 32 is the same as the one defined [line 39](https://github.com/DataDog/chaos-controller/pull/860/files#diff-ed6c59a090681be5bf9f8de4020890ef2433aefb49a758a508383d94f5adda90R39).
So it looks duplicated. Does the unit tests are working if you remove this block of code?

diyarab: The GroupVersion on line 32 is `schema.GroupVersion{Group: 'chaos.datadoghq.com', Version: 'v1beta1'}`, the GroupVersion on line 39 is  `schema.GroupVersion{Group: 'chaos.datadoghq.com', Version: '__internal'}`. If I remove block of code between lines 39 and 44, the tests won't work. 

Here is an error message that appears when I remove that block of code:
``` go
Status:Failure,Message:an error on the server (\"unable to decode an event from the watch stream: unable to decode watch event: no kind \\\"Disruption\\\" is registered for the internal version of group "chaos.datadoghq.com\\\" in scheme \\\"/Users/diyara.bissembayeva/chaos-controller/clientset/v1beta1/scheme/register.go:17\\\"\") has prevented the request from succeeding
```
This just goes back to my comment above.

ptnapoleon: Hmm, are we doing something wrong elsewhere then? The Disruption Kind definitely should only exist in the v1beta1 version

diyarab: we use the [client-gen](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/generating-clientset.md) tool to generate our `clientset`, which relies on specific folder structures and tagging conventions to determine versions. According to the `client-gen` docs, the folder struscture is expects to find API types follows the format `pkg/apis/${GROUP}/${VERSION}/types.go`.

Currently, our project's folder structure led `client-gen` to misinterpret our main version as `internal`, rather than the intended `v1beta1`. This results in a `clientset` that tries to interact with a non-existent internal version, causing the issues we are seeing.

To address this, we would need to reorganize of the chaos-controller's folder structure. I think this is a substantial modification which falls outside the scope of the current PR.

ptnapoleon: Agreed that we wouldn't want to re-organize in this PR. But does the new clientset still let us query v1beta1? (seems like it?) Are we able to force specifying a v1beta1 version in CLI args, rather than inferring from the package structure? But fair, we can discuss this separately from this PR, imo

controllers/clientset_test.go
```diff
@@ -186,9 +186,7 @@ var _ = Describe("Disruption Client", func() {
 				// Assert
 				Eventually(func() watch.Event {
 					event := <-watcher.ResultChan()
-					log.Infow("Event received", "Type", event.Type, "Object", event.Object)
 					if event.Type == eventType {
-						log.Infow("Event matched", "Type", event.Type, "Expected", eventType, "Object", event.Object)
```
ptnapoleon: I'd rather have this in here? We could tune them down to `log.Debugw` instead of `log.Infow`? It would make it much easier to debug a failure in the future

api/v1beta1/register.go
aymericDD: Indeed, you right @diyarab the version is not the same and knows I understand the problem. 

Thanks for explanations 🙏  And I agree we should tackle this in another PR to have a correct usage of the schema.

aymericDD (approved):
Good job 👏

ptnapoleon (approved):

PR #865, opened 2024-05-08
CHAOSPLT-316: Implement authentication in the http eventnotifier

diyarab:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- This PR implements authentication in the http eventnotifier. 

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


eventnotifier/http/auth.go
```diff
@@ -0,0 +1,90 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024 Datadog, Inc.
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023 Datadog, Inc.
```
ptnapoleon, line 8: Weird double header error here?

eventnotifier/http/auth.go
```diff
@@ -0,0 +1,90 @@
...
+	tokenBytes, err := io.ReadAll(res.Body)
+	if err != nil {
+		return "", fmt.Errorf("error when reading token: %w", err)
+	}
+
+	if b.TokenPath == "" {
+		return string(tokenBytes), nil
+	}
+
+	value := gjson.Get(string(tokenBytes), b.TokenPath)
```
ptnapoleon, line 84: We're adding a new dependency for this single `Get` function, is it not possible to parse this data otherwise?

eventnotifier/http/http.go
```diff
@@ -167,21 +175,51 @@ func (n *Notifier) Notify(dis v1beta1.Disruption, event corev1.Event, notifType
...
+		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
+	}
 
+	res, err := n.client.Do(req)
 	if err != nil {
 		return fmt.Errorf("http notifier: error when sending notification: %w", err)
 	}
 
-	n.logger.Debugw("notifier: sending notifier event to http", "disruption", dis.Name, "eventType", event.Type, "message", notif.EventMessage)
+	n.logger.Debugw("http notifier: sending notifier event to http", "disruption", dis.Name, "eventType", event.Type, "message", notif.EventMessage)
```
ptnapoleon, line 192: Let's move this debug log _before_ we send the message, that way we get this info before we exit on error

eventnotifier/http/auth.go
diyarab: It is definitely possible to parse the data in a different way, but the were a couple of reasons to use `gjson`.
It allows to enable per-configuration path queries. It also provides flexibility by allowing us to query and extract data directly from JSON structures without needing to pre-define Go structs for every configuration. Finally, it is optimized for speed and efficiency, which helps avoid the overhead associated with marshaling and unmarshaling.

ptnapoleon (approved):

ptnapoleon (approved):
+1 but I would like a second reviewer (i nominated Claire but it can be anyone)

eventnotifier/http/http.go
```diff
@@ -159,25 +167,58 @@ func (n *Notifier) Notify(dis v1beta1.Disruption, event corev1.Event, notifType
...
+		if header == "" {
+			continue
+		}
+
+		splittedHeader := strings.Split(header, ":")
+		if len(splittedHeader) == 2 {
+			parsedHeaders[splittedHeader[0]] = splittedHeader[1]
+		} else {
+			return nil, fmt.Errorf("invalid headers: Must be in the format: key:value, found %s", header)
+		}
```
taihuynh167, line 216: we probably can ignore this error. Not worth throwing everything else away because of a faulty pair.

eventnotifier/http/http.go
```diff
@@ -159,25 +167,58 @@ func (n *Notifier) Notify(dis v1beta1.Disruption, event corev1.Event, notifType
...
+	parsedHeaders := make(map[string]string)
+
+	for _, header := range headers {
+		if header == "" {
+			continue
+		}
+
+		splittedHeader := strings.Split(header, ":")
+		if len(splittedHeader) == 2 {
+			parsedHeaders[splittedHeader[0]] = splittedHeader[1]
```
taihuynh167, line 213:
```suggestion
			parsedHeaders[splittedHeader[0]] = splittedHeader[1]
			key := splittedHeader[0]
			val := splittedHeader[1]
			parsedHeaders[key] = val
```

eventnotifier/types/types.go
```diff
@@ -5,8 +5,11 @@
 
 package types
 
+import "sigs.k8s.io/controller-runtime/pkg/client"
+
 type NotifiersCommonConfig struct {
 	ClusterName string
+	Client      client.Client
```
taihuynh167, line 12: nit: consider using pointer to avoid duplicating `client` object

```suggestion
	Client      *client.Client
```

taihuynh167 (approved):
I added some minor comments

eventnotifier/http/http.go
diyarab: Incorrect headers could lead to failed authentication or improper request handling, which could compromise the system's functionality and security. Thus, I am not sure it is correct to just dismiss it.

chart/values.yaml
```diff
@@ -43,6 +43,9 @@ controller:
       url: ""
       headersFilepath: ""
       headers: []
+      authURL: ""
+      authHeaders: ""
```
clairecng, line 47:
```suggestion
      authHeaders: []
```
if I'm not mistaken?

eventnotifier/http/auth.go
```diff
@@ -0,0 +1,86 @@
...
+	return bearerAuthTokenProvider{
+		logger,
+		url,
+		client,
+		headers,
+		tokenPath,
+	}
+}
+
+func (b bearerAuthTokenProvider) AuthToken(ctx context.Context) (string, error) {
```
clairecng, line 44: would be nice to have a comment to describe this method

clairecng (changes requested):
You should add documentation on what this feature do [here](https://github.com/DataDog/chaos-controller/blob/main/docs/features.md#http). (We should also definitely indicate that this feature ONLY support Authentication through a bearer token with a Authorization header!)

You could also add more comments in general I think.

Otherwise, 👍 nice one! I'll approve once the changes for documentation are made 👍

clairecng (dismissed):

eventnotifier/types/types.go
```diff
@@ -5,8 +5,11 @@
 
 package types
 
+import "sigs.k8s.io/controller-runtime/pkg/client"
+
 type NotifiersCommonConfig struct {
 	ClusterName string
+	Client      *client.Client
```
ptnapoleon, line 12: Where is this used?

diyarab: This field originated from this [POC](https://github.com/DataDog/chaos-controller/pull/782) and was used to extract information about disruptions. Since we now send a full disruption in the payload, it is no longer necessary and has been removed.

ptnapoleon (approved):

PR #868, opened 2024-06-05
fix: SEGFAULT error during cmd keepalive process

aymericDD:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- The goroutine launched by the `KeppAlive` command program, get the result from a nil ticker channel because the unlock command is released too early. 

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -80,12 +80,12 @@ type backgroundCmd struct {
 	Cmd
 	sync.Mutex
 
-	log            *zap.SugaredLogger
-	processManager process.Manager
-	ticker         *time.Ticker
-	chErr          chan error
-	chQuit         chan int
-	pid            int
+	log             *zap.SugaredLogger
+	processManager  process.Manager // Manager to interact with process
+	ticker          *time.Ticker    // Used to send regular SIGCONT signal to process
+	chErr           chan error      // Used to monitor the exit of the command
+	chKeepAliveQuit chan int        // Used to kill the keepAlive goroutine
+	pid             int             // PID of the process
 }
 
 type factory struct {
@@ -190,7 +190,7 @@ func (w *backgroundCmd) KeepAlive() {
 
 	w.ticker = time.NewTicker(cmdKeepAliveTickDuration)
 
-	w.chQuit = make(chan int)
+	w.chKeepAliveQuit = make(chan int)
 
 	w.log.Debug("monitoring sending SIGCONT signal to process every 1s")
 
@@ -218,8 +218,8 @@ func (w *backgroundCmd) sendSIGCONTSignal() (exit bool, err error) {
 	}
 
 	select {
-	case <-w.chQuit:
-		close(w.chQuit)
+	case <-w.chKeepAliveQuit:
+		close(w.chKeepAliveQuit)
 		w.log.Debug("background process exited, stopping to monitor background process, ticker removed")
 
 		return true, nil
@@ -260,8 +260,8 @@ func (w *backgroundCmd) resetTicker() {
 	w.ticker.Stop()
 	w.ticker = nil
 
-	if w.chQuit != nil {
-		w.chQuit <- 0
+	if w.chKeepAliveQuit != nil {
+		w.chKeepAliveQuit <- 0
 	}
 }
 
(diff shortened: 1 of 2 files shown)


command/command.go
```diff
@@ -84,6 +84,7 @@ type backgroundCmd struct {
 	processManager process.Manager
 	ticker         *time.Ticker
 	chErr          chan error
+	chQuit         chan int
```
ptnapoleon, line 87: Could we comment the purpose of each of these? It would help new readers

command/command_test.go
```diff
@@ -240,7 +240,7 @@ var _ = Describe("BackgroundCmd", func() {
 				SetupMockExpect(cmd, manager, proc, syscall.SIGCONT, findErr, signalErr, times)
 
 				sut.KeepAlive()
-				sut.KeepAlive() // we call it twice volountarily to ensure a single goroutine is created
+				//sut.KeepAlive() // we call it twice voluntarily to ensure a single goroutine is created
```
ptnapoleon, line 243: any idea why we used to call it twice? I don't see how that ensured that only one goroutine is created

command/command.go
```diff
@@ -175,70 +177,84 @@ func (w *backgroundCmd) Start() error {
...
-			proc, err := w.processManager.Find(w.pid)
-			if err != nil {
-				w.log.Errorw("an error occurred when trying to Find process, stopping to monitor background process, ticker removed", "error", err)
+	if w.ticker == nil {
+		return fmt.Errorf("ticker is nil")
+	}
 
-				w.resetTicker()
+	select {
+	case <-w.chQuit:
```
ptnapoleon, line 215: I don't understand this select block, won't we expect that anytime `resetTicker` is called, we end up returning an error here? I must be misunderstanding the control flow

command/command.go
```diff
@@ -175,70 +177,84 @@ func (w *backgroundCmd) Start() error {
...
 		for {
-			if w.ticker == nil {
+			if err := w.sendSIGCONTSignal(); err != nil {
 				return
 			}
+		}
+	}()
+}
 
-			<-w.safeTicker().C
```
ptnapoleon, line 199: okay, i do think i see now why safeTicker wasn't always safe when we were shutting down the injector

command/command_test.go
aymericDD: I mean we don't verify if two goroutines are created, but if another goroutines is created, methods of the processManager will not match the call expectations.

command/command.go
aymericDD: You right, in fact it's not an error. I have changed the code to be more explicit 👍

command/command.go
aymericDD: Yes, the mutex must used during the usage of the chanel.

command/command.go
```diff
@@ -175,70 +177,92 @@ func (w *backgroundCmd) Start() error {
...
-	w.Lock()
-	defer w.Unlock()
-
 	if w.ticker != nil {
 		return
 	}
 
 	w.ticker = time.NewTicker(cmdKeepAliveTickDuration)
 
+	w.chKeepAliveQuit = make(chan int)
```
ptnapoleon, line 193: nit: I dont think we need to prefix this with `ch` just to show its type in the name

ptnapoleon (approved):

command/command.go
```diff
@@ -175,70 +177,92 @@ func (w *backgroundCmd) Start() error {
...
-
 	if w.ticker == nil {
 		return
 	}
 
 	w.ticker.Stop()
 	w.ticker = nil
+
+	if w.keepAliveQuit != nil {
+		w.keepAliveQuit <- 0
```
ptnapoleon, line 264: The issue before was that `keepAliveQuit` was unbuffered, so this would block until someone read from the channel. But when we're called by `Stop()`, who has the lock, we'll never be able to read from this channel in `sendSIGCONTSignal`, as it can't get the lock. If we make it a buffered channel, this won't block anymore. Stop() can complete, and then the goroutine calling sendSIGCONTSignal can complete. I don't see any more stuck pods locally

aymericDD: Really interesting 🤔 I learn something today! Thanks for the investigation and the fix! Good job! 💪

luphaz (approved):

ptnapoleon (approved):

Issue #814, opened 2023-11-22
User Request: Add support for tolerations on injector pods


craig-seeman:
**Is your feature request related to a problem? Please describe.**
Currently the injector pods created do not support having tolerations added to them, making it hard/impossible to have injector pods spawned on certain node setups that have taints applied to them.

**Describe the solution you'd like**
Add the ability for tolerations to be added to the injector pods, defined within values.yaml, much like annotations and labels.

**Describe alternatives you've considered**
N/A

**Additional context**

ptnapoleon:
Thanks @JoshuaBell9181 for adding this feature

PR #874, opened 2024-06-26
User Request: Add support for tolerations on injector pods

JoshuaBell9181:
## What does this PR do?

- [x] Adds new functionality
- [x] Fixes a bug

https://github.com/DataDog/chaos-controller/issues/814

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.

### DIFF
--- a/chart/templates/configmap.yaml
+++ b/chart/templates/configmap.yaml
@@ -84,21 +84,23 @@ data:
         {{ $key }}: {{ $val | quote }}
         {{- end }}
       {{- end }}
+      {{- if .Values.injector.resources }}
       tolerations:
-      {{- range .Values.injector.tolerations }}
-        - key: {{ .key | quote }}
-          {{- if .operator }}
-          operator: {{ .operator | default "Equal" | quote }}
-          {{- end }}
-          {{- if .value }}
-          value: {{ .value | quote }}
-          {{- end }}
-          {{- if .effect }}
-          effect: {{ .effect | quote }}
-          {{- end }}
-          {{- if .tolerationSeconds }}
-          tolerationSeconds: {{ .tolerationSeconds }}
-          {{- end }}
+        {{- range .Values.injector.tolerations }}
+          - key: {{ .key | quote }}
+            {{- if .operator }}
+            operator: {{ .operator | default "Equal" | quote }}
+            {{- end }}
+            {{- if .value }}
+            value: {{ .value | quote }}
+            {{- end }}
+            {{- if .effect }}
+            effect: {{ .effect | quote }}
+            {{- end }}
+            {{- if .tolerationSeconds }}
+            tolerationSeconds: {{ .tolerationSeconds }}
+            {{- end }}
+        {{- end }}
       {{- end }}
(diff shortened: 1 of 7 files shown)


chart/values.yaml
```diff
@@ -87,6 +87,7 @@ injector:
     repo: chaos-injector
   annotations: {} # extra annotations passed to the chaos injector pods
   labels: {} # extra labels passed to the chaos injector pods
+  tolerations: [] # extra tolerations passed to the chaos injector pods
```
ptnapoleon, line 90: I'm truly unsure, but should this be `tolerations : {}`, since we're checking `.key` and `.value` in the chart?

JoshuaBell9181: I think we will still want an array because someone might want to add multiple tolerations.
If we left it as a map it would only support 1 key.

Checkout the kube docs here you can see its a list!
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#:~:text=has%20two%20tolerations%3A-,tolerations%3A,effect%3A%20%22NoExecute%22,-In%20this%20case

ptnapoleon (approved):
Looks great, thanks! I'm going to test the chart templating out a bit, and then merge

chart/templates/configmap.yaml
```diff
@@ -84,6 +84,22 @@ data:
         {{ $key }}: {{ $val | quote }}
         {{- end }}
       {{- end }}
+      tolerations:
```
ptnapoleon, line 87: Could we wrap this key in an `{{- if .Values.injector.tolerations }}` `{{- end }}` block, the way that `annotations` and `labels` are?

chart/values.yaml
ptnapoleon, line 90: Could you add a full commented example toleration? Much like `allowedHosts` or `kubeDNS` keys just below this? Otherwise a user will only know the possible options from reading the code

ptnapoleon (changes requested):
Left two small requests

chart/templates/configmap.yaml
JoshuaBell9181: Good call updated!

chart/values.yaml
JoshuaBell9181: Updated!

chart/templates/configmap.yaml
```diff
@@ -84,21 +84,23 @@ data:
         {{ $key }}: {{ $val | quote }}
         {{- end }}
       {{- end }}
+      {{- if .Values.injector.resources }}
```
ptnapoleon:
```suggestion
      {{- if .Values.injector.tolerations }}
```
why resources?

ptnapoleon (approved):

ptnapoleon:
Thank you, @JoshuaBell9181!

chart/templates/configmap.yaml
JoshuaBell9181: You are to fast !
Copilot auto complete, i noticed after committing lol

PR #878, opened 2024-07-02
feat: verify mocks in the CI

aymericDD:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

- `Verify if mocks are updated.`

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -274,6 +274,11 @@ workflows:
           diff-fails: true
           requires: [prepare-env]
       - go-make:
+          name: mocks
+          target: generate-mocks
+          diff-fails: true
+          requires: [prepare-env]
+      - go-make:
           name: vet
           target: vet
           requires: [prepare-env]


.circleci/config.yml
```diff
@@ -273,6 +273,11 @@ workflows:
           target: manifests
           diff-fails: true
           requires: [prepare-env]
+      - go-make:
+          name: mocks
+          target: generate-mocks
```
luphaz, line 278: I would not expect mock to self generate/update on their own in the CI

Usually, you should instead FAIL if mocks are not properly generated (compare with git diff and exit)

aymericDD: How can I ensure the mocks are updatdate if I don't generate them in the CI and compare the diff' ?

luphaz: The PR as is was not performing any validation or diff, correct?

Let me rephrase then

> I would not expect mock to self generate/update on their own in the CI


I would not expect mock to ONLY self generate/update on their own in the CI WITHOUT performing any follow up assertion.

luphaz: Indeed I misinterpreted the configuration file 😄 

Thanks for clarifying

.circleci/config.yml
```diff
@@ -273,6 +273,12 @@ workflows:
           target: manifests
           diff-fails: true
           requires: [prepare-env]
+      - go-make:
+          executor: ubuntu
```
ptnapoleon, line 277: The other checks above run on the executor `golang`, no? I think they inherit from prepare-env. Does it still fail if you don't specify an executor here?

ptnapoleon, line 277:
```suggestion
```
testing ci without a specified executor...

ptnapoleon: oof, the `go` deps are present now, as expected, but because we run `make header-fix` after generating mocks, we also need our python deps. Not sure how to handle that, as they're on different executors. I see why you were planning to go with `ubuntu` here

ptnapoleon (approved):

PR #880, opened 2024-07-12
Create events for disruption cron component

aymericDD:
## What does this PR do?

- [x] Adds new functionality
- [ ] Fixes a bug

> Upgrade mockery to **v2.43.0**

> Create a webhooks for the **DisruptionCron** component to send events to notifiers through the broadcaster.

List of events:
- `EventDisruptionCronCreated`
- `EventDisruptionCronUpdated`
- `EventDisruptionCronDeleted`

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


chart/templates/webhook.yaml
aymericDD, line 1: Not sure if this webhook is correct, but it's working locally 😇

eventbroadcaster/notifiersink.go
```diff
@@ -29,34 +27,48 @@ type NotifierSink struct {
...
+	s.logger.Infof("CREATE event received: %s", event.Message)
+
+	var obj client.Object
+
+	switch event.InvolvedObject.Kind {
+	case v1beta1.DisruptionKind:
+		disruption, err := s.getDisruption(event)
+		if err != nil {
+			s.logger.Warn(err)
+			return event, fmt.Errorf("eventnotifier: unable to get Disruption object")
```
aymericDD, line 53: Here I have changed the behavior by returning an error rather `nil`, to retry the event. If we return `nil` the event will not be retried and it will increase the chance to miss events.

eventbroadcaster/notifiersink.go
```diff
@@ -29,34 +27,48 @@ type NotifierSink struct {
...
+		}
+
+		obj = &disruption
+	case v1beta1.DisruptionCronKind:
+		disruptionCron, err := s.getDisruptionCron(event)
+		if err != nil {
+			s.logger.Warn(err)
+			return event, fmt.Errorf("eventnotifier: unable to get DisruptionCron object")
```
aymericDD, line 61: Here I have changed the behavior by returning an error rather `nil`, to retry the event. If we return `nil` the event will not be retried and it will increase the chance to miss events.

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+var (
+	disruptionCronWebhookLogger     *zap.SugaredLogger
+	disruptionCronWebhookRecorder   record.EventRecorder
+	disruptionCronWebhookDeleteOnly bool
+)
+
+func (d *DisruptionCron) SetupWebhookWithManager(setupWebhookConfig utils.SetupWebhookWithManagerConfig) error {
+	disruptionCronWebhookRecorder = setupWebhookConfig.Recorder
+	disruptionCronWebhookDeleteOnly = setupWebhookConfig.DeleteOnlyFlag
+	disruptionCronWebhookLogger = setupWebhookConfig.Logger.With("source", "admission-controller")
```
ptnapoleon, line 29: can we distinguish between the disruption webhook logger and the disruption cron webhook logger sources?

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+
+//+kubebuilder:webhook:webhookVersions={v1},path=/validate-chaos-datadoghq-com-v1beta1-disruptioncron,mutating=false,failurePolicy=fail,sideEffects=None,groups=chaos.datadoghq.com,resources=disruptioncrons,verbs=create;update;delete,versions=v1beta1,name=vdisruptioncron.kb.io,admissionReviewVersions={v1,v1beta1}
+
+var _ webhook.Validator = &DisruptionCron{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (d *DisruptionCron) ValidateCreate() (admission.Warnings, error) {
+	log := disruptionCronWebhookLogger.With("disruptionCronName", d.Name, "disruptionCronNamespace", d.Namespace)
+
+	log.Debugw("validating created disruption cron", "spec", d.Spec)
```
ptnapoleon, line 44:
```suggestion
	log.Infow("validating created disruption cron", "spec", d.Spec)
```

api/v1beta1/events.go
```diff
@@ -10,83 +10,91 @@ import (
...
-	// Event only attached to a target
-	TargetEvent DisruptionEventCategory = "TargetEvent"
-	// Event only attached to the disruption
-	DisruptEvent DisruptionEventCategory = "DisruptionEvent"
-	// Event only attached to a chaos pod
-	ChaosPodEvent DisruptionEventCategory = "ChaosPodEvent"
+	// TargetEvent only attached to a target
+	TargetEvent EventCategory = "TargetEvent"
+	// DisruptEvent only attached to the disruption
+	DisruptEvent EventCategory = "Event"
```
ptnapoleon, line 24: Shouldn't this still be DisruptionEvent, if the event below is DisruptionCronEvent?

chart/templates/webhook.yaml
```diff
@@ -44,6 +44,41 @@ webhooks:
     - disruptions
 ---
 apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
```
ptnapoleon, line 47: I think this webhook should be wrapped with  `{{ .Values.controller.disruptionCronEnabled }}`

eventnotifier/datadog/datadog.go
```diff
@@ -53,6 +54,35 @@ func (n *Notifier) GetNotifierName() string {
 	return string(types.NotifierDriverDatadog)
 }
 
+// Notify generates a notification for generic k8s events
+func (n *Notifier) Notify(obj client.Object, event corev1.Event, notifType types.NotificationType) error {
+	dis, ok := obj.(*v1beta1.Disruption)
```
ptnapoleon, line 59: But we end up calling this for disruptionCrons too, aren't we? in notifiersink.go, we end up passing DisruptionCrons to parseEventToNotifier which calls Notify

eventnotifier/http/http.go
```diff
@@ -266,3 +211,129 @@ func (n *Notifier) marshalUserGroups(groups []string) (string, error) {
...
+	if err != nil {
+		return fmt.Errorf("http notifier: couldn't marshal notification: %w", err)
+	}
+
+	logger.With("eventType", event.Type, "message", notifierEvent.EventMessage)
+
+	return n.emitEvent(jsonNotif, logger)
+}
+
+func (n *Notifier) notifyDisruptionCron(disruptionCron v1beta1.DisruptionCron, event corev1.Event, notifType types.NotificationType) error {
```
ptnapoleon, line 248: Why did we choose to implement disruption cron events in http notifier but not the datadog notifier? I don't think the api we have listening for disruptions is yet prepared for disruption crons

eventnotifier/slack/slack.go
```diff
@@ -110,12 +111,17 @@ func (n *Notifier) buildSlackBlocks(dis v1beta1.Disruption, notifType types.Noti
 }
 
 // Notify generates a notification for generic k8s events
-func (n *Notifier) Notify(dis v1beta1.Disruption, event corev1.Event, notifType types.NotificationType) error {
-	headerText := utils.BuildHeaderMessageFromDisruptionEvent(dis, notifType)
+func (n *Notifier) Notify(obj client.Object, event corev1.Event, notifType types.NotificationType) error {
+	dis, ok := obj.(*v1beta1.Disruption)
```
ptnapoleon, line 115: Okay I'm now seeing we're explicitly only implementing http notifications for crons. why not slack too?

main.go
```diff
@@ -361,8 +364,17 @@ func main() {
...
 	}
+
+	logger.Debug("setup webhook for disruption")
+
 	if err = (&chaosv1beta1.Disruption{}).SetupWebhookWithManager(setupWebhookConfig); err != nil {
-		logger.Fatalw("unable to create webhook", "webhook", chaosv1beta1.DisruptionKind, "error", err)
+		logger.Fatalw("unable to create webhook for disruption", "webhook", chaosv1beta1.DisruptionKind, "error", err)
+	}
+
+	logger.Debug("setup webhook for disruption cron")
```
ptnapoleon, line 374: These need to be gated by `if cfg.Controller.DisruptionCronEnabled`

eventnotifier/datadog/datadog.go
aymericDD: Yes, we do, but to reduce the size of the PR I just add the support of `DisruptionCron` for the http notifier. This is why I added this condition.

eventnotifier/http/http.go
aymericDD: In order to reduce the size of the PR I just add the support of `DisruptionCron` for the **http notifier**. Indeed, you make a good point. I should ensure the old fields are still there to not break the existing disruption registering, but the goal of my work is to send `DisruptionCron` to our API.

eventnotifier/slack/slack.go
aymericDD: Also in order to reduce the size of the PR. I can do it in another PR.

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// delete-only mode, reject everything trying to be created
+	if disruptionCronWebhookDeleteOnly {
```
clairecng, line 47: I'm wondering if it's really necessary. We can imagine a user wanting to setup their cron disruptions during a freeze no?

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return nil, nil
+}
+
+func (d *DisruptionCron) ValidateUpdate(_ runtime.Object) (warnings admission.Warnings, err error) {
+	log := logger.With("disruptionCronName", d.Name, "disruptionCronNamespace", d.Namespace)
+
+	log.Debugw("validating updated disruption cron", "spec", d.Spec)
+
+	// delete-only mode, reject everything trying to be created
```
clairecng, line 62:
```suggestion
	// delete-only mode, reject everything trying to be updated
```
Same comment as above here.

eventbroadcaster/notifiersink.go
```diff
@@ -29,34 +27,48 @@ type NotifierSink struct {
...
 
 	broadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: corev1Client.Events("")})
-
-	return
 }
 
 func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
-	dis, err := s.getDisruption(event)
-	if err != nil {
+	s.logger.Infof("CREATE event received: %s", event.Message)
```
clairecng, line 44: is it really necessary? I feel like it would generate a lot of logs
If yes:

```suggestion
	s.logger.Debugf("CREATE event received: %s", event.Message)
```

eventbroadcaster/notifiersink.go
clairecng: in what kind of situation would it suceed again after an error? How many times does it try to retry?

eventnotifier/datadog/datadog.go
clairecng: nitpick: why not follow the same code design as the http notified and just not do anything for the cron disruption? Would be clearer I think:

```
func (n *Notifier) Notify(obj client.Object, event corev1.Event, notifType types.NotificationType) error {
// we only support disruption notifications on Datadog for now
	switch d := obj.(type) {
	case *v1beta1.Disruption:
		return n.notifyDisruption(*d, event, notifType)
	default:
		return nil
}
```

eventnotifier/http/http.go
clairecng: we should create a jira ticket to support all types of notifiers

eventnotifier/noop/noop.go
```diff
@@ -35,9 +36,15 @@ func (n Notifier) GetNotifierName() string {
...
-	n.log.Debugf("NOOP: %s for disruption %s\n", fmt.Sprintf("Notifier %s: %s - %s", string(notifType), event.Reason, event.Message), dis.Name)
+// Notify generates a notification for generic k8s events
+func (n Notifier) Notify(obj client.Object, event corev1.Event, notifType types.NotificationType) error {
+	switch d := obj.(type) {
+	case *v1beta1.Disruption:
+		n.log.Debugf("NOOP: %s for disruption %s\n", fmt.Sprintf("Notifier %s: %s - %s", string(notifType), event.Reason, event.Message), d.Name)
+	case *v1beta1.DisruptionCron:
+		n.log.Debugf("NOOP: %s for disruption cron %s\n", fmt.Sprintf("Notifier %s: %s - %s", string(notifType), event.Reason, event.Message), d.Name)
+	default:
+	}
```
clairecng, line 47: nitpick?:
```suggestion
        notifierMessage := fmt.Sprintf("Notifier %s: %s - %s", string(notifType), event.Reason, event.Message)
	
	switch d := obj.(type) {
	case *v1beta1.Disruption:
		n.log.Debugf("NOOP: %s for disruption %s\n", notifierMessage, d.Name)
	case *v1beta1.DisruptionCron:
		n.log.Debugf("NOOP: %s for disruption cron %s\n", notifierMessage, d.Name)
	default:
	}
```

eventbroadcaster/notifiersink.go
aymericDD: The number of retry is hardcoded in the k8s library:
- https://github.com/DataDog/chaos-controller/blob/main/vendor/k8s.io/client-go/tools/record/event.go#L39
- https://github.com/DataDog/chaos-controller/blob/main/vendor/k8s.io/client-go/tools/record/event.go#L293

api/v1beta1/disruption_cron_webhook.go
aymericDD: Good question, but I don't think if we should do exceptions when the `deleteOnly` mode is enable.

eventbroadcaster/notifiersink.go
aymericDD: > in what kind of situation would it suceed again after an error?

The event can be emitted  and handled before the creation of the resource. In this case, the k8s client will not find the resource and returns an error. So in order to succeed it should retry until the resource is found.

eventnotifier/http/http.go
ptnapoleon: I think we're only going to get errors from any crons we send to the api, as it isn't prepared to listen for those yet. Could we change the structure of our config to enable/disable notifications for each notifier type for each resource type separately?

i.e., instead of
```
http:
  enabled: true
```
could it look more like
```
http:
  enabled:
    disruptions: true
    disruptioncrons: false
```
? wdyt @diyarab @clairecng @aymericDD

aymericDD: I have already tested functionally the new payload of the callback and it works with the chaos-api so no need to add an extra configuration 👍

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -0,0 +1,85 @@
...
+	return nil, nil
+}
+
+func (d *DisruptionCron) ValidateUpdate(_ runtime.Object) (warnings admission.Warnings, err error) {
+	log := logger.With("disruptionCronName", d.Name, "disruptionCronNamespace", d.Namespace)
+
+	log.Infow("validating updated disruption cron", "spec", d.Spec)
+
+	// delete-only mode, reject everything trying to be created
+	if disruptionCronWebhookDeleteOnly {
```
ptnapoleon, line 66: The chaos-controller allows for disruptions to be updated in deleteOnly mode, and I think the same should be true for crons. Sometimes we need to update the status. As long as we aren't creating new resources, or running any disruptions, other business can continue in deleteOnly mode.

api/v1beta1/events.go
```diff
@@ -185,79 +193,97 @@ var Events = map[DisruptionEventReason]DisruptionEvent{
...
 		OnDisruptionTemplateMessage: "Created disruption injection pod for \"%s\"",
-		Category:                    DisruptEvent,
+		Category:                    DisruptionEvent,
 	},
 	EventDisruptionCreated: {
 		Type:                        corev1.EventTypeNormal,
 		Reason:                      EventDisruptionCreated,
-		OnDisruptionTemplateMessage: "Disruption created",
-		Category:                    DisruptEvent,
+		OnDisruptionTemplateMessage: "DisruptionEvent created",
```
ptnapoleon, line 255: I think this block of event messages got caught up in a string replace? These should look like
Disruption Created
DisruptionCron created
DisruptionCron deleted
and so on,
not `DisruptionEvent cron updated`

eventnotifier/http/http.go
```diff
@@ -46,28 +48,48 @@ type Notifier struct {
 	authTokenProvider BearerAuthTokenProvider
 }
 
-type HTTPNotifierEvent struct {
+// NotifierEvent represents a notification event
+type NotifierEvent struct {
```
ptnapoleon, line 52: Since this struct is never used outside the http package, could we either keep the http prefix or stop exporting this struct?

eventnotifier/http/http.go
```diff
@@ -46,28 +48,48 @@ type Notifier struct {
...
+
+type DisruptionEvent struct {
+	DisruptionName string `jsonapi:"attribute" json:"disruption_name"`
+	Disruption     string `jsonapi:"attribute" json:"disruption"`
+	TargetsCount   int    `jsonapi:"attribute" json:"targets_count"`
+}
+
+type DisruptionCronEvent struct {
+	DisruptionCronName string `jsonapi:"attribute" json:"disruption_cron_name"`
+	DisruptionCron     string `jsonapi:"attribute" json:"disruption_cron"`
```
ptnapoleon, line 85: Could we make it clearer that the Disruption and DisruptionCron fields should be the full manifests?

eventnotifier/http/http.go
ptnapoleon: It feels uncomfortable to merge code to main, ungated by a config value, that we expect not to work? It's not always true (we ship bugs, ofc), but ideally any commit on main should be usable for a release

eventbroadcaster/notifiersink.go
ptnapoleon: Could we mention somewhere in a comment that automatic retrying is handled by client-go up to 12 times?

eventnotifier/http/http.go
aymericDD: This struct is used in our internal api to deserialized the body of requests sent by the http notifier of the chaos-controller.

eventnotifier/http/http.go
aymericDD:
```suggestion
type DisruptionCronEvent struct {
	NotifierEvent
	Name     string `jsonapi:"attribute" json:"name"`
	Manifest string `jsonapi:"attribute" json:"manifest"`
}
```
What do you think of this struct?

aymericDD:
```
type DisruptionCronEvent struct {
	NotifierEvent
	Name     string `jsonapi:"attribute" json:"name"`
	Manifest string `jsonapi:"attribute" json:"manifest"`
}

type DisruptionEvent struct {
	NotifierEvent
	Name         string `jsonapi:"attribute" json:"name"`
	Manifest     string `jsonapi:"attribute" json:"manifest"`
	TargetsCount int    `jsonapi:"attribute" json:"targets_count"`
}

```

What do you think of those structs?

ptnapoleon: I certainly think it's more clear. And we're breaking the api anyway with the restructuring, so I'm +1. @diyarab and @clairecng ?

eventnotifier/http/http.go
```diff
@@ -266,3 +221,134 @@ func (n *Notifier) marshalUserGroups(groups []string) (string, error) {
...
+		return fmt.Errorf("http notifier: couldn't marshal disruption: %w", err)
+	}
+
+	notifierEvent.DisruptionEvent = &DisruptionEvent{
+		TargetsCount:   len(dis.Status.TargetInjections),
+		DisruptionName: dis.Name,
+		Disruption:     string(disruptionStr),
+	}
+
+	// Deprecated fields
```
ptnapoleon, line 248: Fantastic to keep these set still, it makes the migration so much easier

aymericDD:
Replaced by this PR: https://github.com/DataDog/chaos-controller/pull/885

PR #886, opened 2024-07-23
Minor improvements + bugfixes to safemode

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

- Rename some functions for clarity
- Fix a bug where we were estimating targeted pods incorrectly, by not filtering on namespace
- Includes debug logging, which was necessary to identify these bugs
- Fix a bug where we didn't count total pods or nodes correctly when the count was over 1,000
- Fix a bug where we didn't consider network.services to be a filter on network traffic
- Skip estimating targets if the thresholds are set to 100%

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


api/v1beta1/disruption_webhook.go
```diff
@@ -609,8 +617,20 @@ func safetyNetCountNotTooLarge(r *Disruption) (bool, string, error) {
...
+	logger.Debugw("comparing estimated target count to total existing targets",
+		"disruptionName", r.Name,
+		"disruptionNamespace", r.Namespace,
+		"namespaceThreshold", namespaceThreshold,
+		"clusterThreshold", clusterThreshold,
+		"estimatedEligibleTargetsCount", targetCount,
+		"namespaceCount", namespaceCount,
+		"totalCount", totalCount,
+		"calculatedPercentOfTotal", userCountVal/float64(totalCount),
+		"estimatedTargetCount", userCountVal,
```
ptnapoleon, line 629: I was unsure if we should try to make these tags at all intelligible to users who don't have the code in front of them, so I did so. The alternative is to capture every intermediate variable and tag them with their exact names, e.g. `"targetCount", targetCount`

api/v1beta1/disruption_webhook_test.go
```diff
@@ -602,6 +601,91 @@ var _ = Describe("Disruption", func() {
...
+				newDisruption = makeValidNetworkDisruption()
+			})
+
+			AfterEach(func() {
+				k8sClient = nil
+				newDisruption = nil
+			})
+
+			When("only services are defined", func() {
+				Context("", func() {
```
aymericDD, line 626:
```suggestion
```

api/v1beta1/disruption_webhook_test.go
```diff
@@ -602,6 +601,91 @@ var _ = Describe("Disruption", func() {
...
+
+						_, err := newDisruption.ValidateCreate()
+
+						Expect(err).ShouldNot(HaveOccurred())
+					})
+				})
+			})
+
+			When("no filters are defined", func() {
+				Context("", func() {
```
aymericDD, line 675:
```suggestion
```

api/v1beta1/disruption_webhook_test.go
```diff
@@ -602,6 +601,91 @@ var _ = Describe("Disruption", func() {
...
+
+						_, err := newDisruption.ValidateCreate()
+
+						Expect(err).ShouldNot(HaveOccurred())
+					})
+				})
+			})
+
+			When("various host filters are defined", func() {
+				Context("should pass validation", func() {
```
aymericDD, line 639:
```suggestion
			        DescribeTable("should pass validation", func(hosts []NetworkDisruptionHostSpec) {
			                // Arrange
					newDisruption.Spec.Network.Hosts = hosts
 
                                        // Action
					_, err := newDisruption.ValidateCreate()

                                        // Assert
					Expect(err).ShouldNot(HaveOccurred())
				},
					Entry("with just a port filter", []NetworkDisruptionHostSpec{{Port: 80}}),
					Entry("with just a protocol filter", []NetworkDisruptionHostSpec{{Protocol: "tcp"}}),
					Entry("with just a hostname filter", []NetworkDisruptionHostSpec{{Host: "localhost"}}),
					Entry("with all possible host filters together", []NetworkDisruptionHostSpec{{Host: "localhost", Port: 443, Protocol: "udp"}}),
				)
```

nit

api/v1beta1/disruption_webhook.go
```diff
@@ -460,18 +460,18 @@ func (r *Disruption) initialSafetyNets() ([]string, error) {
 		}
 
 		if r.Spec.Network != nil {
-			if caught := safetyNetNeitherHostNorPort(*r); caught {
-				logger.Debugw("the specified disruption either contains no Hosts or contains a Host which has neither a port nor a host. The more ambiguous, the larger the blast radius.", "SafetyNet Catch", "Network")
+			if caught := safetyNetMissingNetworkFilters(*r); caught {
+				logger.Debugw("the specified disruption either contains no Host or Service filters. This will result in all network traffic being affected.", "SafetyNet Catch", "Network")
```
aymericDD, line 464: nit: The variable `logger` is shadowing the global `logger` variable of the `v1beta1` package. It will be preferable to rename it to avoid side effects during the execution of unit tests.

api/v1beta1/disruption_webhook.go
```diff
@@ -508,6 +508,11 @@ func safetyNetCountNotTooLarge(r *Disruption) (bool, string, error) {
 		}
 	}
 
+	if namespaceThreshold >= 1 && clusterThreshold >= 1 {
```
aymericDD, line 511:
```suggestion
	if namespaceThreshold >= MaxNamespaceThreshold && clusterThreshold >= MaxClusterThreshold {
```
nit: use constant rather magic numbers

api/v1beta1/disruption_webhook.go
```diff
@@ -626,43 +644,45 @@ func safetyNetCountNotTooLarge(r *Disruption) (bool, string, error) {
...
 }
 
-// safetyNetNeitherHostNorPort is the safety net regarding missing host and port values.
-// it will check against all defined hosts in the network disruption spec to see if any of them have a host and a
-// port missing. The more generic a hosts tuple is (Omitting fields such as port), the bigger the blast radius.
-func safetyNetNeitherHostNorPort(r Disruption) bool {
+// safetyNetMissingNetworkFilters is the safety net regarding missing host, service, or port filters.
+// it will check if any filters have been placed to limit which network traffic is disrupted
+func safetyNetMissingNetworkFilters(r Disruption) bool {
 	if r.Spec.Unsafemode != nil && r.Spec.Unsafemode.DisableNeitherHostNorPort {
```
aymericDD, line 650: This part is not covered by unit test

api/v1beta1/disruption_webhook.go
```diff
@@ -626,43 +644,45 @@ func safetyNetCountNotTooLarge(r *Disruption) (bool, string, error) {
...
 		return false
 	}
 
-	// if hosts are not defined, this also falls into the safety net
-	if r.Spec.Network.Hosts == nil || len(r.Spec.Network.Hosts) == 0 {
-		return true
+	if r.Spec.Network == nil {
```
aymericDD, line 654: This part is not covered by unit test

api/v1beta1/disruption_webhook.go
```diff
@@ -626,43 +644,45 @@ func safetyNetCountNotTooLarge(r *Disruption) (bool, string, error) {
...
 
 	for _, path := range r.Spec.DiskFailure.Paths {
 		if strings.TrimSpace(path) == "/" {
-			return true, "the specified path for the disk failure disruption targeting a node must not be \"/\"."
+			return true
 		}
 	}
 
-	return false, ""
+	return false
```
aymericDD, line 687: This part is not covered by unit test

api/v1beta1/disruption_webhook_test.go
```diff
@@ -514,7 +514,6 @@ var _ = Describe("Disruption", func() {
 
 			JustBeforeEach(func() {
 				newDisruption = makeValidDiskFailureDisruption()
-				controllerutil.AddFinalizer(newDisruption, chaostypes.DisruptionFinalizer)
```
aymericDD, line 517: Why we remove the finalizer here?

aymericDD (dismissed):
Nice improvement 💯  I let some suggestions concerning unit tests and some other, but nothing required.

api/v1beta1/disruption_webhook_test.go
ptnapoleon: It wasn't relevant to the testing at all, seems to have only been included because this was copy-pasted from the finalizer tests at the top of the file

api/v1beta1/disruption_webhook.go
ptnapoleon: I don't think we're shadowing the global logger, but directly using it? What side effects will we see?

api/v1beta1/disruption_webhook.go
ptnapoleon: I knew this looked wrong, great idea

api/v1beta1/disruption_webhook.go
ptnapoleon: Disagree? We don't test it explicitly, but any test that validates a disruption w/o a network spec and gets no errors, is confirming this works.

api/v1beta1/disruption_webhook.go
ptnapoleon: Good point, I'll add

api/v1beta1/disruption_webhook.go
```diff
@@ -508,6 +508,11 @@ func safetyNetCountNotTooLarge(r *Disruption) (bool, string, error) {
 		}
 	}
 
+	if namespaceThreshold >= 1 && clusterThreshold >= 1 {
+		// Don't waste time or memory counting, if we allow 100% of resources to be targeted
+		return false, "", nil
+	}
+
 	if r.Spec.Level == chaostypes.DisruptionLevelPod {
```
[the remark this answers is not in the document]
ptnapoleon: I agree, but I'll open a ticket to do so in a follow-up PR

api/v1beta1/disruption_webhook.go
aymericDD: It doesn't because it's not a target level pod.

api/v1beta1/disruption_webhook.go
aymericDD: So does this conditions is required? It looks this condition is made before the call of this one and it's a private method.

api/v1beta1/disruption_webhook.go
aymericDD: image

aymericDD: I made the comment at the wrong plage 🙈 It's line 117

ptnapoleon: oh yes, this is true. But it's taking it directly from the global logger, and it isn't new to this PR, but how we've had that logger used forever. If we want to change that, I'm 100% on board, but it's definitely unrelated to these changes

api/v1beta1/disruption_webhook.go
ptnapoleon: We do check it in the caller, that's true, but it's still safe to double-check here too? We don't want to rely on the caller and accidentally panic when we check Network.Services

api/v1beta1/disruption_webhook.go
ptnapoleon: A pod level spec would return false at L678. A node level spec with a valid path returns false here at L687. I see now that we don't test any valid paths in the unit tests. I'll add another test

api/v1beta1/disruption_webhook_test.go
```diff
@@ -602,6 +610,79 @@ var _ = Describe("Disruption", func() {
...
+				newDisruption = makeValidNetworkDisruption()
+			})
+
+			AfterEach(func() {
+				k8sClient = nil
+				newDisruption = nil
+			})
+
+			When("only services are defined", func() {
+				Context("testing ValidateCreate", func() {
```
aymericDD, line 635:
```suggestion
```

api/v1beta1/disruption_webhook_test.go
```diff
@@ -602,6 +610,79 @@ var _ = Describe("Disruption", func() {
...
+				},
+					Entry("with just a port filter", []NetworkDisruptionHostSpec{{Port: 80}}),
+					Entry("with just a protocol filter", []NetworkDisruptionHostSpec{{Protocol: "tcp"}}),
+					Entry("with just a hostname filter", []NetworkDisruptionHostSpec{{Host: "localhost"}}),
+					Entry("with all possible host filters together", []NetworkDisruptionHostSpec{{Host: "localhost", Port: 443, Protocol: "udp"}}),
+				)
+			})
+
+			When("no filters are defined", func() {
+				Context("testing ValidateCreate", func() {
```
aymericDD, line 663:
```suggestion
```

aymericDD (approved):
LGTM, but if you can remove the `Context` in comments it will be perfect 👌

api/v1beta1/disruption_webhook_test.go
```diff
@@ -503,6 +503,90 @@ var _ = Describe("Disruption Webhook", func() {
...
+				})
+
+				It("should reject the disruption at node level", func() {
+					allowNodeFailure = false
+					newDisruption.Spec.Level = chaostypes.DisruptionLevelNode
+
+					_, err := newDisruption.ValidateCreate()
+
+					Expect(err).Should(HaveOccurred())
+					Expect(err.Error()).Should(ContainSubstring("at least one of the initial safety nets caught an issue"))
```
aymericDD, line 540:
```suggestion
					Expect(err).Should(MatchError(ContainSubstring("at least one of the initial safety nets caught an issue")))
```

nit

aymericDD (dismissed):

aymericDD (approved):

PR #892, opened 2024-07-31
Reduce warning and error logs from notifiers

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- We're no longer warning on every event that comes from the controller's `recordEventOnTarget`
- We're no longer logging an error when the user info is missing in a slack notification, as that happens once for every disruption
- Includes the resource annotations in the Create event sent by the validation webhooks. This makes sure the UserInfo is available in the notifier

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -250,7 +250,7 @@ func (r *Disruption) ValidateCreate() (admission.Warnings, error) {
 	}
 
 	// send informative event to disruption to broadcast
-	recorder.Event(r, Events[EventDisruptionCreated].Type, string(EventDisruptionCreated), Events[EventDisruptionCreated].OnDisruptionTemplateMessage)
+	recorder.AnnotatedEventf(r, r.Annotations, Events[EventDisruptionCreated].Type, string(EventDisruptionCreated), Events[EventDisruptionCreated].OnDisruptionTemplateMessage)
 
 	return nil, nil
 }
--- a/api/v1beta1/disruption_webhook.go
+++ b/api/v1beta1/disruption_webhook.go
@@ -250,7 +250,7 @@ func (r *Disruption) ValidateCreate() (admission.Warnings, error) {
 	}
 
 	// send informative event to disruption to broadcast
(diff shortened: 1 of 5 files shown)


eventnotifier/slack/slack.go
```diff
@@ -241,7 +241,7 @@ func (n *Notifier) buildSlackMessage(obj client.Object, event corev1.Event, noti
 	}
 
 	if err != nil {
-		logger.Errorw("unable to retrieve user info", "error", err)
```
aymericDD, line 244: IMO it's a warning because without the user info we are not able to send the notification the user 😕

eventbroadcaster/notifiersink.go
```diff
@@ -46,7 +46,7 @@ func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
 
 	obj, err := s.getObject(event)
 	if err != nil {
-		s.logger.Warn(err)
+		s.logger.Debug(err) // Involved object is a target Pod or Node, we don't send notifications for these.
```
aymericDD, line 49: 👍

eventnotifier/slack/slack.go
ptnapoleon: Yes, this was a hard choice. It's definitely a warning if it happens when we don't expect, but we do expect it at least once (so far). Maybe I can find a way not to send any events until we have the UserInfo

api/v1beta1/disruption_webhook.go
```diff
@@ -250,7 +250,7 @@ func (r *Disruption) ValidateCreate() (admission.Warnings, error) {
 	}
 
 	// send informative event to disruption to broadcast
-	recorder.Event(r, Events[EventDisruptionCreated].Type, string(EventDisruptionCreated), Events[EventDisruptionCreated].OnDisruptionTemplateMessage)
```
clairecng, line 253: 😄

clairecng (approved):

PR #893, opened 2024-08-01
fix: rework the disruption cron events

aymericDD:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Add the disruption cron resource serialized in the annotation of the disruption cron events.

## Code Quality Checklist

## Testing

    - `x`
    - [x] locally.


eventbroadcaster/eventbroadcaster.go
```diff
@@ -18,3 +18,13 @@ func EventBroadcaster() record.EventBroadcaster {
 
 	return eventBroadcaster
 }
+
+func DisruptionCronEventBroadcaster() record.EventBroadcaster {
+	correlator := record.CorrelatorOptions{
```
aymericDD, line 23: This broadcaster is dedicated to the disruption cron reconciler, because the broadcaster drops events. I'm not confident of this configuration, but it works. For what I understand here, we allow 1000 events to be sent per seconds before the aggregation, but not sure at all.

controllers/disruption_cron_controller.go
```diff
@@ -57,11 +62,42 @@ func (r *DisruptionCronReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			if err := r.emmitEvent(instance, chaosv1beta1.EventDisruptionCronDeleted); err != nil {
+				r.log.Warnw("failed to emit event", "err", err)
+			}
+		}
+
 		return ctrl.Result{}, nil
 	}
 
+	if controllerutil.ContainsFinalizer(instance, chaostypes.DisruptionCronFinalizer) {
+		if err := r.emmitEvent(instance, chaosv1beta1.EventDisruptionCronUpdated); err != nil {
```
ptnapoleon, line 86: We're going to send this event every time the Disruption is reconciled without a deletion timestamp, except for the very first reconciliation. That removes any use of this to indicate that a user updated the resource. If we wanted to count reconciliations, a metric would be better.

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -92,8 +88,5 @@ func (d *DisruptionCron) ValidateDelete() (warnings admission.Warnings, err erro
 
 	log.Infow("validating deleted disruption cron", "spec", d.Spec)
 
-	// send informative event to disruption cron to broadcast
-	disruptionCronWebhookRecorder.Event(d, Events[EventDisruptionCronDeleted].Type, string(EventDisruptionCronDeleted), Events[EventDisruptionCronDeleted].OnDisruptionTemplateMessage)
```
ptnapoleon, line 96: If the goal was to send the resource via an annotation, why did we need to move it to the reconcile loop? We can use the same approach in the webhook

ptnapoleon (changes requested):

eventbroadcaster/notifiersink.go
```diff
@@ -65,56 +65,69 @@ func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
...
+			if disruptionCronString == "" {
+				return nil, fmt.Errorf("eventnotifier: empty disruptionCron annotation")
+			}
+
+			if err := json.Unmarshal([]byte(disruptionCronString), &disruptionCron); err != nil {
+				return nil, fmt.Errorf("eventnotifier: failed to unmarshal disruptionCron from annotation: %w", err)
+			}
+		} else {
+			s.logger.Debugw("event does not have disruptionCron annotation")
+			s.logger.Debugw("fetching disruptionCron from API server")
```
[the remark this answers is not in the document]
aymericDD: I put it during the development. Should I keep it?

types/types.go
```diff
@@ -120,9 +120,10 @@ const (
 	// DisruptionNamespaceLabel is the label used to identify the disruption namespace for a chaos pod. This is used to determine pod ownership.
 	DisruptionNamespaceLabel = GroupName + "/disruption-namespace"
 
-	finalizerPrefix     = "finalizer." + GroupName
-	DisruptionFinalizer = finalizerPrefix
-	ChaosPodFinalizer   = finalizerPrefix + "/chaos-pod"
+	finalizerPrefix         = "finalizer." + GroupName
+	DisruptionFinalizer     = "disruption." + finalizerPrefix
+	DisruptionCronFinalizer = "disruptioncron." + finalizerPrefix
```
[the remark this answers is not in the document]
aymericDD: 🤯 Thanks for noticing this regression and also the inconsistency 🙈

api/v1beta1/events.go
```diff
@@ -258,19 +258,19 @@ var Events = map[EventReason]Event{
 	EventDisruptionCronCreated: {
 		Type:                        corev1.EventTypeNormal,
 		Reason:                      EventDisruptionCronCreated,
-		OnDisruptionTemplateMessage: "DisruptionCron created",
+		OnDisruptionTemplateMessage: "DisruptionCron created with hash %s",
```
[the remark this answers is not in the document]
aymericDD: I made this change for the `EventDisruptionCronUpdated` because a user can edit multiple times a disruption cron.  The problem is the agregatorre only allow the first emitted event. By customizing the message it consider the event as a new one and don't drop it. For what I know of disruption a same event is not sent twice time because a disruption is immutable. Also for disruption it's working in this state and I want to modify it the less as I can to reduce the risk of regression.

controllers/disruption_cron_controller.go
aymericDD: By adding the hash spec of the disruption in the message, the event will be not dropped only if the spec has changed, and we want to track those kind of changes.

api/v1beta1/disruption_cron_webhook.go
aymericDD: The webhook is okay, if the chaos-api is up and running. In the scenario where the API is down and the chaos-controller failed to sent the event and is restarted the event will never be re-emitted. By emitting from the reconciler we are sure to re-emit the events and reduce the risk of losing events.

eventbroadcaster/eventbroadcaster.go
aymericDD: From my observation It seems the aggregator is dropping event even if its message is different when it reach the limit. So if we set 2 events per 300 seconds, the third event emitted with a same key  will be dropped by the aggregatore and we will not have the last state of the updated disruption cron. Only the first will be broadcasted to the notifiers. Another behavior I have noticed is, if a second event is emitted with the same message and the same key linked to the same object, it will be dropped even if it has not reach the limit of events. So by adding the hash of the spec inside the message we are only taking in account the modifications of the spec, because the message will of the event will be the same as the previous one. 
 
The 1000 events per seconds is really huge and I used this configuration for my test cases. I agree we should reduce to something more realistic. I propose to set the max events to 10 during a period of 300 seconds. It's completely arbitrary, but it will allow the user to edit a disruption cron an enough amount of times. WDYT?

ptnapoleon (changes requested):
As discussed in Slack, I would prefer (but feel free to rebut, always!):

1) We leave these three events in the webhook, and use annotations to pass down the resource info
2) We add a _new_ event, that goes in the reconcile loop
3) We don't separate broadcasters by resource, but rather by notifier. 1 for the http notifier, and 1 for the others. The http broadcaster should send far more events

aymericDD:
I relocated the broadcaster from the disruption cron reconciler to the webhooks. 
I will divide step 2 and step 3 into two separate PRs.

ptnapoleon (approved):

eventbroadcaster/notifiersink.go
```diff
@@ -65,56 +65,63 @@ func (s *NotifierSink) Create(event *corev1.Event) (*corev1.Event, error) {
...
-				},
+		var (
+			disruption v1beta1.Disruption
+			err        error
+		)
+
+		annotations := event.GetAnnotations()
+
+		// If the event has the annotation, use it to unmarshal the disruption
+		if disruptionString, ok := annotations[v1beta1.EventDisruptionAnnotation]; ok {
```
ptnapoleon, line 78: We aren't setting this yet, could you implement this change in the disruption_webhook.go as well, please?

ptnapoleon (changes requested):

eventbroadcaster/notifiersink.go
ptnapoleon: I am now having a serious headache trying to remember, but in https://github.com/DataDog/chaos-controller/pull/892/files , I passed down the UserInfo in AnnotatedEvents, but I think I accidentally (or intentionally?) removed the commit that consumes that UserInfo annotation in the notifiersink.go.

So Disruptions still are missing UserInfo sometimes, even though I had a working patch, I think my rebases messed that up.

Thank you kindly, but could you please add to `disruption_webhook.go` the same changes you made to `disruption_cron_webhook.go`? 🙏

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -92,8 +100,25 @@ func (d *DisruptionCron) ValidateDelete() (warnings admission.Warnings, err erro
 
 	log.Infow("validating deleted disruption cron", "spec", d.Spec)
 
+	// During the validation of the deletion the timestamp does not exist so we need to set it before emitting the event
+	d.DeletionTimestamp = &metav1.Time{Time: time.Now()}
```
ptnapoleon, line 104: Okay this worries me a tiny bit. We don't need to do something similar in disruptions (as they live and send events for several minutes after being deleted), but the discrepancy here between the real timestamp and this one could confuse users.

We can make sure that any servers consuming the http events, are willing to take the latest deletion timestamp they get? That way if new events come later, that have the real deletionTimestamp of the object, those are used? wdyt?

ptnapoleon (approved):

PR #894, opened 2024-08-02
DisruptionCrons should validate their DisruptionTemplates

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Before I update all the unit tests, do we like this approach? DisruptionTemplates won't have any selectors, so I can't call the original Validate functions. But I obviously want to re-use the code, so I've added an optional requireSelectors argument

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_cron_webhook_test.go
+++ b/api/v1beta1/disruption_cron_webhook_test.go
@@ -366,9 +366,11 @@ var _ = Describe("DisruptionCron Webhook", func() {
 })
 
 func makeValidDisruptionCron() *DisruptionCron {
+	validDisruption := makeValidNetworkDisruption()
 	return &DisruptionCron{
 		TypeMeta: metav1.TypeMeta{
 			Kind: DisruptionCronKind,
 		},
+		Spec: DisruptionCronSpec{DisruptionTemplate: validDisruption.Spec},
 	}
 }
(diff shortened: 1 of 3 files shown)


diyarab (dismissed):

api/v1beta1/disruption_cron_webhook_test.go
```diff
@@ -121,6 +120,23 @@ var _ = Describe("DisruptionCron Webhook", func() {
 				})
 			})
 
+			When("disruptionTemplate is invalid", func() {
+				When("the count is invalid", func() {
```
ptnapoleon, line 124: So I could add multiple tests and use DescribeTable for the various entries. But a valid DisruptionSpec is quite long, so it gets verbose. I could pass in a transformation function that takes a valid spec and breaks it in a specific way. But I feel the various cases for why a DisruptionSpec gets rejected are tested elsewhere. As long as we confirm we do validate the spec.disruptionTemplate of a DisruptionCron using DisruptionSpec's Validate, we can trust that all other cases are covered indirectly

clairecng: I'm ok with only testing like this as well.
Could we add a small comment there explaining that though?

clairecng (approved):

api/v1beta1/disruption_cron_webhook_test.go
ptnapoleon, line 124:
```suggestion
				When("the count is invalid", func() {
				// Other forms of invalid disruptions are covered by the disruption_webook_test.go
				// we just want to confirm we do validate the disruptionTemplate as part of the disruption cron webhook
```

aymericDD (approved):

PR #904, opened 2024-09-05
DisruptionCrons should not halt if they've missed 100 starts

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Add a warning log when `getNextSchedule` finds that `now` is earlier than the last scheduled disruption
- DisruptionCrons should not halt when they miss more than 100 starts. Previously, the only remedy was to wipe their status or delete/re-create them. It is sufficiently safe for them to just keep trying
- `getNextSchedule` should not take `DelayedStartTolerance` into account. By doing so, it was making it impossible to ever send the TooLate metric, because we wouldn't consider starts we had missed for being too late

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_cron_controller.go
+++ b/controllers/disruption_cron_controller.go
@@ -289,6 +289,8 @@ func (r *DisruptionCronReconciler) getNextSchedule(instance *chaosv1beta1.Disrup
 	}
 
 	if earliestTime.After(now) {
+		r.log.Warnw("getNextSchedule has found itself in the past", "earliestTime", earliestTime.GoString(), "now", now.GoString())
+
 		return time.Time{}, sched.Next(now), nil
 	}
 
--- a/controllers/disruption_cron_controller.go
+++ b/controllers/disruption_cron_controller.go
@@ -288,15 +288,6 @@ func (r *DisruptionCronReconciler) getNextSchedule(instance *chaosv1beta1.Disrup
 		earliestTime = instance.ObjectMeta.CreationTimestamp.Time
 	}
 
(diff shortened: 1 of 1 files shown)


aymericDD (approved):
LGTM 👍

controllers/disruption_cron_controller.go
```diff
@@ -289,27 +288,14 @@ func (r *DisruptionCronReconciler) getNextSchedule(instance *chaosv1beta1.Disrup
...
 
-	if instance.Spec.DelayedStartTolerance.Duration() > 0 {
-		// controller is not going to schedule anything below this point
-		schedulingDeadline := now.Add(-instance.Spec.DelayedStartTolerance.Duration())
-
-		if schedulingDeadline.After(earliestTime) {
-			earliestTime = schedulingDeadline
-		}
-	}
-
```
diyarab, line 300: Note: without the `earliestTime` adjustment, the controller has more opportunities to run the disruptions within the allowed window. This will lead to increased attempts to execute the disruption as soon as resources are available, rather than potentially skipping them.

ptnapoleon: Could you share the spec / status / scheduling gap you think will cause multiple attempts? I believe we're still going to return a single missedRun and the same `next` scheduled time. We don't allow concurrent reconciles, and we definitely dont allow concurrent disruptions, so i dont think we'll have any problems

diyarab (approved):

PR #909, opened 2024-09-19
CHAOSPLT-367: When in DeleteOnly mode, the chaos-controller should attempt to remove all disruptions

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Sometimes we need an emergency procedure to remove all existing disruptions. It's not enough to simply refuse to start new ones. The DeleteOnly flag will now do both

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -241,10 +241,6 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			return ctrl.Result{}, nil
 		}
 	} else {
-		if err := r.validateDisruptionSpec(instance); err != nil {
-			return ctrl.Result{Requeue: false}, err
-		}
-
 		// initialize all safety nets for future use
 		if instance.Spec.Unsafemode == nil || !instance.Spec.Unsafemode.DisableAll {
 			// initialize all relevant safety nets for the first time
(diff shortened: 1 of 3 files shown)


controllers/disruption_controller.go
```diff
@@ -908,17 +915,6 @@ func (r *DisruptionReconciler) emitKindCountMetrics(instance *chaosv1beta1.Disru
...
-	err := instance.Spec.Validate()
-	if err != nil {
-		r.recordEventOnDisruption(instance, chaosv1beta1.EventInvalidSpecDisruption, err.Error(), "")
-
-		return err
-	}
-
-	return nil
-}
-
```
diyarab, line 921: why remove this?

controllers/disruption_controller.go
```diff
@@ -241,8 +242,14 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{}, nil
 		}
 	} else {
-		if err := r.validateDisruptionSpec(instance); err != nil {
-			return ctrl.Result{Requeue: false}, err
```
diyarab, line 245: Are we not going to be validating the spec? I know we do it in a webhook, do you think it is redundant to do it during reconcile loop?

aymericDD: If I understand correctly, the validation occurs in the validation webhook. So no need to verify it in the reconcile loop.

aymericDD (approved):

controllers/disruption_controller.go
ptnapoleon: I did look back on the git blame, and I added this check years before we had the webhooks. It's now worse than redundant, it's actively harmful. If we somehow have a disruption with an invalid spec get past the webhooks, we actually can't delete it or remove it's finalizer via the reconciler, we'd have to do that all manually with kubectl

PR #931, opened 2024-10-17
Allow for overlapping disruptions when the first disruption is complete

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- we have no unit tests for this feature, and i think doing so in CI would be tricky with the current setup. What I could do, is test that if we create a second disruption, we still only have the expected number of chaos pods? So rather than testing the change, we test we didn't regress?
- If a disruption has completed, and all of its injectors exited in a healthy state, then it should be 100% safe for a new disruption to overlap the same targets. We don't need to wait for the injector pods to be removed, as long as they're cleaned and in a Completed state.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -1150,9 +1150,9 @@ func (r *DisruptionReconciler) getEligibleTargets(ctx context.Context, instance 
 		}
 
 		// skip targets already targeted by a chaos pod from another disruption with the same kind if any
-		if len(chaosPods) != 0 {
-			anyChaosPodsRunning := false
 
+		anyChaosPodsRunning := false
+		if len(chaosPods) != 0 {
 			// chaosPods might all be in a Completed state. If any of these injectors aren't terminated, or terminated in an unready state, we set alreadyDisrupted
 			// because if all chaosPods for this target _are_ Completed, then this target is not already injected, and the current instance can proceed
 			for _, chaosPod := range chaosPods {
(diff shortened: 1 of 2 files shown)


controllers/disruption_controller.go
```diff
@@ -1150,7 +1150,21 @@ func (r *DisruptionReconciler) getEligibleTargets(ctx context.Context, instance
...
 		// skip targets already targeted by a chaos pod from another disruption with the same kind if any
+		anyChaosPodsRunning := false
+
 		if len(chaosPods) != 0 {
+			// chaosPods might all be in a Completed state. If any of these injectors aren't terminated, or terminated in an unready state, we set alreadyDisrupted
+			// because if all chaosPods for this target _are_ Completed, then this target is not already injected, and the current instance can proceed
+			for _, chaosPod := range chaosPods {
+				for _, containerStatuses := range chaosPod.Status.ContainerStatuses {
+					if containerStatuses.State.Terminated == nil || containerStatuses.State.Terminated.ExitCode != 0 {
+						anyChaosPodsRunning = true
```
aymericDD, line 1161:
```suggestion
						anyChaosPodsRunning = true
						break
```

nit: I don't think it's useful to continue the iteration after setting the anyChaosPodsRunning variable to true 🤔

aymericDD (approved):

controllers/disruption_controller.go
ptnapoleon: great point!

aymericDD (approved):
💯

PR #941, opened 2024-12-03
CHAOSPLT-531: Add a config for the minimum cron frequency

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- One might consider a disruption that runs every minute to be too often. One might even want to prevent that. Now we can!

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/api/v1beta1/disruption_cron_webhook.go
+++ b/api/v1beta1/disruption_cron_webhook.go
@@ -130,6 +130,15 @@ func (d *DisruptionCron) ValidateUpdate(oldObject runtime.Object) (admission.War
 		return nil, err
 	}
 
+	// If a DisruptionCron is already more frequent than the minimum frequency, we don't want to
+	// block updates on other parts of the spec or status. But if the schedule is being updated,
+	// we do want to enforce that it happens more often than the minimum frequency
+	if oldObject.(*DisruptionCron).Spec.Schedule != d.Spec.Schedule {
+		if err := d.validateMinimumFrequency(minimumCronFrequency); err != nil {
+			return nil, err
+		}
+	}
+
 	// send informative event to disruption cron to broadcast
 	d.emitEvent(EventDisruptionCronUpdated)
 
--- a/api/v1beta1/disruption_cron_webhook.go
+++ b/api/v1beta1/disruption_cron_webhook.go
@@ -219,7 +219,15 @@ func (d *DisruptionCron) validateMinimumFrequency(minFrequency time.Duration) er
 	nextDisruptionStarts := schedule.Next(now)
 	nextDisruptionCompletes := nextDisruptionStarts.Add(specDuration)
 
+	// Measure, "frequency", the time between when we would schedule the next two disruptions.
+	// We don't want to measure from "now", because the cron standard will try to run at whole intervals, e.g.,
+	// a schedule for "every 15 minutes", created at 1:05, will try to run the first disruption at 1:15. So we find the next two intervals,
+	// which would be 1:15 and 1:30, and find the difference
 	frequency := schedule.Next(nextDisruptionStarts).Sub(nextDisruptionStarts)
+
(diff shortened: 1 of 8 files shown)


api/v1beta1/disruption_cron_webhook.go
```diff
@@ -196,3 +203,36 @@ func (d *DisruptionCron) validateDisruptionCronSpec() error {
 
 	return nil
 }
+
+func (d *DisruptionCron) validateMinimumFrequency(minFrequency time.Duration) error {
```
aymericDD, line 207: The Schedule is part of the disruption cron spec. Is it better to do it in the `validateDisruptionCronSpec` method instead? 
The `validateDisruptionCronSpec`  already checks if the `spec.Schedule` is valid.  
WDYT of calling the `validateMinimumFrequency` in the `validateDisruptionCronSpec` method instead?

api/v1beta1/disruption_cron_webhook.go
```diff
@@ -196,3 +203,36 @@ func (d *DisruptionCron) validateDisruptionCronSpec() error {
...
+	schedule, err := cron.ParseStandard(d.Spec.Schedule)
+	if err != nil {
+		return fmt.Errorf("spec.Schedule must follow the standard cron syntax: %w", err)
+	}
+
+	specDuration := defaultDuration
+	if d.Spec.DisruptionTemplate.Duration.Duration() > 0 {
```
aymericDD, line 214: To cover this case, can you add this test in the `success cases` of the `ValidateCreate` method:

``` golang
When("the disruption template duration is greater than 0", func() {
    It("should validate the disruption cron successfully", func() {
	    // Arrange
	    disruptionCron := makeValidDisruptionCron()
	    disruptionCron.Spec.DisruptionTemplate.Duration = "1s"
    
	    disruptionCronJSON, err := json.Marshal(disruptionCron)
	    Expect(err).ShouldNot(HaveOccurred())
    
	    expectedAnnotation := map[string]string{
		    EventDisruptionCronAnnotation: string(disruptionCronJSON),
	    }
    
	    By("sending the EventDisruptionCronCreated event to the broadcast")
	    mockEventRecorder := mocks.NewEventRecorderMock(GinkgoT())
	    mockEventRecorder.EXPECT().
		    AnnotatedEventf(
			    disruptionCron,
			    expectedAnnotation,
			    Events[EventDisruptionCronCreated].Type,
			    string(EventDisruptionCronCreated),
			    Events[EventDisruptionCronCreated].OnDisruptionTemplateMessage,
		    )
    
	    disruptionCronWebhookRecorder = mockEventRecorder
    
	    // Act
	    warnings, err := disruptionCron.ValidateCreate()
    
	    // Assert
	    Expect(warnings).To(BeNil())
	    Expect(err).ShouldNot(HaveOccurred())
    })
})
```

aymericDD (approved):
💯 

I let some comments with suggestions, but nothing required.  🙇

ptnapoleon:
This needs to check in ValidateUpdate, if the user just altered the frequency to be more/too often

api/v1beta1/disruption_cron_webhook_test.go
```diff
@@ -498,6 +605,24 @@ var _ = Describe("DisruptionCron Webhook", func() {
 				})
 			})
 
+			When("the schedule frequency becomes shorter", func() {
+				It("should not return an error", func() {
```
aymericDD, line 609:
```suggestion
				It("should return an error", func() {
```

aymericDD (approved):
💯

api/v1beta1/disruption_cron_webhook.go
aymericDD: Thanks for the explanations 🙇 
It makes sense for me 👍

PR #944, opened 2024-12-18
CHAOSPLT-68: Replace ddmark with go-validator

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- The kubebuilder struct annotations are only used to generate the CRD spec yaml, which is then validated by the k8s api.
- We would like to validate these requirements against specs even w/o access to the CRD spec file or kubernetes, only making basic code calls to Spec.Validate(). We used to use ddmark, but now we replace that with go-validator, an MIT licensed open source package.
- Supplements 100% of disruption (crons to come in another PR) Spec kubebuilder annotations with struct tags. Adds a validation of all struct tags to the Spec.Validate() function
- Adds custom translations to improve error messages
- Adds unit tests

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


api/v1beta1/validations.go
```diff
@@ -118,6 +123,120 @@ func ValidateCount(count *intstr.IntOrString) error {
 	return nil
 }
 
+func newGoValidator() (*validator.Validate, ut.Translator, error) {
+	englishLocale := en.New()
+	uni := ut.New(englishLocale, englishLocale)
+
+	translator, _ := uni.GetTranslator("en")
+
+	validate := validator.New()
```
nathantournant, line 132: Small note related to the following:
> Special Notes:
  If new to using validator it is highly recommended to initialise it using the WithRequiredStructEnabled option which is opt-in to new behaviour that will become the default behaviour in v11+. See documentation for more details.
  `validate := validator.New(validator.WithRequiredStructEnabled())`

https://pkg.go.dev/github.com/go-playground/validator/v10#readme-baked-in-validations

api/v1beta1/grpc_disruption.go
```diff
@@ -45,20 +45,20 @@ var ErrorMap = map[string]codes.Code{
 type GRPCDisruptionSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
-	Port      int                  `json:"port"`
-	Endpoints []EndpointAlteration `json:"endpoints"`
+	Port      int                  `json:"port" validate:"gte=1,lte=65535"`
+	Endpoints []EndpointAlteration `json:"endpoints" validate:"dive"`
```
aymericDD, line 49: What dive means? Recursively?

api/v1beta1/validations.go
```diff
@@ -118,6 +123,146 @@ func ValidateCount(count *intstr.IntOrString) error {
 	return nil
 }
 
+func newGoValidator() (*validator.Validate, ut.Translator, error) {
```
aymericDD, line 126: WDYT of splitting this function in multiple functions to reduce its size to facilitate its readability?

api/v1beta1/validations.go
ptnapoleon: Thanks for pointing this out! I also ran into this document, but I could have very easily missed that

api/v1beta1/grpc_disruption.go
ptnapoleon: Yes! Nested structs will be recursed automatically, but slices and maps need to be tagged with "dive" explicitly for validation to occur on each item. Where's the best place to explain things like this? In CONTRIBUTING.md?

api/v1beta1/disruption_types.go
```diff
@@ -747,7 +751,7 @@ func (s DisruptionSpec) validateGlobalDisruptionScope(requireSelectors bool) (re
 		}
 	}
 
-	if s.GRPC != nil && s.Level != chaostypes.DisruptionLevelPod {
+	if s.GRPC != nil && s.Level == chaostypes.DisruptionLevelNode {
```
ptnapoleon, line 754: Turns out that the "defaults" aren't appropriately set when we directly call Spec.Validate(), so grpc didn't realize an empty string meant pods.

api/v1beta1/validations.go
```diff
@@ -118,6 +123,168 @@ func ValidateCount(count *intstr.IntOrString) error {
...
```
clairecng, line 126:
```suggestion
// newGoValidator instantiate a validator which will validate required, greater-than, lower-than and one-of fields.
func newGoValidator() (*validator.Validate, ut.Translator, error) {
```

api/v1beta1/validations.go
```diff
@@ -118,6 +123,168 @@ func ValidateCount(count *intstr.IntOrString) error {
...
+	if err != nil {
+		return fmt.Errorf("could not validate struct tags: %w", err)
+	}
+
+	err = validate.Struct(s)
+
+	if err != nil {
+		// this check is only needed when your code could produce
+		// an invalid value for validation such as interface with nil
+		// value most including myself do not usually have code like this.
```
clairecng, line 268: can you clarify this sentence `value most including myself`? Sorry I don't 100% understand what this means 💀

api/validator_test.go
```diff
@@ -110,6 +110,382 @@ var _ = Describe("Validator", func() {
 	JustBeforeEach(func() {
 		err = validator.Validate()
 	})
+
+	Describe("validating top-level spec", func() {
```
clairecng, line 114: in most of your cases, I think you could probably do a describeTable instead? I _think_ it would take less lines to achieve the same result.

clairecng (approved):
lgtm!

api/v1beta1/validations.go
ptnapoleon: I have literally no idea, it makes no sense, maybe my IDE auto-added it? I'm sure the last thing I typed was "with nil value". I'll fix up

api/v1beta1/validations.go
```diff
@@ -118,6 +123,168 @@ func ValidateCount(count *intstr.IntOrString) error {
...
+}
+
+func registerGteTranslation(validate *validator.Validate, translator ut.Translator) error {
+	return validate.RegisterTranslation("gte", translator, func(ut ut.Translator) error {
+		// The {idx} values are interpolated using the arguments to the ut.T("gte", ...) call in the function below
+		return ut.Add("gte", "{0} is set to {1}, but must be greater or equal to {2}", true)
+	}, func(ut ut.Translator, fe validator.FieldError) string {
+		// values that are constrained by "gte" can include: int, *int, uint
+		// fe.Value() is an interface{}, so we need to check each of these type options, one by one
+		i, ok := fe.Value().(int)
```
aymericDD, line 168: This code looks duplicated with the registerLteTranslation function. WDYT of creating a private function to deduplicate the code.

Example:

``` golang
func registerTranslation(key string, ut ut.Translator, fe validator.FieldError) string {
	// values that are constrained by "lte" can include: int, *int, uint
	// fe.Value() is an interface{}, so we need to check each of these type options, one by one
	i, ok := fe.Value().(int)
	if !ok {
		iPtr, k := fe.Value().(*int)
		if !k {
			unsignedVal, k3 := fe.Value().(uint)
			if !k3 {
				// this will be directly seen by the user if their field fails validation.
				return fmt.Sprintf("could not determine value %v for field %s", fe.Value(), fe.Field())
			}

			i = int(unsignedVal)
		} else {
			if iPtr == nil {
				i = 0
			} else {
				i = *iPtr
			}
		}
	}

	iStr := strconv.Itoa(i)
	t, _ := ut.T(key, fe.Namespace(), iStr, fe.Param())

	return t
}
```

api/v1beta1/validations.go
```diff
@@ -118,6 +123,168 @@ func ValidateCount(count *intstr.IntOrString) error {
...
+			return fmt.Sprintf("could not permit value \"%v\" for field %s, try one of \"%s\"", fe.Value(), fe.Field(), userOptionsString)
+		}
+
+		t, _ := ut.T("oneofci", fe.Namespace(), userChoiceStr, userOptionsString)
+
+		return t
+	})
+}
+
+func ValidateStructTags(s DisruptionSpec) error {
```
aymericDD, line 260: Does this function needs to be public?

aymericDD: Also why not link this function to the struct itself?

aymericDD (approved):
Nice 👌

api/v1beta1/validations.go
ptnapoleon: It didn't need to be public. Instead of linking it to the struct, I replaced the arg with `interface {}` and I'll re-use it identically for DisruptionCrons later

PR #953, opened 2025-01-27
CHAOSPLT-580: Send metric when cron or rollout self-deletes

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- Allows for configuring the 24h "target missing" threshold for crons or rollouts
- Sends a metric when a cron or rollout is deleted because its target has been missing for too long

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_rollout_controller.go
+++ b/controllers/disruption_rollout_controller.go
@@ -242,6 +242,8 @@ func (r *DisruptionRolloutReconciler) handleTargetResourceMissingPastExpiration(
 		return fmt.Errorf("failed to delete instance: %w", err)
 	}
 
+	r.handleMetricSinkError(r.MetricsSink.MetricMissingTargetDeleted(DisruptionRolloutTags))
+
 	return nil
 }
 
--- a/controllers/disruption_rollout_controller.go
+++ b/controllers/disruption_rollout_controller.go
@@ -24,11 +24,12 @@ import (
 var DisruptionRolloutTags = []string{}
 
 type DisruptionRolloutReconciler struct {
-	Client      client.Client
-	Scheme      *runtime.Scheme
-	BaseLog     *zap.SugaredLogger
-	log         *zap.SugaredLogger
-	MetricsSink metrics.Sink
+	Client                         client.Client
+	Scheme                         *runtime.Scheme
+	BaseLog                        *zap.SugaredLogger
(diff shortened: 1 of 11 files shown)


o11y/metrics/metrics.go
```diff
@@ -49,6 +49,7 @@ type Sink interface {
 	MetricTooLate(tags []string) error
 	MetricTargetMissing(duration time.Duration, tags []string) error
 	MetricMissingTargetFound(tags []string) error
+	MetricMissingTargetDeleted(tags []string) error
```
clairecng, line 52: we a growing list of  metrics now... Should we rename them to be more explicit on its usage?
For example in this file we have `MetricCleaned` but it could be confusing. 

And I'd also like to move the comment declaration on the struct itself as well, so that in one look we understand each metric. 

It's probably unrealistic to have both the code and external documentation on each metric we emit so if we have a very clear code file it's fine with me!

controllers/disruption_rollout_controller.go
```diff
@@ -241,6 +242,8 @@ func (r *DisruptionRolloutReconciler) handleTargetResourceMissingPastExpiration(
 		return fmt.Errorf("failed to delete instance: %w", err)
 	}
 
+	r.handleMetricSinkError(r.MetricsSink.MetricMissingTargetDeleted(DisruptionRolloutTags))
```
clairecng, line 245: should we have a different metric for the rollout? How do we differentiate rollout vs cron?

ptnapoleon: It is a different metric, they dont share a prefix:
```

	metricPrefixRolloutController = "chaos.rollout.controller."
	metricPrefixCronController    = "chaos.cron.controller."
```

o11y/metrics/metrics.go
ptnapoleon: I do think having the comments here would be nice. Would we want to duplicate them, or just move them out of the implementation code? renaming everything isn't bad, but that's another PR

clairecng: I think duplicating is better? But then we'll need to remember to keep both files updated on comments.

clairecng (approved):
LGTM

PR #954, opened 2025-01-28
Remove unused selector cache from old dynamic targeting implementation

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- While looking at our metrics, I went to see what the selector cache metric was for. I found that this cache and metric was added in #523. The functionality is now handled differently, thanks to the watchers from #681. However, this map and metric remain. We don't ever fill this map. We don't ever read from this map. I've checked our use of this metric, and it's consistently zero, which matches the expectation that we aren't using this map. I've removed all this code

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -70,19 +70,12 @@ type DisruptionReconciler struct {
 	log                        *zap.SugaredLogger
 	SafetyNets                 []safemode.Safemode
 	ExpiredDisruptionGCDelay   *time.Duration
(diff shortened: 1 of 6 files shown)


clairecng (approved):
we didn't test this behavior either?

🧽

ptnapoleon:
I believe we tested dynamic targeting, yes, just not directly the cache behind it

PR #959, opened 2025-02-11
Move the DisruptionSpec Explain feature into the api package

ptnapoleon:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

- This is very useful, but we'd like it to be available as a library, not just as CLI output using fmt.Println. I'm moving it to api, out of chaosli, and updating the explanations

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


api/v1beta1/disruption_types.go
```diff
@@ -875,6 +883,145 @@ func (s DisruptionSpec) DisruptionCount() int {
...
+	}
+
+	if s.OnInit {
+		explanation = append(explanation, fmt.Sprintf("spec.onInit is true. "+
+			"The disruptions will be launched during the initialization of the targeted pods."+
+			"This requires some extra setup on your end, please [read the full documentation](https://github.com/DataDog/chaos-controller/blob/main/docs/features.md#applying-a-disruption-on-pod-initialization)"))
+	}
+
+	countSuffix := ""
+	if s.Count.Type == intstr.Int {
```
clairecng, line 954: here should we also specify that a 100 int value is considered to be 100%? Since it had previously led to confusion for the user

api/v1beta1/network_disruption.go
```diff
@@ -714,3 +835,74 @@ func (s *NetworkDisruptionSpec) UpdateHostsOnCloudDisruption(cloudManager clouds
...
+
+	if len(s.Hosts) != 0 {
+		explanation = append(explanation, "\tnetwork.hosts will apply filters so that the chosen network failure affects the traffic between the target and the following groups:")
+		for _, host := range s.Hosts {
+			explanation = append(explanation, host.Explain())
+		}
+	}
+
+	if len(s.Services) != 0 {
+		explanation = append(explanation, "\tnetwork.services is a convenience feature for targeting kubernetes services in the _same kubernetes cluster_. The chaos-controller will handle "+
```
clairecng, line 885: 👍

luphaz: Should we get rid of this?

In the end it does not seem that convenient and causes confusion IMO

Most people focus on using this instead of FQDN, and as we are multi-cluster, this is causing a gap in their ability to properly create "cluster agnostic" specs.

If we get rid of this feature (cluster specific feature), we can instead
- have a single way to define the spec, less cognitive overhead
- stop having question about why it does not work when the service is not on the same cluster?
- provide simple working out of the box example, that just works, always

Not saying we should get rid of this in the community edition of the chaos-controller, but in our internal documentation, IMO, we should stop advertising about this, it is counter productive

clairecng: we don't have much internal documentation to get rid of tbh...

luphaz: I would really like us to connect with Doggo maintainer so we can hijack the GitHub documentation for Datadog engineers and prevent them for doing mistakes

As we are redirecting them to this documentation, it would be ideal to fix this documentation for Datadog engineers, and avoid presenting them informations they can't use.

With Doggo, we should be able to hijack the github pages, and tailor the information for our internal use

aymericDD (approved):
Thanks!💪 

_RIP the emojis_ 😿

api/v1beta1/disruption_types.go
ptnapoleon: It's not! We reject an int value of 100 on create, because users thought it meant 100%. I'll try to clarify that here.

clairecng: yeah I meant the whole behavior around 100 :oops:

api/v1beta1/network_disruption.go
```diff
@@ -631,6 +702,30 @@ func (h NetworkDisruptionHostSpec) Validate() error {
 	return nil
 }
 
+func (h NetworkDisruptionHostSpec) Explain() string {
+	hostExplanation := ""
+
+	if h.Flow == FlowIngress {
+		fmt.Println("ACKS to incoming traffic. [See the docs for info](https://github.com/DataDog/chaos-controller/blob/main/docs/network_disruption/flow.md#q-why-are-there-limitations-on-ingress) ")
```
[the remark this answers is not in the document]
ptnapoleon: I also linked the docs for onInit disruptions. I felt that these two options are simply not understandable without a longer explanation. We can continue to link the overall docs separate from explain output?

api/v1beta1/network_disruption.go
ptnapoleon: Claire is right that our only internal documentation on this feature suggests not using it. I am in support of removing it, though I see the benefit in situations where an fqdn can't be constructed (theoretically some community users could be in this situation).

api/v1beta1/disruption_types.go
ptnapoleon: I can't find the check for this anymore? looking elsewhere...

PR #983, opened 2025-06-16
CHAOSPLT-751: Add a slackUserEmail field to spec.reporting

JadenR07:
## What does this PR do?

- [ ] Adds new functionality
- [ ] Fixes a bug

**Please briefly describe your changes as well as the motivation behind them:**

Currently, when a disruption is launched from the Chaos UI, we include user information in the request to the Chaos API. This allows us to send Slack messages to the user if there are errors during the Atlas workflow that creates the disruption.

However, if the disruption is successfully created and only fails later while running, the user is not notified. This is because the mechanism for sending Slack error messages relies on the UserInfo annotation, which identifies the disruption creator and maps their email to a Slack user.

The problem is that when a disruption is created via the UI, the UserInfo is set to the disruption-worker service account, not the actual user who launched it—so we can’t determine who to notify when a runtime error occurs.

- Changed slackChannel and purpose to not required

  - Purpose: Ties into [384](https://datadoghq.atlassian.net/browse/CHAOSPLT-384)

- Removed UserInfo from the slack message(the returned struct in buildSlackMessage)
  - Purpose: UserInfo was only used within the file and primarily for the purpose of email verification. The rest of the parameters in UserInfo don’t make much impact on that file. Previously we primarily needed the email stored in UserInfo's username. Now, UserEmail can cover that since it will have the real human email and be used by all thef unctions that previously called UserInfo.username.

- Added userInfo locally to the buildSlackMessage function so we can use userInfo for the fallback mechanism.

- Added 3 test cases:

  - The userInfo username is not configured, defaults to the SlackUserEmail
  
  - Both UserInfo username and SlackUserEmail are defined but differ in value
  
    - Should prioritize the SlackUserEmail
  
  - The slack configuration is successful but the SlackUserEmail is invalid
  
    - Should fallback to the userInfo username

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


eventnotifier/slack/slack.go
```diff
@@ -41,7 +41,7 @@ type slackMessage struct {
 	BodyText    string
 	BodyBlock   slack.SectionBlock
 	InfoBlocks  []*slack.TextBlockObject
-	UserInfo    authv1.UserInfo
+	UserEmail   string
```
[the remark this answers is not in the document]
JadenR07: would it not be enough to have user info locally to the buildslackmessage function? I set it up in a way that it either sets the email from spec.reporting or it checks if it can't, then it uses the UserInfo local to the function to then set the UserEmail. So UserEmail ends up set no matter what and the functions that rely on slack message's and it's user email are all called after storing calling and slack message in a variable

go.mod
```diff
@@ -23,8 +23,8 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/miekg/dns v1.1.55
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/onsi/ginkgo/v2 v2.14.0
-	github.com/onsi/gomega v1.30.0
+	github.com/onsi/ginkgo/v2 v2.23.4
```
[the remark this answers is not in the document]
ptnapoleon:
```
git checkout origin/main -- go.mod go.sum vendor/

git add .

git commit -m 'undo go.mod changes'

```

should do it!

eventnotifier/slack/slack.go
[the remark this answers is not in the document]
ptnapoleon: Okay I see that you are right, we only use the email address from UserInfo, so it's fine to replace it entirely

eventnotifier/slack/slack.go
```diff
@@ -180,16 +180,17 @@ func (n *Notifier) notifyForDisruptionCron(disruptionCron *v1beta1.DisruptionCro
 }
 
 func (n *Notifier) sendMessageToUserChannel(slackMsg slackMessage, logger *zap.SugaredLogger) error {
-	emailAddr, err := mail.ParseAddress(slackMsg.UserInfo.Username)
-	if err != nil {
-		logger.Infow("username could not be parsed as an email address", "err", err, "username", slackMsg.UserInfo.Username)
+	// emailAddr, err := mail.ParseAddress(slackMsg.UserInfo.Username)
```
ptnapoleon, line 183: Hmm, there's a small problem with moving this check from `sendMessageToUserChannel` to `buildSlackMessage`. Now, in situations where we can't parse the username as an email address, we're going to be logging a warning rather than an info. We might need to double check here that UserEmail is defined and is a valid email address.

eventnotifier/slack/slack.go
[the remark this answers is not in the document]
ptnapoleon: It could be enough to make it local to buildSlackMessage, yes. Looks like we only ever extract the username

eventnotifier/slack/slack.go
```diff
@@ -228,28 +228,59 @@ func (n *Notifier) buildSlackMessage(obj client.Object, event corev1.Event, noti
...
+	}
+
+	//do the parse regardless to make sure we have a valid email
+	userEmailAddress, err := mail.ParseAddress(userEmail)
+
+	userEmail = userEmailAddress.Address
+
+	if err != nil {
+		logger.Warnw("unable to retrieve user info", "error", err)
+	}
```
ptnapoleon, line 275: I think we might be able to replace all of L258-L275 with:
```suggestion
if userEmail == "" {
  userEmail = userInfo.Username
}
```

wdyt? We'll keep the same email correctness check in `sendMessageToUserChannel`. That will retain the old behavior

api/v1beta1/disruption_types.go
```diff
@@ -124,17 +124,18 @@ type Reporting struct {
 	// It's expected to follow slack naming conventions https://api.slack.com/methods/conversations.create#naming or slack channel ID format
 	// +kubebuilder:validation:MaxLength=80
 	// +kubebuilder:validation:Pattern=(^[a-z0-9-_]+$)|(^C[A-Z0-9]+$)
-	// +kubebuilder:validation:Required
-	SlackChannel string `json:"slackChannel,omitempty" chaos_validate:"required"`
+	SlackChannel string `json:"slackChannel,omitempty"`
 	// Purpose determines contextual informations about the disruption
 	// a brief context to determines disruption goal
 	// +kubebuilder:validation:MinLength=10
-	// +kubebuilder:validation:Required
```
diyarab, line 132: I think we want to keep the purpose required

api/v1beta1/disruption_types.go
```diff
@@ -124,17 +124,18 @@ type Reporting struct {
...
 	Purpose string `json:"purpose,omitempty"`
 	// MinNotificationType is the minimal notification type we want to receive informations for
 	// In order of importance it's Info, Success, Warning, Error
 	// Default level is considered Success, meaning all info will be ignored
 	MinNotificationType eventtypes.NotificationType `json:"minNotificationType,omitempty"`
+	// SlackUserEmail is the email of the user to send reporting information to
+	// +kubebuilder:validation:Email
```
diyarab, line 137: I might be wrong, but I don't think this marker exists? 
https://book.kubebuilder.io/reference/markers/crd-validation

chart/values.yaml
```diff
@@ -36,8 +36,8 @@ controller:
     noop:
       enabled: true
     slack:
-      enabled: false
-      tokenFilepath: ""
+      enabled: true
+      tokenFilepath: /etc/chaos-controller/slack/token
```
ptnapoleon, line 40: I think we want to leave this part of the oss chart unchanged

examples/slack_reporting_network_drop.yaml
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2025 Datadog, Inc.
+
+apiVersion: chaos.datadoghq.com/v1beta1
+kind: Disruption
+metadata:
+  name: reporting-network-drop
+  namespace: chaos-demo
+spec:
+  reporting: # optional, add custom notification for this disruption
+    slackUserEmail: jaden.ritchie@datadoghq.com # required, custom slack channel to send notifications to (can be a name or slack channel ID)
```
ptnapoleon, line 13:
```suggestion
    slackUserEmail: first.last@example.com # required, custom slack channel to send notifications to (can be a name or slack channel ID)
```

api/v1beta1/disruption_types.go
```diff
@@ -124,17 +124,19 @@ type Reporting struct {
...
-	// +kubebuilder:validation:Required
 	Purpose string `json:"purpose,omitempty"`
 	// MinNotificationType is the minimal notification type we want to receive informations for
 	// In order of importance it's Info, Success, Warning, Error
 	// Default level is considered Success, meaning all info will be ignored
 	MinNotificationType eventtypes.NotificationType `json:"minNotificationType,omitempty"`
+	// SlackUserEmail is the email of the user to send reporting information to
+	// +kubebuilder:validation:MaxLength=320
+	// +kubebuilder:validation:Pattern=(^[a-z0-9-_]+$)|(^C[A-Z0-9]+$)
+	SlackUserEmail string `json:"slackUserEmail,omitempty"`
```
ptnapoleon, line 139: Okay, just two more changes, could you add this field to the output of `Explain()` as well as validating during ValidateCreate that this field is an email address?

api/v1beta1/disruption_types.go
ptnapoleon: I agree with making purpose optional!

eventnotifier/slack/slack_test.go
```diff
@@ -1019,45 +1019,521 @@ var _ = Describe("Slack Notifier", func() {
...
-				DescribeTable("it should not return the error", func(obj k8sclient.Object) {
-					// Arrange
-					expectedUserInfo := v1.UserInfo{Username: "not-an-valid-email"}
-					switch d := obj.(type) {
-					case *v1beta1.Disruption:
-						Expect(d.SetUserInfo(expectedUserInfo)).To(Succeed())
-					case *v1beta1.DisruptionCron:
-						Expect(d.SetUserInfo(expectedUserInfo)).To(Succeed())
-					}
+				Context("with the slack configurastion not in the resource and the invalud user is from userInfo", func() {
```
ptnapoleon, line 1272:
```suggestion
				Context("with the slack configuration not in the resource and the invalid user is from userInfo", func() {
```

eventnotifier/slack/slack_test.go
```diff
@@ -1019,45 +1019,521 @@ var _ = Describe("Slack Notifier", func() {
...
-						},
-					}),
-					Entry("a disruptionCron", &v1beta1.DisruptionCron{
-						TypeMeta: metav1.TypeMeta{
-							Kind: v1beta1.DisruptionCronKind,
-						},
-					}),
-				)
-			})
+				Context("with the slack configuariton in the resource and slackUserEmail is not valid", func() {
```
ptnapoleon, line 1309:
```suggestion
				Context("with the slack configuraiton in the resource and slackUserEmail is not valid", func() {
```

eventnotifier/slack/slack_test.go
```diff
@@ -1019,45 +1019,521 @@ var _ = Describe("Slack Notifier", func() {
...
+						Entry("an error notification with a disruptionCron", &v1beta1.DisruptionCron{
+							TypeMeta: metav1.TypeMeta{
+								Kind: v1beta1.DisruptionCronKind,
+							},
+						}, types.NotificationError),
+					)
+
+				})
+
+				Context("slackUserEmail differs from the userInfoUsername", func() {
```
ptnapoleon, line 1408: How is this different from `the UserInfo username and SlackUserEmail are configured but differ in value`?

api/v1beta1/disruption_types.go
ptnapoleon, line 139:
```suggestion
	SlackUserEmail string `json:"slackUserEmail,omitempty" chaos_validate:"email"`
```
I have some confidence this will work

eventnotifier/slack/slack_test.go
```diff
@@ -1019,45 +1019,521 @@ var _ = Describe("Slack Notifier", func() {
...
-						},
-					}),
-					Entry("a disruptionCron", &v1beta1.DisruptionCron{
-						TypeMeta: metav1.TypeMeta{
-							Kind: v1beta1.DisruptionCronKind,
-						},
-					}),
-				)
-			})
+				Context("with the slack configuraiton in the resource and slackUserEmail is not valid", func() {
```
ptnapoleon, line 1309:
```suggestion
				Context("with the slack configuration in the resource and slackUserEmail is not valid", func() {
```

PR #992, opened 2025-08-22
[CHAOSPLT-1102][FIX][NETWORK] handle DNS response truncation

aymericDD:
## What does this PR do?

- [ ] Adds new functionality
- [x] Fixes a bug

Implements UDP-first with TCP fallback to resolve DNS truncation issues that caused incomplete IP resolution. Adds EDNS0 support with 4KB buffer size to reduce UDP truncation frequency.

The previous implementation ignored the TC (truncated) bit in DNS responses, causing services with many IPs (load balancers, CDNs) to return incomplete address lists. This fix ensures all IPs are discovered by automatically retrying truncated responses over TCP.

## Code Quality Checklist

## Testing

    - `x`
    - [ ] locally.


network/dns.go
```diff
@@ -96,3 +99,26 @@ func (c dnsClient) Resolve(host string) ([]net.IP, error) {
...
+		client.Net = "udp"
+		// Increase EDNS buffer size to reduce truncation
+		// Refer to RFC 5966 https://www.rfc-editor.org/rfc/rfc5966
+		dnsMessage.SetEdns0(4096, true)
+	default:
+		return response, fmt.Errorf("unknown protocol %s. Supported protocols are tcp and udp", protocol)
+	}
+
+	for _, server := range resolvers {
+		response, _, err = client.Exchange(&dnsMessage, fmt.Sprintf("%s:53", server))
```
clairecng, line 120: should we compile the errors in case there's multiple instead of overwriting the last error?

clairecng: what happens for the response? We overwrite every time?

aymericDD: > what happens for the response? We overwrite every time?

You’re right — this was a regression. The response should not be overridden. The code is updated so that it no longer overrides the response. Thanks for catching that! 🙇

PR #1058, opened 2026-04-08
feat(disruption): disk full injection.

Zenithar:
## What does this PR do?

- [x] Adds new functionality

Adds a new `diskFull` disruption kind that genuinely fills a target pod volume using the `fallocate(2)` syscall, causing real ENOSPC errors on all subsequent write operations. This fills a gap where existing disruptions (DiskPressure = I/O throttling, DiskFailure = eBPF on `openat` only) don't simulate actual disk exhaustion visible to monitoring and all syscalls.

### Features

- **Volume fill via ballast file**: Creates a ballast file via `fallocate(2)` syscall (instant, O(1) on ext4/xfs) to genuinely consume disk space. Falls back to writing zeros on unsupported filesystems.
- **Safety**: 1Mi minimum free space floor (overridable via `unsafeMode.allowDiskFullNoFloor`). Pod-level only. Webhook warning for ephemeral-storage eviction risk.
- **Pure Go fallocate**: Vendored `fallocate/` package (adapted from [detailyang/go-fallocate](https://github.com/detailyang/go-fallocate), MIT) — no dependency on `fallocate` or `dd` binaries in the injector image.

### How it differs from existing disruptions

| Disruption | Mechanism | ENOSPC on writes? | Visible to `df`/monitoring? |
|---|---|---|---|
| **Disk Pressure** | Cgroup blkio throttling | No | No |
| **Disk Failure** | eBPF on `openat` only | Only on file open | No |
| **Disk Full (new)** | Real space allocation | Yes (all syscalls) | Yes |

### Example

```yaml
kind: Disruption
metadata:
  name: disk-full-test
spec:
  selector:
    app: my-service
  count: 1
  level: pod
  duration: 10m
  diskFull:
    path: "/data"
    capacity: "95%"
```

## Code Quality Checklist

## Testing

    - [x] by adding new `unit` tests.
    - [x] locally.

### Test coverage

- **Spec validation**: capacity/remaining mutual exclusivity, boundary values, GenerateArgs, Explain
- **Injector**: creation, inject with capacity/remaining, dry-run, remaining > available (skip), inject+clean round trip, idempotent cleanup

### Files changed (24 files, ~1350 lines)

| Component | Files |
|---|---|
| CRD spec + validation | `api/v1beta1/disk_full.go`, `disruption_types.go`, `disruption_webhook.go`, `safemode.go` |
| Injector | `injector/disk_full.go` (ballast file via fallocate) |
| CLI | `cli/injector/disk_full.go`, `cli/injector/main.go` |
| fallocate package | `fallocate/` (4 platform-specific files, adapted from go-fallocate MIT) |
| Safemode | `safemode/safemode_disk_full.go`, `safemode/safemode.go` |
| Types | `types/types.go` (`DisruptionKindDiskFull`) |
| Docs | `docs/disk_full.md`, `docs/disruption_catalogue.md` |
| Tests | `api/v1beta1/disk_full_test.go`, `injector/disk_full_test.go` |


api/v1beta1/disruption_types.go
aymericDD:
```suggestion
	if s.CPUPressure == nil && s.MemoryPressure == nil && s.DiskPressure == nil && s.DiskFailure == nil && s.DiskFull == nil && s.Network == nil && s.GRPC == nil && s.DNS == nil && s.ContainerFailure == nil && s.NodeFailure == nil && s.PodReplacement == nil {
```

aymericDD:
```suggestion
		if s.CPUPressure != nil || s.MemoryPressure != nil || s.DiskPressure != nil || s.DiskFailure != nil || s.DiskFull == nil || s.Network != nil || s.GRPC != nil || s.DNS != nil || s.NodeFailure != nil || s.PodReplacement != nil {
```

aymericDD, line 738:
```suggestion
			s.DiskFailure != nil ||
			s.DiskFull == nil
			{
```

aymericDD:
The diskFull disruption creates a ballast file on the host filesystem via the injector pod. The injector pod mounts the host root at /mnt/host, but that mount has ReadOnly: true — which was correct for all existing injectors (network, CPU, etc.) that only read the host. diskFull must write to the host, so it gets read-only file system ENOSPC before even starting.

**Root cause**                                                                                                                                                              

`services/chaospod.go:573` — the host VolumeMount unconditionally sets ReadOnly: true:

``` golang
     {
         Name:      "host",
         MountPath: "/mnt/host",
         ReadOnly:  true,   // ← must be false for diskFull
     },
```
**Fix**

1. Add a hostWritable bool parameter to generateChaosPodSpec

`services/chaospod.go:466`

Change signature:

``` golang
     func (m *chaosPodService) generateChaosPodSpec(..., hostWritable bool) corev1.PodSpec {
```

Inside the function, use the parameter:

```
     {
         Name:      "host",
         MountPath: "/mnt/host",
         ReadOnly:  !hostWritable,
     },
```

2. Pass kind == DisruptionKindDiskFull from the call site

`services/chaospod.go:332`

``` golang
     Spec: m.generateChaosPodSpec(
         targetNodeName,
         terminationGracePeriod,
         activeDeadlineSeconds,
         args,
         hostPathDirectory,
         hostPathFile,
         kind == chaostypes.DisruptionKindDiskFull,  // hostWritable
     ),
```

     **Critical files**

     - `services/chaospod.go` — only file to modify

Zenithar:
Many thanks for the deep investigation. I will fix that ASAP. I still have concerns about allowing write to a complete FS for writing a ballast in a dedicated directory. It will allow someone with access to the pod to alter the disrupted pod/node for purposes other than the expected disruption.

I will propose a security gate.

cli/injector/disk_full.go
```diff
@@ -0,0 +1,63 @@
...
+				log.Debugln("skipping this injector because path cannot be found on specified container")
+				continue
+			}
+
+			injectors = append(injectors, inj)
+		}
+	},
+}
+
+func init() {
```
aymericDD, line 57: `allowDiskFullNoFloor` is accepted by the webhook, but ignored by the injector

The intended behavior appears to be:
- by default, disk-full disruptions must leave at least 1 MiB free
- if the user explicitly opts into `unsafeMode.allowDiskFullNoFloor: true`, that safety floor is disabled
- in that unsafe mode, requests like `capacity: "100%"` or `remaining: "0"` should be allowed and should actually try to fill to zero free space

That is exactly what the webhook says. In `api/v1beta1/disruption_webhook.go:746`, the safety check starts with:

- if `r.Spec.Unsafemode.AllowDiskFullNoFloor` is true, it returns false, ""
- meaning: “don’t block this request”

So the admission layer is explicitly allowing these requests.

The problem is that the injector layer still does this unconditionally in `injector/disk_full.go:118`:
- it computes bytesToFill
- then it clamps it so at least 1 MiB remains:
    `bytesToFill = availableBytes - minFreeSpaceBytes`

There is no condition tied to unsafe mode there. It always runs.

That means the system behaves like this:
- webhook: “100% is allowed if allowDiskFullNoFloor=true”
- injector: “I will still never use the last MiB”

So the override is only accepted syntactically, not honored behaviorally.

Why the reviewer also mentions “no override bit is passed down”
Because the chaos pod/injector only receives the generated disk-full args from `api/v1beta1/disk_full.go:97`, and those args include only:

```
- disk-full
  - --path
  - optionally --capacity
  - optionally --remaining
```

There is no flag or parameter carrying allowDiskFullNoFloor into the injector. So even if the top-level CRD had unsafe mode set, the injector has no information that it should skip the clamp.

Concrete example:

- suppose the filesystem has 500 MiB available
- user sets capacity: "100%" and unsafeMode.allowDiskFullNoFloor: true
- computeBytesToFill() will likely calculate something close to “fill all remaining free space”
- then injector clamps that to 500 MiB - 1 MiB = 499 MiB

So the user asked for “completely full”, the webhook said “allowed”, but runtime behavior still leaves slack. That is exactly why the reviewer calls it a regression: the new unsafe-mode  option exists, but cannot actually do what it claims.

It breaks a user-visible feature end to end. Users who rely on the override will think they can create a true ENOSPC-at-zero-free-space scenario, but they cannot.

injector/disk_full.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (i *diskFullInjector) GetDisruptionKind() types.DisruptionKindName {
+	return types.DisruptionKindDiskFull
+}
+
+func (i *diskFullInjector) Inject() error {
+	return i.injectVolumeFill()
+}
+
+func (i *diskFullInjector) injectVolumeFill() error {
```
aymericDD, line 97: P2 Use fragment size instead of block size for statfs math — `injector/disk_full.go:104-111`

On Linux, `statfs.Blocks` and `statfs.Bavail` are reported in `Frsize` units, not `Bsize`. Multiplying by `Bsize` works only when those happen to match, so on filesystems where they differ (for example some network/FUSE mounts) the computed totalBytes and availableBytes are wrong, which means capacity and remaining requests will overfill or underfill the volume.

From codex ℹ️ 

_I was not able to reproduce the bug highlighted by Codex_

aymericDD:
Could you also create an example file to test locally the disruption:

`example/disk_full.yaml`

``` yaml
# Unless explicitly stated otherwise all files in this repository are licensed
# under the Apache License Version 2.0.
# This product includes software developed at Datadog (https://www.datadoghq.com/).
# Copyright 2026 Datadog, Inc.

kind: Disruption
metadata:
  name: disk-full
  namespace: chaos-demo
spec:
  level: pod
  selector:
    service: demo-curl
  count: 1
  duration: 10m
  diskFull:
    path: "/mnt/data"
    capacity: "95%"
```

aymericDD:
Could you also update the `examples/complete.yaml` please

aymericDD:
Could you also update the `docs/README.md` to add a link to the `docs/disk_full.md` disruption please

aymericDD:
Full review comments:

- [P1] Remove the pod-path HostPath mount — /Users/aymeric.daurelle/go/src/github.com/DataDog/chaos-controller/services/chaospod.go:689-695
  When `diskFull.path` targets a normal pod volume such as an emptyDir/PVC mounted at `/data`, this value is a container mount path, not a node path. Adding it as a `HostPath` with type `Directory` makes kubelet require `/data` to exist on the node, so the chaos pod can fail before the injector resolves the real `/var/lib/kubelet/pods/...` backing path.

- [P2] Fix the rejected 100% example — /Users/aymeric.daurelle/go/src/github.com/DataDog/chaos-controller/docs/disk_full.md:81-81
  This documented manifest is rejected by `safetyNetDiskFullMinFreeSpace`, because `capacity >= 100` returns an error before `unsafeMode.allowDiskFullNoFloor` is considered. Users following the example cannot create the disruption unless the webhook behavior or the example is changed.

Zenithar:
> Full review comments:
> 
> * [P1] Remove the pod-path HostPath mount — /Users/aymeric.daurelle/go/src/github.com/DataDog/chaos-controller/services/chaospod.go:689-695
>   When `diskFull.path` targets a normal pod volume such as an emptyDir/PVC mounted at `/data`, this value is a container mount path, not a node path. Adding it as a `HostPath` with type `Directory` makes kubelet require `/data` to exist on the node, so the chaos pod can fail before the injector resolves the real `/var/lib/kubelet/pods/...` backing path.
> * [P2] Fix the rejected 100% example — /Users/aymeric.daurelle/go/src/github.com/DataDog/chaos-controller/docs/disk_full.md:81-81
>   This documented manifest is rejected by `safetyNetDiskFullMinFreeSpace`, because `capacity >= 100` returns an error before `unsafeMode.allowDiskFullNoFloor` is considered. Users following the example cannot create the disruption unless the webhook behavior or the example is changed.

@aymericDD, it would be interesting to have configuration validation tests. Because most of the configuration files acting as examples are not validated in CI. That would help to spot such discrepancies, especially when we introduce new feature or schema breaking changes.

injector/disk_full.go
Zenithar: Verified correct as-is. We keep  because  on Darwin (used for unit tests) has no  field. On ext4/xfs — the common case for this disruption — , so the result is correct. A comment at  documents the FUSE/network-mount edge case.

cli/injector/disk_full.go
Zenithar: Fully wired end-to-end now. The fix spans three files:
1. `services/chaospod.go:298-302` — appends `--allow-no-floor` to chaos pod args when `unsafeMode.allowDiskFullNoFloor=true`
2. `cli/injector/disk_full.go:28,37` — reads the flag and passes `AllowNoFloor: allowNoFloor` into `DiskFullInjectorConfig`
3. `injector/disk_full.go:122` — checks `i.config.AllowNoFloor` and skips the 1 MiB clamp when true

Additionally, during verification I found and fixed a related inconsistency in `safetyNetDiskFullMinFreeSpace` (disruption_webhook.go): `capacity: "100%"` was blocked unconditionally even with `AllowDiskFullNoFloor=true`, while the semantically equivalent `remaining: "0"` was allowed. Moved the `AllowDiskFullNoFloor` early-return before the capacity-100% check so both paths are treated consistently. Added 5 webhook tests covering the corrected behaviour.

aymericDD:
> > Full review comments:
> > 
> > * [P1] Remove the pod-path HostPath mount — /Users/aymeric.daurelle/go/src/github.com/DataDog/chaos-controller/services/chaospod.go:689-695
> >   When `diskFull.path` targets a normal pod volume such as an emptyDir/PVC mounted at `/data`, this value is a container mount path, not a node path. Adding it as a `HostPath` with type `Directory` makes kubelet require `/data` to exist on the node, so the chaos pod can fail before the injector resolves the real `/var/lib/kubelet/pods/...` backing path.
> > * [P2] Fix the rejected 100% example — /Users/aymeric.daurelle/go/src/github.com/DataDog/chaos-controller/docs/disk_full.md:81-81
> >   This documented manifest is rejected by `safetyNetDiskFullMinFreeSpace`, because `capacity >= 100` returns an error before `unsafeMode.allowDiskFullNoFloor` is considered. Users following the example cannot create the disruption unless the webhook behavior or the example is changed.
> 
> @aymericDD, it would be interesting to have configuration validation tests. Because most of the configuration files acting as examples are not validated in CI. That would help to spot such discrepancies, especially when we introduce new feature or schema breaking changes.

@Zenithar Good point, we can create a ticket for that with a design proposal and for now you can rely on e2e-test.

aymericDD:
@codex review

services/chaospod.go
```diff
@@ -655,6 +668,46 @@ func (m *chaosPodService) generateChaosPodSpec(targetNodeName string, terminatio
...
+						Type: &hostPathDirectory,
+					},
+				},
+			},
+			corev1.Volume{
+				Name: "disk-full-target",
+				VolumeSource: corev1.VolumeSource{
+					HostPath: &corev1.HostPathVolumeSource{
+						Path: diskFullPath,
+						Type: &hostPathDirectory,
```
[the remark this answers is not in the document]
Zenithar: Fixed: removed the `disk-full-target` Volume and VolumeMount. The `disk-full-kubelet-pods` mount at `/mnt/host/var/lib/kubelet/pods` is sufficient — the injector resolves `spec.Path` to the actual host-side backing path at runtime via `Runtime().HostPath()` (covering emptyDir, PVC, ConfigMap, Secret). No unconditional spec-path HostPath needed.

aymericDD:
@codex review

injector/disk_full.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	hostPath := filepath.Clean(mountHost + path)
+
+	// validate path exists
+	if _, err := os.Stat(hostPath); err != nil {
+		return nil, fmt.Errorf("target path %s does not exist: %w", hostPath, err)
+	}
+
+	ballastPath := filepath.Join(hostPath, ballastFilePrefix+config.Disruption.DisruptionNamespace+"-"+config.Disruption.DisruptionName)
```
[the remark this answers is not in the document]
aymericDD: This comment make sense.

services/chaospod.go
```diff
@@ -663,6 +675,30 @@ func (m *chaosPodService) generateChaosPodSpec(targetNodeName string, terminatio
...
+	// the ballast file. The injector resolves spec.Path to the actual host-side backing path at
+	// runtime via Runtime().HostPath() (e.g. /var/lib/kubelet/pods/<uid>/volumes/...), which
+	// covers emptyDir, PVC, ConfigMap, and Secret volumes.
+	if diskFullPath != "" {
+		podSpec.Volumes = append(podSpec.Volumes,
+			corev1.Volume{
+				Name: "disk-full-kubelet-pods",
+				VolumeSource: corev1.VolumeSource{
+					HostPath: &corev1.HostPathVolumeSource{
+						Path: "/var/lib/kubelet/pods",
```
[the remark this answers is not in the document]
aymericDD: This seems fair too 🤔

examples/disk_full.yaml
aymericDD, line 1: The disruptions results as an error: 

``` json
{
  "level": "error",
  "ts": 1781097797384.7185,
  "caller": "injector/main.go:402",
  "message": "disruption injection failed",
  "disruption_uid": "271dc0b3-6c69-46de-81da-e7bac30bcfda",
  "disruption_name": "disk-full",
  "disruption_namespace": "chaos-demo",
  "target_name": "read-file",
  "target_node_name": "lima-aymeric-daurelle",
  "error": "error creating ballast file /mnt/host/var/lib/rancher/k3s/storage/pvc-2aaaba4e-d0a6-4eb4-a601-8a514d392d15_chaos-demo_demo/.chaos-diskfull-chaos-demo-disk-full: open /mnt/host/var/lib/rancher/k3s/storage/pvc-2aaaba4e-d0a6-4eb4-a601-8a514d392d15_chaos-demo_demo/.chaos-diskfull-chaos-demo-disk-full: read-only file system",
  "stacktrace": "main.inject\n\t/go/src/github.com/chaos-controller/cli/injector/main.go:402\nmain.injectAndWait\n\t/go/src/github.com/chaos-controller/cli/injector/main.go:572\ngithub.com/spf13/cobra.(*Command).execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1019\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1071\nmain.main\n\t/go/src/github.com/chaos-controller/cli/injector/main.go:139\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:290"
}
```

aymericDD:
**Bug: `diskFull` fails on k3s with "read-only file system"**

When targeting a PVC-backed volume on a k3s cluster, the injector fails with:
```
error creating ballast file /mnt/host/var/lib/rancher/k3s/storage/pvc-.../...: read-only file system
```

**Root cause:** The chaos pod adds a writable override mount for `/var/lib/kubelet/pods` (punching through the read-only `/mnt/host`), but on k3s the local-path-provisioner stores PVCs at `/var/lib/rancher/k3s/storage/` — which falls under the read-only parent mount.

**Proposed fix:** Rather than hardcoding the kubelet path, resolve the writable mount path dynamically at chaos pod creation time using the Kubernetes API:
1. The controller already fetches the target `pod` object — find which volume is mounted at `spec.diskFull.path`
2. If it's a PVC → fetch the `PersistentVolumeClaim` → bound `PersistentVolume` → read `pv.Spec.HostPath.Path` or `pv.Spec.Local.Path` → use `filepath.Dir()` as the writable mount
3. emptyDir / ConfigMap / Secret → always fall back to `/var/lib/kubelet/pods` (kubelet-managed, distro-agnostic)

This works for any Kubernetes distribution without configuration or runtime detection heuristics, since the PV host path is always available from the API.

services/chaospod.go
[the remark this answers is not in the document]
aymericDD: Related to:
- https://github.com/DataDog/chaos-controller/pull/1058#discussion_r3388677360
- https://github.com/DataDog/chaos-controller/pull/1058#issuecomment-4670863442

aymericDD (changes requested):

Zenithar:
> **Bug: `diskFull` fails on k3s with "read-only file system"**
> 
> When targeting a PVC-backed volume on a k3s cluster, the injector fails with:
> 
> ```
> error creating ballast file /mnt/host/var/lib/rancher/k3s/storage/pvc-.../...: read-only file system
> ```
> 
> **Root cause:** The chaos pod adds a writable override mount for `/var/lib/kubelet/pods` (punching through the read-only `/mnt/host`), but on k3s the local-path-provisioner stores PVCs at `/var/lib/rancher/k3s/storage/` — which falls under the read-only parent mount.
> 
> **Proposed fix:** Rather than hardcoding the kubelet path, resolve the writable mount path dynamically at chaos pod creation time using the Kubernetes API:
> 
> 1. The controller already fetches the target `pod` object — find which volume is mounted at `spec.diskFull.path`
> 2. If it's a PVC → fetch the `PersistentVolumeClaim` → bound `PersistentVolume` → read `pv.Spec.HostPath.Path` or `pv.Spec.Local.Path` → use `filepath.Dir()` as the writable mount
> 3. emptyDir / ConfigMap / Secret → always fall back to `/var/lib/kubelet/pods` (kubelet-managed, distro-agnostic)
> 
> This works for any Kubernetes distribution without configuration or runtime detection heuristics, since the PV host path is always available from the API.

@aymericDD I don't consider this an error. Trying to create a file on a read-only volume is a configuration error, not a code error. 

For simplicity, I recommend using `emptyDir` so you don't have to handle the path for pod-level disruption. Concerning node-level disruption, I'm torn between removing the base path enforcement, allowing the user to run the disruption in any location, and limiting the user to specific paths; both have pros/cons.

Zenithar:
# Disk-Full Disruption Behaviour Assessment

**Date:** 2026-06-11  
**Cluster:** Lima k3s single-node (`lima-thibault-normand`, k3s v1.26.3+k3s1)  
**Operator version:** `zenithar/chaos-controller/disk_full_disruption` (commit 13817d703 + fixes from this session)  
**Assessed by:** Thibault NORMAND

---

## 1. Expectations (established before cluster interaction)

| ID | Description | Code Location | Level | Status |
|----|-------------|---------------|-------|--------|
| EX-001 | Disruption targets only pods with the matching label selector | `controllers/disruption_controller.go:994` | pod | PASS |
| EX-002 | Controller resolves PVC container path → PV HostPath via Kubernetes API | `services/chaospod.go:711-733` | pod | PASS |
| EX-003 | Chaos pod `disk-full-writable` volume mounts the PV host path writable | `services/chaospod.go:683-701` | pod | PASS |
| EX-004 | Injector resolves `/mnt/data` to the k3s backing path via `Runtime().HostPath()` | `injector/disk_full.go:65` | pod | PASS |
| EX-005 | Ballast file created at `hostPath/.chaos-diskfull-<ns>-<name>-<target>` | `injector/disk_full.go:83` | pod | PASS |
| EX-006 | `capacity: "50%"` fills disk to 50% of total volume size | `injector/disk_full.go:204-224` | pod | PASS |
| EX-007 | `remaining: "80Gi"` leaves exactly 80Gi free (fills the difference) | `injector/disk_full.go:227-238` | pod | PASS |
| EX-008 | Containers without the target path receive a warn log and are skipped | `injector/disk_full.go:70-73` | pod | PASS |
| EX-009 | 1 MiB safety floor clamps fill on second pulse (volume already near target) | `injector/disk_full.go:127-144` | pod | PASS |
| EX-010 | Clean() removes the ballast file; disk returns to baseline | `injector/disk_full.go:248-263` | pod | PASS |
| EX-011 | Clean() is idempotent (no error if file already removed) | `injector/disk_full.go:252-254` | pod | PASS |
| EX-012 | Webhook rejects `level: node` with `disk full disruptions can only be applied at the pod level` | `api/v1beta1/disruption_types.go:816-817` | node | PASS |
| EX-013 | Webhook rejects `containers:` field with disk-full | `api/v1beta1/disruption_types.go:762-763` | pod | PASS |
| EX-014 | Webhook rejects `path: /` | `api/v1beta1/disk_full.go:38-39` | pod | PASS |
| EX-015 | Webhook rejects both `capacity` and `remaining` set simultaneously | `api/v1beta1/disk_full.go:49-55` | pod | PASS |
| EX-016 | `capacity: "100%"` is accepted when safemode is disabled (no floor check) | `api/v1beta1/disruption_webhook.go:275` (safemode gate) | pod | PASS |

---

## 2. Pod-level disruption — PVC-backed volume (k3s local-path)

### 2.1 Test: `capacity: "50%"` on `demo-curl` PVC

**Target:** `demo-curl-547d5bc489-r25pc` (PVC `demo` backed by k3s HostPath PV)  
**PV host path:** `/var/lib/rancher/k3s/storage/pvc-8bd0e3c3-d279-4ade-8552-a418be70a2da_chaos-demo_demo`

**Manifest:**

```yaml
kind: Disruption
metadata:
  name: disk-full-pvc
  namespace: chaos-demo
spec:
  level: pod
  selector:
    service: demo-curl
  count: 1
  duration: 5m
  diskFull:
    path: /mnt/data
    capacity: "50%"
```

**Disk state:**

| Phase | Used | Avail | Use% |
|-------|------|-------|------|
| Baseline | 4.2G | 92G | 5% |
| After injection | 48G | 48G | 50% |
| After cleanup | 4.3G | 92G | 5% |

**Chaos pod volume spec (key section):**
```json
{
  "name": "disk-full-writable",
  "hostPath": {
    "path": "/var/lib/rancher/k3s/storage/pvc-8bd0e3c3-d279-4ade-8552-a418be70a2da_chaos-demo_demo",
    "type": "DirectoryOrCreate"
  }
}
```
Mounted at: `/mnt/host/var/lib/rancher/k3s/storage/pvc-8bd0e3c3-d279-4ade-8552-a418be70a2da_chaos-demo_demo`

**Key injector log entries:**
```json
{"level":"info","message":"injecting disk full disruption",
  "path":"/mnt/host/var/lib/rancher/k3s/storage/pvc-8bd0e3c3-d279-4ade-8552-a418be70a2da_chaos-demo_demo",
  "ballastPath":"/mnt/host/.../  .chaos-diskfull-chaos-demo-disk-full-pvc-demo-curl-547d5bc489-r25pc",
  "bytesToFill":46872190976,"totalBytes":102888095744,"availableBytes":98316238848}
{"level":"info","message":"disk full disruption injected successfully"}
```

**Observations:**
- Dynamic PV resolution via `resolveDiskFullWritablePath` correctly identified k3s storage path.
- Three containers without `/mnt/data` mount received `warn` + skip, not errors.
- Second injection pulse logged `"volume already at or past target fill level, skipping injection"` — 1 MiB floor working correctly.
- Cleanup instant: chaos pod terminated, ballast file removed, disk back to 5%.

---

### 2.2 Test: `remaining: "80Gi"` on `demo-storage-0` (StatefulSet with shared PVC)

**Target:** `demo-storage-0` (PVC `shared-storage-demo-storage-0` backed by k3s HostPath PV)  
**PV host path:** `/var/lib/rancher/k3s/storage/pvc-721998ce-0747-4683-a47c-f024a4a31371_chaos-demo_shared-storage-demo-storage-0`

**Manifest:**

```yaml
kind: Disruption
metadata:
  name: disk-full-remaining
  namespace: chaos-demo
spec:
  level: pod
  selector:
    service: demo-storage
  count: 1
  duration: 5m
  diskFull:
    path: /mnt/shared
    remaining: "80Gi"
```

**Disk state:**

| Phase | Used | Avail | Use% |
|-------|------|-------|------|
| Baseline | 4.1G | 91.7G | 4% |
| After injection | 15.8G | 80.0G | 16% |
| After cleanup | 3.8G | 92.0G | 4% |

**Observations:**
- `remaining: "80Gi"` filled exactly 80Gi target (12.0 GiB ballast file, `bytesToFill=12879540224`).
- Ballast filename: `.chaos-diskfull-chaos-demo-disk-full-remaining-demo-storage-0` — per-target suffix prevents collision between StatefulSet replicas.
- Two containers in `demo-storage-0` share the same `/mnt/shared` volume — both containers would see the same fill effect (correct by design, documented in webhook rejection of `containers:` field).

---

## 3. Webhook validation

| Guard | Input | Expected rejection | Actual rejection | Status |
|-------|-------|--------------------|------------------|--------|
| Node level | `level: node` | "disk full disruptions can only be applied at the pod level" | `* Spec: disk full disruptions can only be applied at the pod level` | PASS |
| Containers specified | `containers: [read-file]` | container isolation error | `* Spec: disk full disruptions apply to the entire volume, specifying certain containers does not isolate the disruption` | PASS |
| Path `/` | `path: /` | path disallowed | `* Spec: path '/' is not allowed for disk full disruptions; it would fill the node's shared container storage` | PASS |
| Mutually exclusive | `capacity: "50%"` + `remaining: "1Gi"` | exactly one required | `* Spec: capacity and remaining are mutually exclusive, only one can be set` | PASS |
| `capacity: "100%"` | No `unsafeMode` | rejected by 1 MiB floor check | **ACCEPTED** (safemode disabled on this cluster) | FINDING-01 |

---

## 4. Code flow summary

```
Disruption CR created
  └── ValidateCreate webhook
        ├── Spec.Validate() — path, capacity%, remaining quantity, mutual exclusion
        ├── level != node check — api/v1beta1/disruption_types.go:816
        └── containers == [] check — api/v1beta1/disruption_types.go:762
                (safemode guards for floor — only if enableSafemode=true)

Reconcile loop
  └── GenerateChaosPodsOfDisruption
        └── GenerateChaosPodOfDisruption
              └── resolveDiskFullWritablePath(disruption, targetName)
                    ├── level == node → return spec.Path directly
                    └── level == pod → resolvePodVolumeHostPath (Reader, 10s timeout)
                          ├── Get Pod from API (uncached Reader)
                          ├── diskFullVolumeAtPath — deepest prefix match
                          ├── PVC volume → resolvePVCHostPath
                          │     ├── Get PVC (Reader)
                          │     ├── Get PV (Reader)
                          │     └── return HostPath.Path or Local.Path
                          └── emptyDir/ConfigMap/Secret → /var/lib/kubelet/pods
              └── generateChaosPodSpec — mounts resolved path writable as disk-full-writable

Chaos pod (injector) runs
  └── NewDiskFullInjector
        └── Runtime().HostPath(ctx, containerID, spec.Path) — 30s timeout
              → resolves container path to host path via container runtime
  └── injectVolumeFill
        ├── syscall.Statfs — get totalBytes, availableBytes
        ├── computeBytesToFill (capacity% or remaining quantity)
        ├── 1 MiB floor check (skip/clamp if AllowNoFloor=false)
        ├── dry-run guard
        └── fallocate.Fallocate — Linux: fallocate(2), Darwin: write zeros
  └── Clean()
        └── os.Remove(ballastPath) — idempotent on ErrNotExist
```

### Relevant files

| File | Role |
|------|------|
| `api/v1beta1/disk_full.go` | DiskFullSpec, Validate(), GenerateArgs() |
| `api/v1beta1/disruption_types.go:762,816` | Level and containers-field restrictions |
| `api/v1beta1/disruption_webhook.go:762-784` | safetyNetDiskFullMinFreeSpace (safemode-gated) |
| `injector/disk_full.go` | NewDiskFullInjector, Inject, Clean |
| `services/chaospod.go:707-808` | Dynamic PV host path resolution |
| `controllers/disruption_controller.go:14-15` | RBAC markers for PVC/PV read access |
| `fallocate/` | Linux fallocate(2) / Darwin zero-write |

---

## 5. Findings

| ID | Description | Severity | Status |
|----|-------------|----------|--------|
| FINDING-01 | `capacity: "100%"` is accepted when `enableSafemode=false` (default in dev clusters). The 1 MiB floor check in `safetyNetDiskFullMinFreeSpace` is gated behind `if enableSafemode`. A production operator with safemode disabled can accidentally schedule a disk-filling disruption with no floor, leaving 0 bytes. | Warn | Known design — operators should enable safemode in prod |
| FINDING-02 | If the k8s node goes into DiskPressure during an active disk-full disruption (e.g., 50% capacity test on a small disk), new chaos pods for subsequent disruptions are **evicted** by the kubelet before they can inject. The node needs ~10s after cleanup to clear DiskPressure. | Info | Expected kubelet behaviour; use conservative capacity values |
| FINDING-03 | Containers without the targeted path receive `warn: could not apply injector on container` — correct skip behaviour, but callers observing `injectedTargetsCount` may see `PartiallyInjected` until the matching container injects. | Info | Expected; documented in skip condition EX-008 |
| FINDING-04 | `resolveDiskFullWritablePath` uses `context.Background()` (not the reconcile ctx). If the reconcile is cancelled mid-flight, the PV resolution continues for up to 10s. | Info | Low-impact; background context with 10s ceiling |

---

## 6. Infrastructure changes made during assessment

| Change | File | Reason |
|--------|------|--------|
| Added `persistentvolumeclaims` and `persistentvolumes` RBAC | `controllers/disruption_controller.go:14-15` | Controller lacked read access to PVCs/PVs needed for dynamic path resolution |
| Switched `resolvePodVolumeHostPath` and `resolvePVCHostPath` to use `m.config.Reader` | `services/chaospod.go:739,769,778` | The cache-backed `Client` has no informers for PVC/PV types; `Reader` bypasses the cache and calls the API directly |

---

## 7. Test artifacts

| File | Description |
|------|-------------|
| `examples/disk_full.yaml` | Existing example (updated selector to `service: demo-curl` for this cluster) |
| `/tmp/disk_full_assess_pod_pvc.yaml` | Pod-level, capacity 50%, `service: demo-curl` — main happy-path test |
| `/tmp/disk_full_assess_remaining.yaml` | Pod-level, remaining 80Gi, `service: demo-storage` — remaining-mode and StatefulSet test |

aymericDD:
@codex review

builderstest/disruption.go
```diff
@@ -89,6 +89,13 @@ func (b *DisruptionBuilder) WithDisruptionKind(kind types.DisruptionKindName) *D
 				if b.Spec.DiskFailure == nil {
 					b.Spec.DiskFailure = &v1beta1.DiskFailureSpec{}
 				}
+			case types.DisruptionKindDiskFull:
+				if b.Spec.DiskFull == nil {
+					b.Spec.DiskFull = &v1beta1.DiskFullSpec{
+						Path:     "/data",
```
aymericDD, line 95: Why are you defining default value for the disruption?

aymericDD (approved):
I stamp it to unblock you

PR #1065, opened 2026-04-14
[CHAOS-209] Fix watcher lifecycle and stuck-on-removal false positives

aymericDD:
## What does this PR do?

- [x] Fixes a bug

This PR addresses a false-positive `IsStuckOnRemoval` status for disruptions with no chaos pods, improves controller reconcile performance by moving the orphan watcher scan out of the hot path, fixes watcher state getting stuck when a disruption is deleted, and adds observability improvements (tracing + runtime metrics).

**Bug fix — skip stuck-on-removal when no chaos pods exist**
A disruption with no chaos pods can always be cleaned immediately. Previously, when the deletion timeout expired, `IsStuckOnRemoval` was set regardless, emitting a false-positive status, warning event, and metric. The flag is now guarded by a chaos pod existence check.

**Performance — orphan watcher scan moved out of reconcile loop**
`RemoveAllOrphanWatchers` was called on every reconcile with one direct API server call per stored watcher manager (O(n)). With many disruptions, this pushed reconcile times over 1 minute. The scan now runs in the existing 5-minute background goroutine using a single `List` call, reducing reconcile-path overhead to zero.

**Bug fix — watcher cleanup gated on uncached API confirmation**
The cached controller-runtime client can transiently return `NotFound` during cache lag or at startup. Previously, the reconciler would remove live watcher state on that false `NotFound`, leaving the disruption unmonitored until the next reconcile event. A direct uncached `APIReader.Get` call now confirms the disruption is truly gone before triggering watcher cleanup.

**Bug fix — stale watcher manager eviction on disruption recreation**
If a disruption was deleted and recreated under the same namespace/name, the watcher manager cache could hold a stale entry keyed by the old `UID`. The manager now tracks each disruption's `UID` and evicts the stale entry on mismatch, preventing dead watcher state from being reused for a new object.

**Feature — targeted O(1) watcher removal on deletion**
`RemoveWatchersForDisruption` is a new method that removes watchers for a single disruption immediately when reconcile determines the object is gone. Previously, cleanup relied on the 5-minute background orphan sweep, leaving watchers alive in the interim.

**Observability — enhanced tracing in Reconcile and watcher management**
- Added spans for orphaned chaos pod handling and watcher synchronization in `Reconcile`
- Introduced `endSpan` helper to streamline span management and error recording
- Added detailed spans for add/remove watcher operations with disruption context attributes
- Improved `SpanContextMutator` to record errors in the disruption span

**Observability — runtime metrics reporting in Datadog tracer**
Enabled Go runtime metrics (heap, goroutines, etc.) reporting to the Datadog Agent.

## Code Quality Checklist

## Testing

    - [x] Added watcher manager unit tests for UID-change eviction and targeted `RemoveWatchersForDisruption` cleanup.
    - Verified disruptions with no chaos pods no longer emit `IsStuckOnRemoval` status/event/metric on deletion timeout
    - Confirmed reconcile times no longer degrade with high disruption counts
    - Verified watcher state is cleaned up immediately on disruption deletion without waiting for the orphan sweep
    - [x] locally.

### DIFF
--- a/controllers/disruption_controller.go
+++ b/controllers/disruption_controller.go
@@ -64,6 +64,7 @@ import (
 // DisruptionReconciler reconciles a Disruption object
 type DisruptionReconciler struct {
 	Client                     client.Client
+	APIReader                  client.Reader
 	BaseLog                    *zap.SugaredLogger
 	Scheme                     *runtime.Scheme
 	Recorder                   record.EventRecorder
@@ -82,21 +83,6 @@ type DisruptionReconciler struct {
 }
 
 const TargetsCountLogLimit = 50
-
-func endSpan(s trace.Span, err error) {
-	if s == nil {
-		return
-	}
-
-	if err != nil {
-		s.RecordError(err)
-		s.SetStatus(codes.Error, err.Error())
-	} else {
-		s.SetStatus(codes.Ok, "")
-	}
-
-	s.End()
-}
 
 func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
 	instance := &chaosv1beta1.Disruption{}
@@ -157,6 +143,15 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			orphanErr := r.ChaosPodService.HandleOrphanedChaosPods(ctx, req)
 			endSpan(orphanSpan, orphanErr)
 			err = orphanErr
+
+			// Confirm deletion via uncached reader before tearing down watchers.
+			// The cached client can transiently return NotFound during cache lag or startup;
+			// removing live watcher state in that case leaves the disruption unmonitored
+			// until the next reconcile event.
+			uncachedErr := r.APIReader.Get(ctx, req.NamespacedName, &chaosv1beta1.Disruption{})
+			if apierrors.IsNotFound(uncachedErr) {
+				r.DisruptionsWatchersManager.RemoveWatchersForDisruption(ctx, req.NamespacedName)
+			}
 		}
 
 		return ctrl.Result{}, client.IgnoreNotFound(err)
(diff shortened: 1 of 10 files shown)


controllers/disruption_controller.go
```diff
@@ -186,19 +250,43 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
-				r.log.Infow("instance seems stuck on removal, the deletion time expired, please check manually")
+				chaosPods, stuckErr := r.ChaosPodService.GetChaosPodsOfDisruption(ctx, instance, nil)
+				if stuckErr != nil {
+					endSpan(stuckSpan, stuckErr)
 
-				// Update the status of the 'instance' to reflect that it's stuck on removal.
-				if err := r.Client.Status().Update(ctx, instance); err != nil {
-					return ctrl.Result{}, fmt.Errorf("error marking the disruption stuck on removal: %w", err)
+					return ctrl.Result{}, fmt.Errorf("error getting chaos pods to check if disruption is stuck on removal: %w", stuckErr)
```
clairecng, line 261: in that error case should we assume that the disruption is stuck for safety?

aymericDD: This should not happened, but I think it's okay, to mark an error and retry at the next reconcile WDYT?

clairecng: yes ok!

controllers/disruption_controller.go
```diff
@@ -131,46 +134,107 @@ func (r *DisruptionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if client.IgnoreNotFound(err) == nil {
 			// If we're reconciling but without an instance, then we must have been triggered by the pod informer
 			// We should check for and delete any orphaned chaos pods
-			err = r.ChaosPodService.HandleOrphanedChaosPods(ctx, req)
+			ctx, orphanSpan := otel.Tracer(tracer.InstrumentationScopeDisruption).Start(ctx, "disruption.handle_orphaned_chaos_pods",
+				trace.WithSpanKind(trace.SpanKindInternal),
+				trace.WithAttributes(
+					attribute.String("disruption.name", req.Name),
```
clairecng: nitpick: we could add those attribute keys as const somewhere (the same way we did with metric keys) to not make any mistake in the future about re-using them

clairecng (approved):

controllers/disruption_controller.go
aymericDD: Good idea 👍 I will do that 🫡

clairecng (approved):

PR #1092, opened 2026-06-12
[FIX][CHAOS] Fix disk-failure cgroupv2 filtering and relative-path disruption

aymericDD:
## Motivation

The eBPF disk-failure injector had two gaps:

1. **cgroupv2 sub-cgroup miss** — When targeting a pod-level disruption, the old PID filter only matched the container's main PID and its direct children. `kubectl exec` creates a `containerd exec-<id>` sub-cgroup whose processes share a different cgroup than the target, so they escaped disruption entirely.

2. **Relative-path opens not disrupted** — If the target process does `cd /mnt/data && cat file`, the kernel receives `openat(AT_FDCWD, "file")` with a relative path. The eBPF program only compared absolute path arguments against `filter_path`, so this open was never disrupted even though `/mnt/data/file` matches the configured prefix.

## Changes

**cgroupv2 process filtering (`cgroup/`, `injector/disk_failure.go`, `ebpf/disk-failure/`)**
- Added `CgroupV2Path()` to `cgroup.Manager` interface and implementation — returns the absolute cgroupv2 unified hierarchy path on v2, empty string on v1.
- Injector passes `-cgroup-path` to `bpf-disk-failure` when a cgroupv2 path is available; the BPF program populates a `BPF_MAP_TYPE_CGROUP_ARRAY` and uses `bpf_current_task_under_cgroup()` to match the container and all its sub-cgroups.
- Gate disk-failure on `HaveCgroupArrayMapType` (alongside the existing `HavePerfEventArrayMapType` check) so unsupported kernels get a clear error before BPF load.

**Relative-path disruption (`injector/disk_failure.go`, `ebpf/disk-failure/`)**
- Injector stats the filter path inside the container via `<ProcRoot>/<pid>/root<path>` and passes two inode/device pairs:
  - `filter_dir_inode` / `filter_dir_dev`: parent directory inode for basename-prefix matching (`cwd=/parent && cat dir/file`)
  - `filter_dir_inode2` / `filter_dir_dev2`: the path's own inode when it is a directory, for exact-CWD matching (`cwd=/dir && cat file`)
- BPF `check_relative_path` runs both checks and only fires when `dirfd == AT_FDCWD`, preventing false matches on non-CWD dirfd opens.
- Device IDs are converted from glibc `encode_dev` format to kernel MKDEV encoding (`major<<20|minor`) to match `super_block.s_dev` — without this the device guard silently rejected every relative open.
- `ProcRoot` is injected via config (default `/proc`) so unit tests set it to a nonexistent path and remain deterministic on both macOS and Linux CI.

## QA Instructions

**cgroupv2 filtering**
1. Deploy a pod disruption with `diskFailure` on a cgroupv2 cluster.
2. `kubectl exec` into the target pod and run `cat /target/file` — it should return the configured error code.
3. Verify that without the fix (`use_cgroup_filter=0`), exec'd processes were not disrupted.

**Relative-path disruption**
1. Configure `paths: [/mnt/data/file]` and have the target process do `cd /mnt/data && cat file` — should now return the configured error.
2. Configure `paths: [/mnt/data]` (directory) and do `cd /mnt/data && cat anyfile` — should be disrupted; `cat ../other` should not.
3. Absolute path opens (existing behavior) must be unaffected.

## Blast Radius

- Only the disk-failure injector is modified; no other disruption kinds are affected.
- The new cgroup array map is unconditionally included in the BPF object; the new `HaveCgroupArrayMapType` gate ensures unsupported kernels fail fast with a clear error rather than crashing at BPF load.
- Relative-path filtering is best-effort (only CWD-relative opens via `AT_FDCWD` are matched; multi-component relative opens from ancestor directories are not yet supported).

## Documentation

- [How to reliably make frontend production changes](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/2286781319/Frontend+Production+Changesp)
- [PR Description Guidelines](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3120334387/PR+Checklist+and+Guidelines+for+Web+UI)
- [Deploy steps](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/1457750032/)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aymericDD:
Closed in favor of https://github.com/DataDog/chaos-controller/pull/1095

